Issue #22, opened 2015-08-03
Support internationalization


kinduff:
Locale files or a way to translate error messages.

outoftime:
:100:

kinduff:
I'll handle this one

PR #23, opened 2015-08-04
Support internationalization

kinduff:
#### What's this PR do?
- [x] Integrates [i18next](https://github.com/i18next/i18next) as internationalization support
- [x] Generates english locale with HTML, CSS and JS errors
- [x] Integrate locale errors to validations

#### Issues
- [x] i18next doesn't load `json` translation file even though it's inside the directory
  - Solved by moving locales directory from `src` to `static`
- [x] `i18n` method undefined when calling it from validation files
- [x] Fix locale entries with semantic error keys

#### Related Issues
#22


kinduff:
@outoftime Having troubles with this integration, I'm not too experienced with node and react. Mind taking a look?

screenshot2

kinduff:
Loading error has been fixed, I'm just missing the `Uncaught ReferenceError: i18n is not defined` one.

outoftime:
@kinduff this is awesome!

In the node/CommonJS module system you need a `require` statement at the top of any file when you need to access a module defined elsewhere. So, at the top of the various `validations` files, you just need to put this line:

```javascript
var i18n = require('i18next-client');
```

As you already do at the top of `validations/index.js`.

I have a few other thoughts I’ll leave inline in the diff, but this is really helpful and looking good!

package.json
```diff
@@ -16,7 +16,8 @@
     "PrettyCSS": "^0.3.10",
     "react": "^0.13.3",
     "reactify": "^1.1.1",
-    "watchify": "^3.2.3"
+    "watchify": "^3.2.3",
+    "i18next-client": "1.10.2"
```
outoftime: Nitpick, but I like to keep the dependencies alphabetized here : )

Also I prefer to use a `^`-prefixed version dependency rather than locking it into a specific version

src/validations/index.js
```diff
@@ -1,4 +1,11 @@
+var i18n = require('i18next-client');
+i18n.init({
+  lng: 'en',
```
outoftime: Can we grab this from the browser? With a fallback to English for languages we don’t support?

src/validations/index.js
```diff
@@ -1,4 +1,11 @@
+var i18n = require('i18next-client');
+i18n.init({
```
outoftime: It looks like `i18next` supports a [`resStore` argument](http://i18next.com/pages/doc_init.html) that would allow us to bundle the translations into the main browserify bundle, but in this case it’s probably better to use the current approach. Worth loading another file to avoid downloading a bunch of translations in languages we don’t need

src/validations/index.js
```diff
@@ -1,4 +1,11 @@
...
+  lng: 'en',
+  debug: true
+});
+
 module.exports = {
+  i18n: i18n,
```
outoftime: No need to export this here—as I noted in the main PR thread, you can just `require` it directly in other modules that you need it in.

static/locales/dev/translation.json
```diff
@@ -0,0 +1,70 @@
+{
```
outoftime: Looks like this file is a dupe?

static/locales/en/translation.json
```diff
@@ -0,0 +1,70 @@
...
+    "extra-tokens-after-value": "Looks like you're missing a semicolon on the line before this one.",
+    "illegal-token-after-combinator": "After a + or > in a selector, you need to specify the name of another element, class, or ID",
+    "invalid-token": "This line doesn't look like valid CSS.",
+    "invalid-value": "__error__ isn't a meaningful value for this property. Double-check what values you can use here.",
+    "require-value": "Put a value for __error__ after the colon.",
+    "selector-expected": "Use a comma to separate multiple tag names, classes, or IDs.",
+    "unknown-property": "__error__ isn't a property that CSS understands. Double-check the name of the property that you want to use."
+  },
+  "html": {
+    "align": "Don't use the \"align\" attribute. Instead, use the CSS text-align property",
```
outoftime: I think for these we should add an extra layer of nesting? E.g. `html.banned-attributes.align` etc.

static/locales/en/translation.json
```diff
@@ -0,0 +1,70 @@
...
+    "background": "Don't use the \"background\" attribute. Instead, use the CSS background property",
+    "bgcolor": "Don't use the \"bgcolor\" attribute. Instead, use the CSS background-color property",
+    "border": "",
+    "frameborder": "Don't use the \"__attribute__\" attribute. Instead, use the CSS border property",
+    "marginwidth": "Don't use the \"marginwidth\" attribute. Instead, use the CSS margin-left and margin-right properties",
+    "marginheight": "Don't use the \"marginheight\" attribute. Instead, use the CSS margin-top and margin-bottom properties",
+    "scrolling": "Don't use the \"scrolling\" attribute. Instead, use the CSS overflow property",
+    "width": "Don't use the \"width\" attribute. Instead, use the CSS width property",
+    "E002": "Use lower case for attribute names",
```
outoftime: I think we should keep the knowledge of error codes in the `validation` module itself—so the keys here would be semantic ones, like `lower-case-attribute` in this case.

static/locales/en/translation.json
outoftime: Update, looks like you did exactly that for [banned tags](https://github.com/outoftime/learnpad/pull/23/files#diff-0a65f52f0c38fc8ad0ff528d5ed105c5R34)

static/locales/en/translation.json
```diff
@@ -0,0 +1,70 @@
...
+      "tt": "You shouldn't use the <tt> tag. Use the <code> tag instead."
+    },
+    "E017": "Use lower case for tag names",
+    "E027": "Put a <title> tag inside your <head> tag",
+    "E028": "You have more than one <title> tag in your document; you should only have one",
+    "E030": "You have an opening tag somewhere in the HTML that doesn't have a closing tag to match it.",
+    "E036": "Lines should be indented with four spaces.\nUse the 'tab' key to increase indentation and the 'delete' key to decrease indentation."
+  },
+  "javascript": {
+    "E019": "There is a __error_a__ on this line that needs a closing __error_a_match__",
```
outoftime: `a` is also kind of a weird artifact of the way `jshint` structures error objects—better to give the translation params semantic names e.g. `opening_symbol` and `closing_symbol` here

static/locales/en/translation.json
```diff
@@ -0,0 +1,70 @@
...
+    "W112": "The string on this line needs to end with another quotation mark",
+    "W116": {
+      "case-1": "You should use === instead of ==",
+      "case-2": "You should use !== instead of !=",
+      "case-3": "There should be a __error_a__ where you wrote __error_b__"
+    },
+    "W117": "You haven't declared __error_a__.\nTry: var __error_a__;",
+    "W123": "You already declared __error_a__ somewhere else."
+  }
+}
```
outoftime: There’s a small handful of other places where English copy is hard-coded in, particularly in [`errorList.jsx`](https://github.com/outoftime/learnpad/blob/master/src/components/errorList.jsx#L28)—would be great to get those into translations as well (totally down to do that in a separate PR if you prefer tho)

kinduff:
@outoftime Sweet notes! I'll update in a bit

kinduff:
@outoftime It's working like a charm now, I'm just missing semantics here.

static/locales/en/translation.json
kinduff: Yeah, let's drop that one into another PR

src/validations/index.js
kinduff: Does it automagically, with `fallbackLng` we got it covered now.

src/validations/index.js
kinduff: @outoftime Is this the right place to init `i18next-client`?

kinduff:
@outoftime This looks better now, waiting for review :panda_face:

src/validations/css/css.js
```diff
@@ -1,17 +1,18 @@
...
   "property missing ':'": function() {
-    return "Put a colon (:) between the property and the value.\nTry: color: red";
+    i18n.t("css.property-missing-colon")
   },
 
   "selector missing": function() {
-    return "Start every block of CSS with a selector, such as an element name or class name.\nTry: p {\n  color: red;\n}";
+    i18n.t("css.selector-missing")
   }
 };
```
kinduff: Having troubles rendering this errors

~~~
Couldn't find a human description for Error: undefined:1:3: missing '{'
    at error (http://127.0.0.1:8080/compiled/bundle.js:45467:15)
    at declarations (http://127.0.0.1:8080/compiled/bundle.js:45652:25)
    at rule (http://127.0.0.1:8080/compiled/bundle.js:45965:21)
    at rules (http://127.0.0.1:8080/compiled/bundle.js:45522:70)
    at stylesheet (http://127.0.0.1:8080/compiled/bundle.js:45486:21)
    at Object.module.exports [as parse] (http://127.0.0.1:8080/compiled/bundle.js:45969:20)
    at module.exports (http://127.0.0.1:8080/compiled/bundle.js:119411:20)
    at module.exports (http://127.0.0.1:8080/compiled/bundle.js:119371:23)
    at React.createClass.handleUpdate (http://127.0.0.1:8080/compiled/bundle.js:119314:5)
    at null.<anonymous> (http://127.0.0.1:8080/compiled/bundle.js:119162:18)
~~~

kinduff: Fixed with `return`. Ruby got me lazy.

src/validations/index.js
outoftime: @kinduff great question—I would say `application.js` is probably the right place for this, actually. As things get more complex we’ll probably need some sort of `initializers` structure but for now `application.js` would be the place for “general setup of the environment”

outoftime:
@kinduff this is looking GREAT! Will be happy to merge once the `i18n` initialization has moved over to `application.js`—thank you so much!

kinduff:
@outoftime Ready! :panda_face:

outoftime:
@kinduff fantastic! it’s happening!!

PR #27, opened 2015-08-19
Add libraries

outoftime:
Possible to toggle libraries on and off and have them included in the page markup.

To do:

- [x] Update validation configuration appropriately (e.g. `$` should be an allowed global iff you have jQuery enabled)
- [x] Persist library choices along with sources
- [x] Fix partial-preview bug (which is even worse now)
- [x] Make toolbar look reasonable

Follow ups:

- [ ] Add *additional* validators (e.g. `bootlint`) where appropriate
- [ ] Switch languages appropriately (e.g. JSX if you’re using React)
- [ ] “Conflict” warnings? (e.g. you probably shouldn’t use Angular and React together)

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Some more stuff that would be good:
 Learnpad uses **React** to manage view state, **Ace** as the code editor, and
 **Browserify** to package the client-side application.
 
-Right now, it includes **htmllint**, **css**, **PrettyCSS", **jshint**, and **jscs** for
+Right now, it includes **htmllint**, **css**, **PrettyCSS**, **jshint**, and **jscs** for
 style checking, although as of now jscs is unused.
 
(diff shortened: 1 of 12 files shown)


kinduff:
Looking good!

outoftime:
@kinduff thanks! I’d say the only blocker to merging it is in fact the way is it *looking*—I’m so bad at design. But hopefully will be able to get a reasonable layout together in the next couple of days.

kinduff:
We could integrate Bootstrap or use some CSS framework to set it up quickly. WDYT?

outoftime:
@kinduff I have always found that Bootstrap makes me slower actually, it’s just kind of confusing. but will give it a shot if I run into a wall with hand-rolling it

PR #30, opened 2015-09-13
Refactor to Flux architecture

outoftime:
Refactor application structure to conform to [Flux application architecture](https://facebook.github.io/flux/docs/overview.html). Information flow is mediated by the dispatcher. Stores provide the source of truth for data state, and subscribe to the dispatcher’s event notifications. Actions provide useful shorthand for building events to send to the dispatcher.

Components subscribe to store changes. This involves making components less hierarchical, and distributing state management among more components (i.e., it’s not just the `Workspace` component that has state now). It’s not immediately obvious where the sweet spot is here, and it’s probably more a matter of intuition. One important observation is that, while components will nearly always generate actions locally, that doesn’t mean they have to manage their own state.

Generally my feeling is that the app had reached a point where an application architecture composed solely of React components was too tightly coupled to handle ever-increasing complexity. The Flux-based architecture seems much more modular, and certainly more testable. Of course, the rubber hits the road when the time comes to actually add a new feature: hopefully Flux should lower the cost of this substantially.


src/actions/ProjectActions.js
```diff
@@ -0,0 +1,41 @@
...
+
+var AppDispatcher = require('../dispatcher/AppDispatcher');
+var ProjectConstants = require('../constants/ProjectConstants');
+var defaults = require('../config').defaults;
+
+var ProjectActions = {
+  create: function() {
+    AppDispatcher.dispatch({
+      actionType: ProjectConstants.PROJECT_CREATED,
+      project: _.cloneDeep(defaults)
```
outoftime: This isn’t the right place to create the project—should happen in the `ProjectStore`

src/actions/ProjectActions.js
```diff
@@ -0,0 +1,41 @@
...
+      actionType: ProjectConstants.PROJECT_LIBRARY_TOGGLED,
+      projectKey: projectKey,
+      libraryKey: libraryKey
+    });
+  },
+
+  loadFromStorage: function(project) {
+    AppDispatcher.dispatch({
+      actionType: ProjectConstants.PROJECT_LOADED_FROM_STORAGE,
+      projectKey: project.projectKey,
```
outoftime: shouldn’t need this

src/components/Editor.jsx
```diff
@@ -0,0 +1,93 @@
...
+    this.editor.moveCursorTo(line, column);
+    this.editor.focus();
+  },
+
+  _setupEditor: function(containerElement) {
+    this.editor = ACE.edit(this.getDOMNode());
+    this._configureSession(this.editor.getSession());
+  },
+
+  _onChange: function() {
```
outoftime: It’s weird that we are directly messing with the editor’s state here rather than using component state to mediate it (or maybe using component state would just be unnecessary indirection)

src/components/Editor.jsx
```diff
@@ -0,0 +1,93 @@
...
+    ErrorStore.removeChangeListener(this._onChange);
+  },
+
+  componentWillReceiveProps: function(nextProps) {
+    if (nextProps.projectKey !== this.props.projectKey) {
+      this._startNewSession(this._getSource(nextProps.projectKey));
+    }
+  },
+
+  shouldComponentUpdate: function(nextProps) {
```
outoftime: Unnecessary argument

src/components/Output.jsx
```diff
@@ -0,0 +1,54 @@
...
+  },
+
+  _onChange: function() {
+    this.setState(this._calculateState());
+  },
+
+  _calculateState: function() {
+    return {
+      hasErrors: ErrorStore.anyErrors(this.props.projectKey),
+      errors: ErrorStore.getErrors(this.props.projectKey)
```
outoftime: Does this qualify as [derived state](https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html)?

PR #50, opened 2016-02-04
Catch and report runtime errors

outoftime:
Runtime errors in JavaScript code are now propagated back to the environment via a postmessage. Currently runtime errors are only displayed as annotations in the editor; they don’t populate the error list because in the case of a runtime error we don’t want to replace the whole preview window. Instead, I’ll follow up with functionality to slide up an error list at the bottom of the preview in the case of a runtime error.

Another followup is to “translate” runtime errors into student-friendly messages. Unfortunately as far as I know there’s no better option here than simply making a bunch of regular expressions to map the matrix of error types + commonly used browsers.

### DIFF
--- a/src/reducers/runtimeErrors.js
+++ b/src/reducers/runtimeErrors.js
@@ -12,6 +12,9 @@ function runtimeErrors(stateIn, action) {
     case 'RUNTIME_ERROR_ADDED':
       return state.push(Immutable.fromJS(action.payload.error));
 
+    case 'RUNTIME_ERRORS_CLEARED':
+      return emptyList;
+
     default:
       return state;
   }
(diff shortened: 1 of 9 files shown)


src/reducers/runtimeErrors.js
```diff
@@ -0,0 +1,20 @@
...
+
+function runtimeErrors(stateIn, action) {
+  var state = stateIn;
+  if (state === undefined) {
+    state = emptyList;
+  }
+
+  switch (action.type) {
+    case 'RUNTIME_ERROR_ADDED':
+      return state.push(Immutable.fromJS(action.payload.error));
```
outoftime: need to clear runtime errors when they are fixed

outoftime: (that’s done)

PR #70, opened 2016-02-15
Use a webworker for validation

outoftime:
To reduce load time, use a webworker for validations rather than including all validators in the main application bundle.


outoftime:
Doesn’t look like a great idea—the size of the main application bundle isn’t substantially reduced, but the worker script is absolutely huge (thanks mostly to the need to include `jsdom` for `slowparse`, I’m guessing)

PR #79, opened 2016-03-05
Log in and save work remotely

outoftime:
The major user-facing change here is that you can now log in to Popcode using your GitHub account, and once you do so, your work will be saved to a remote server and available from other computers when you log in to the same account.

From a technical standpoint, both GitHub authentication and persistence of the workspace are done using Firebase, which turns out to be an excellently suited tool for the job.

A related change is that if you are *not* logged in, the tool no longer persists your work. The local storage behavior would make sense if we expected students to typically use the same computer every day, but in many classroom environments that is not at all the case. In situations where students are using a different shared laptop every day, local storage is actually counterproductive since the effect will be opening Popcode and potentially seeing someone else’s work.

So, no more local storage. If you work logged out and then try to close the page, we pop up a confirmation warning that you’re about to lose work, and instruct you to log in to keep your work.

If you are working logged out and then log in, your work is immediately added as a project in your workspace in remote storage.

There are also a handful of other related changes here, including:

* Don’t use `es6-promise` explicitly; instead, use `browserify-polyfill`, since `Promise` is a language feature in ES2015
* Fix an unrelated bug with a state update cycle in IE if you have a runtime error

A couple more things that should probably happen:

- [x] We still probably want to migrate the *current* project from local to Firebase storage when you log in, ideally only if you’ve made changes to it
- [x] Should warn you when you’re closing the window logged out that you will lose work
- [ ] Should still keep current workspace state in `sessionStorage` to prevent inadvertent work loss


outoftime:
@v @willthefirst @edgenard would love to hear your thoughts on this!

edgenard:
Hey Mat,

The only thing I can comment on for now is the product decision since I'm still not up to speed on any of the technical  things. One of the problems with JSBIN is how sometimes students would forget to log in or save the bin and lose a bunch work. It would often discourage them. From what I understand(and of course I could be wrong)  students could visit popcode previously and do several projects  that would be saved in local storage but now if they visit popcode they would 

1. Have any previously saved projects in local storage but could not make new ones until they are logged in. 
2. Once logged in student work will be saved both locally and persisted to firebase
3. Still unsure of what should happen if they have several projects in local storage and then log in

 It seems that the safest place is to have their projects persisted to firebase(or some server).  Would it be against the idea for popcode to require log ins so that everything can be saved? 

If mandatory log ins would be against the spirit of popcode,would it be a pain to migrate everything to firebase once a student has logged in? 

In either case could the UI reflect(maybe with some big red button) that this work won't be saved until logged in?

outoftime:
Hey @edgenard—this is really helpful! Particularly this:

> One of the problems with JSBIN is how sometimes students would forget to log in or save the bin and lose a bunch work. It would often discourage them.

One bias I have is that in my ScriptEd classroom, the students use Windows accounts that have centralized user profiles, meaning they’re essentially on the “same computer” every class. But that’s not true for most classrooms, and I think it makes sense to design the product around the assumption that every time a student accesses Popcode, it will be from a computer they’ve never used before.

So, I think a good framework for thinking about local storage would be, “local storage is as good as no storage”. And considering the downsides of local storage (including as a vector for students copying each others’ work, intentionally or otherwise), I’m inclined to drop it entirely.

So, my proposal would be:

* If you are not logged in, your work is ephemeral; it’s just a playground.
* When you close the window logged out, we pop up a warning that you are about to lose your work, and that you should log in before closing if you want to save it.
* When you log in, we immediately persist your current work to remote storage and give you access to all projects you’d worked on previously.

How does that sound?

Also in terms of this:

> students could visit popcode previously and do several projects that would be saved in local storage

In theory, yes, but migrating current users isn’t really an issue because there basically are no current users : )

outoftime:
@edgenard just finished changing this to work as discussed above (logged out = no storage, warning when closing). I updated the description of the PR as well. Does this feel like the right direction?

edgenard:
@outoftime I think that the proposal sounds great! I think it is definitely a good direction since it makes the work more portable from machine to machine. 

P.S.  A bit off-topic but I'll be missing the meeting where you unveil popcode because I'm in the Florida for the next two weeks, but I'm learning as much react and es6 as I can so I can start helping out more.

README.md
```diff
@@ -83,7 +83,7 @@ lifecycle of a user interaction:
 * Action creators also perform other business logic, such as initiating
   [validation](https://github.com/popcodeorg/popcode/tree/master/src/validations)
   of project code and persisting changes to
-  [storage](https://github.com/popcodeorg/popcode/blob/master/src/services/Storage.js).
+  [persistent storage](https://github.com/popcodeorg/popcode/blob/master/src/persistors).
```
outoftime: Probably more README updates needed

src/actions/index.js
```diff
@@ -17,70 +26,87 @@ function getCurrentProject(state) {
...
           dispatch({
             type: 'CURRENT_PROJECT_LOADED_FROM_STORAGE',
             payload: {project},
           });
 
-          validateAllSources(dispatch, getCurrentProject(getState()));
+          dispatch(validateAllSources(getCurrentProject(getState())));
         });
       } else {
         createProject()(dispatch, getState);
```
outoftime: Just use normal `dispatch` here

src/actions/index.js
```diff
@@ -163,6 +205,45 @@ function clearRuntimeErrors() {
   };
 }
 
+function resetWorkspace() {
+  return (dispatch) => {
+    dispatch({type: 'RESET_WORKSPACE'});
+    dispatch(loadCurrentProjectFromStorage());
+    dispatch(loadAllProjects());
```
outoftime: Don’t need all this

src/persistors/FirebasePersistor.js
```diff
@@ -0,0 +1,66 @@
...
+          if (error) {
+            reject(error);
+          } else {
+            resolve();
+          }
+        });
+    });
+  }
+
+  saveCurrentProject(project) {
```
outoftime: should return a `Promise`

src/reducers/projects.js
```diff
@@ -2,6 +2,8 @@ import Immutable from 'immutable';
 import {readFileSync} from 'fs';
 import path from 'path';
 
+const emptyList = new Immutable.Map();
```
outoftime: This is more of an `emptyMap`

src/reducers/__tests__/projects-test.js
```diff
@@ -84,7 +86,7 @@ describe('projects', () => {
       const expected = stateIn.setIn(
         ['12345', 'sources', 'css'],
         action.payload.newValue
-      ).toJS();
+      ).setIn(['12345', 'updatedAt'], timestamp).toJS();
```
outoftime: Use batch mutation

src/reducers/projects.js
```diff
@@ -48,6 +50,9 @@ function projects(stateIn, action) {
       return state.setIn(
         [action.payload.projectKey, 'sources', action.payload.language],
         action.payload.newValue
+      ).setIn(
+        [action.payload.projectKey, 'updatedAt'],
+        action.meta.timestamp
```
outoftime: Use batch mutation

src/reducers/projects.js
```diff
@@ -66,8 +71,14 @@ function projects(stateIn, action) {
           }
           return enabledLibraries.add(libraryKey);
         }
+      ).setIn(
+        [action.payload.projectKey, 'updatedAt'],
+        action.meta.timestamp
```
outoftime: Use batch mutation

PR #82, opened 2016-03-07
Don't autoclose HTML tags

v:
Unfortunately, this will also disable smart indenting of HTML
(but it won't disable smart indenting of CSS or JS).

There is no fine-grained way of doing this in the version of
brace that we're using. We have to enable/disable ALL behaviours

### DIFF
--- a/src/components/Editor.jsx
+++ b/src/components/Editor.jsx
@@ -15,8 +15,6 @@ class Editor extends React.Component {
       this._editor.setValue(nextProps.source);
     }
 
-    //this disables autoclosing of HTML tags
-    this._editor.setBehavioursEnabled(false);
     this._editor.getSession().setAnnotations(nextProps.errors);
   }
 
@@ -37,9 +35,14 @@ class Editor extends React.Component {
     if (containerElement) {
       this._editor = ACE.edit(containerElement);
       this._configureSession(this._editor.getSession());
+      this._disableAutoClosing();
     } else {
       this._editor.destroy();
     }
+  }
+
+  _disableAutoClosing() {
+    this._editor.setBehavioursEnabled(false);
   }
 
   _startNewSession(source) {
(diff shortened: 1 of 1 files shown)


src/components/Editor.jsx
```diff
@@ -15,6 +15,8 @@ class Editor extends React.Component {
       this._editor.setValue(nextProps.source);
     }
 
+    //this disables autoclosing of HTML tags
```
outoftime: Rather than putting a comment here, maybe it would be worth just creating a separate instance method e.g. `this._disableAutoClosing()` whose name documents its behavior? I usually like to avoid inline comments as they tend to get out of sync with the thing they’re describing : )

src/components/Editor.jsx
```diff
@@ -15,6 +15,8 @@ class Editor extends React.Component {
...
+    this._editor.setBehavioursEnabled(false);
```
outoftime: I think we want this in the `_setupEditor` method (which is where the `_editor` property is initialized); as written, this will re-apply the setting each time the source is updated, which doesn’t seem necessary.

outoftime:
@v awesome, thank you! I left a couple of comments, let me know what you think (also the CI is failing, although if you replace the comment with a self-documenting method, the lint it detected will no longer be in effect)

src/components/Editor.jsx
v: totally makes sense

src/components/Editor.jsx
v: good call

v:
@outoftime - updated this PR. Sorry for the long turnaround time

outoftime:
Woohoo, thanks @v! This looks great and I plan to release it this evening!

outoftime:
@v just merged, will be in production in a few minutes!

PR #86, opened 2016-03-19
Suppress lower-case-attribute-name error

v:
When there's an unterminated open tag, the next tag is
interpreted as an attribute name. This triggers an
additional HTML lint error because "<" is not a lower
case character.

We're suppressing this error to make it less
confusing for students.


outoftime:
Woohoo!

PR #87, opened 2016-03-25
Behavior tests

outoftime:
Test the application end-to-end, with the whole component tree operating in a (virtual) DOM.

The main goal here is to get a good test harness/set of conventions in place. There are only a handful of actual tests in this PR, although I still think a handful of end-to-end tests are far more useful than a whole directory full of unit tests.

Quick hits:

* Using Mocha instead of Jest as the test framework
* Set up a copy of the entire `<Application/>` component in a virtual DOM for each test
* Aggressively abstract out the details of triggering actions and testing UI state, primarily using functions in the `test/utils/helpers` module
* Avoid mocking unless absolutely necessary (e.g. Firebase); use fully-functional mocks (e.g. `mockFirebase`)

### To do

- [x] Fix warnings when tests are run
- [x] Shouldn’t have to specify `LOG_REDUX_ACTIONS=false` when running tests
- [x] Make this the default for `npm test`


outoftime:
@v WDYT?

PR #99, opened 2016-04-26
Inline CSS and JavaScript libraries

outoftime:
Popcode should be fully functional even with an unreliable internet connection. So, instead of pulling down libraries (e.g. jQuery, Bootstrap) via CDN, inline them directly into the application package.

This is done by installing them via bower, and then reading them in as strings using brfs.


outoftime:
Need to test all the libraries. Also there’s currently a chicken-egg problem with installing the bower components if they’re not already there.

PR #145, opened 2016-08-13
Full support for Flexbox

outoftime:
Supports all flexbox properties:

* `align-content`
* `align-items`
* `align-self`
* `flex-direction`
* `flex-flow`
* `flex-wrap`
* `justify-content`
* `order`


outoftime:
Upstream patch at https://github.com/fidian/PrettyCSS/pull/45

PR #152, opened 2016-08-17
Distinguish between missing opening curly and bogus character in selector

outoftime:
This is a missing opening curly brace:

```css
p
  display: block;
}
```

This is a typo in the selector:

```css
p; div {
  display: block;
}
```

Both look the same to our parsers, though. We can distinguish between the two by checking what type of token tripped the `block-expected` error in PrettyCSS—if it’s an identifier or whitespace, we can guess that it’s a missing curly; otherwise, we can surmise that it’s an invalid character in the selector.


outoftime:
Note that the behavior here isn’t consistently amazing—in particular, PrettyCSS will often identify the missing curly as being on the *last* line, rather than the line where it should be. I can’t think of a good way to deal with this, though, so I’m shipping this for now and making a separate ticket to figure out better behavior with line numbers.

PR #157, opened 2016-08-21
Replace alert() in preview with non-blocking sweetalert

alexpelan:
* Added new module that mimics the structure of the existing libraries file for things that should be injected to the preview by Popcode developers, not the user.
* I had to bower install to get the project to start when checking out a fresh copy, so I added that step
* Didn't do anything test-wise since it doesn't look like there are any existing javascript validation tests

The sweetalert functionality is...fairly meh. Since it doesn't block, if there are multiple alerts, then it just renders the last one as a preview. And we can't replace prompt with it. We could do a lot more by patching it down the line, but it would take some dev effort to have the user writing synchronous, blocking code and us transforming it to be asynchronous with this stuff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,12 @@ This will start a local static server, and open it in your browser. The first
 pageload will be rather slow as it compiles the bundle; after you change files,
 assets are recompiled incrementally and your browser automatically reloads.
 
+When you're done, lint and make sure tests pass before opening a pull request:
+
+```bash
+$ npm test
+```
+
 ## License ##
 
 Popcode is distributed under the MIT license. See the attached LICENSE file
--- a/src/components/Preview.jsx
+++ b/src/components/Preview.jsx
@@ -13,7 +13,7 @@ class Preview extends React.Component {
     bindAll(this, '_handlePopOutClick');
   }
 
-  _generateDocument() {
+  _generateDocument(nonBlockingAlerts = false) {
     if (!this.props.isValid) {
       return '';
     }
@@ -26,7 +26,12 @@ class Preview extends React.Component {
 
     return generatePreview(
       project,
-      {targetBaseTop: true, propagateErrorsToParent: true, breakLoops: true}
+      {
+        targetBaseTop: true,
+        propagateErrorsToParent: true,
+        breakLoops: true,
+        nonBlockingAlerts,
+      }
     ).documentElement.outerHTML;
   }
 
@@ -56,7 +61,7 @@ class Preview extends React.Component {
           onClick={this._handlePopOutClick}
         />
         <PreviewFrame
-          src={this._generateDocument()}
+          src={this._generateDocument(true)}
           onFrameWillRefresh={this.props.onClearRuntimeErrors}
           onRuntimeError={this.props.onRuntimeError}
         />
--- a/src/util/generatePreview.js
+++ b/src/util/generatePreview.js
@@ -3,6 +3,7 @@ import pick from 'lodash/pick';
 import base64 from 'base64-js';
 import loopProtect from 'loop-protect';
 import libraries from '../config/libraries';
+import previewFrameLibraries from '../config/previewFrameLibraries';
 
 const DOMParser = window.DOMParser;
 const parser = new DOMParser();
@@ -38,6 +39,18 @@ const errorHandlerScript = `(${(() => {
   };
 }).toString()}());`;
 
+const alertReplacementScript = `(${(() => {
+  const _swal = window.swal;
+
+  Object.defineProperty(window, // eslint-disable-line prefer-reflect
+    'alert', {
+      value: (message) => {
+        _swal(message);
+      },
+    });
+  delete window.swal; // eslint-disable-line prefer-reflect
+}).toString()}());`;
+
 class PreviewGenerator {
   constructor(project, options = {}) {
     this._project = project;
@@ -49,7 +62,7 @@ class PreviewGenerator {
     this.previewBody = this._ensureElement('body');
 
     this.previewText = (this.previewBody.innerText || '').trim();
-    this._attachLibraries();
+    this._attachLibraries(options.nonBlockingAlerts);
 
     if (options.targetBaseTop) {
       this._addBase();
@@ -58,6 +71,11 @@ class PreviewGenerator {
     if (options.propagateErrorsToParent) {
       this._addErrorHandling();
     }
+
+    if (options.nonBlockingAlerts) {
+      this._addAlertHandling();
+    }
+
     this._addJavascript(pick(options, 'breakLoops'));
   }
 
@@ -119,19 +137,36 @@ class PreviewGenerator {
     this.previewBody.appendChild(scriptTag);
   }
 
-  _attachLibraries() {
+  _addAlertHandling() {
+    const scriptTag = this.previewDocument.createElement('script');
+    scriptTag.innerHTML = alertReplacementScript;
+    this.previewBody.appendChild(scriptTag);
+  }
+
+  _attachLibraries(includePreviewFrame = false) {
     this._project.enabledLibraries.forEach((libraryKey) => {
       const library = libraries[libraryKey];
-      const css = library.css;
-      const javascript = library.javascript;
-      if (css !== undefined) {
-        castArray(css).forEach(this._attachCssLibrary.bind(this));
-      }
-      if (javascript !== undefined) {
-        castArray(javascript).
-          forEach(this._attachJavascriptLibrary.bind(this));
-      }
+      this._attachLibrary(library);
     });
+
+    if (includePreviewFrame) {
+      Object.keys(previewFrameLibraries).forEach((libraryKey) => {
+        const library = previewFrameLibraries[libraryKey];
+        this._attachLibrary(library);
+      });
+    }
+  }
+
+  _attachLibrary(library) {
+    const css = library.css;
+    const javascript = library.javascript;
+    if (css !== undefined) {
+      castArray(css).forEach(this._attachCssLibrary.bind(this));
+    }
+    if (javascript !== undefined) {
+      castArray(javascript).
+        forEach(this._attachJavascriptLibrary.bind(this));
+    }
   }
 
   _attachCssLibrary(css) {
--- a/bower.json
+++ b/bower.json
@@ -35,6 +35,7 @@
     "roboto-webfont-bower": "roboto-webfont#^0.1.1",
     "fontawesome": "^4.6.3",
     "mustache.js": "mustache#^2.2.1",
-    "handlebars": "^4.0.5"
+    "handlebars": "^4.0.5",
+    "sweetalert": "^1.1.3"
   }
 }
--- a/src/config/previewFrameLibraries.js
+++ b/src/config/previewFrameLibraries.js
@@ -1,0 +1,21 @@
+import fs from 'fs';
+import path from 'path';
+
+const previewFrameLibraries = {
+  sweetalert: {
+    name: 'sweetalert',
+    javascript: fs.readFileSync(
+      path.join(__dirname,
+        '../../bower_components/sweetalert/dist/sweetalert.min.js'
+        )
+    ),
+    css: fs.readFileSync(
+      path.join(__dirname,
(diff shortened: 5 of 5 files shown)


outoftime:
Sick!!!

outoftime:
> The sweetalert functionality is...fairly meh. Since it doesn't block, if there are multiple alerts, then it just renders the last one as a preview. And we can't replace prompt with it.

Given that, should we ditch sweetalert and make `alert()` and `prompt()` noops a la jsbin?

README.md
```diff
@@ -103,7 +103,13 @@ Pretty easy. Just check out the code, then run:
 $ npm install
 ```
 
-That'll pull down the dependencies. Then run:
+That'll pull down the dependencies. 
+
+```bash
+$ bower install
```
outoftime: Surprised this was necessary, actually—`bower install` should be a postinstall hook…

package.json
```diff
@@ -116,6 +116,7 @@
     "postcss-cssnext": "^2.7.0",
     "sinon": "^1.17.4",
     "sinon-chai": "^2.8.0",
+    "sweetalert": "^1.1.3",
```
outoftime: Don’t forget to update the shrinkwrap file as well! (I should really get something into CI to check that)

src/config/alwaysOnLibraries.js
```diff
@@ -0,0 +1,21 @@
+import fs from 'fs';
+import path from 'path';
+
+const alwaysOnLibraries = {
```
outoftime: Another possible approach: put SweetAlert in the normal libraries list, but put a flag in the library definition for `alwaysOn` (or maybe `internal`) that suppresses it from the library picker and signals to the preview generator that it should always be included. WDYT?

src/util/generatePreview.js
```diff
@@ -38,6 +39,12 @@ const errorHandlerScript = `(${(() => {
   };
 }).toString()}());`;
 
+const alertReplacementScript = `(${(() => {
+  window.alert = (message) => {
```
outoftime: But doing a `defineProperty` on `Window.prototype` is so much nerdier!

src/util/generatePreview.js
```diff
@@ -38,6 +39,12 @@ const errorHandlerScript = `(${(() => {
...
+    swal(message); // eslint-disable-line no-undef
```
outoftime: Another option would be to call `window.swal` here to avoid the lint situation

outoftime: Also, is it worth doing a `delete window.swal` once we have bound it to `alert`? Would be nice to have the environment the students’ code is running in be as pristine as possible.

src/util/generatePreview.js
```diff
@@ -58,6 +65,7 @@ class PreviewGenerator {
     if (options.propagateErrorsToParent) {
       this._addErrorHandling();
     }
+    this._addAlertHandling();
```
outoftime: Is this happening unconditionally? We definitely want to keep the browser default functionality when we’re generating a pop-out preview I think

src/util/generatePreview.js
```diff
@@ -119,21 +127,36 @@ class PreviewGenerator {
...
+      this._attachLibrary(library);
+    });
+
+    Object.keys(alwaysOnLibraries).forEach((libraryKey) => {
+      const library = alwaysOnLibraries[libraryKey];
+      this._attachLibrary(library);
     });
   }
 
+  _attachLibrary(library) {
```
outoftime: 👍🏻👍🏻👍🏻

outoftime:
@alexpelan this is bomb!! left a bunch of comments, all just starting points for discussion

README.md
alexpelan: Hmm, I see it there - but I ran bower install, not the version installed in gulp - so maybe it was a bower version mismatch on my machine or something? I'll remove it til we see if other folks have problems.

package.json
alexpelan: K, for some reason I thought it worked automagically, haven't used shrinkwrap too extensively.

src/config/alwaysOnLibraries.js
alexpelan: The reason I went with the separate file versus an attribute was not having to deal with all of the complexity of handling it in the picker, etc. Seemed like this is likely to be a one-off?

src/util/generatePreview.js
alexpelan: fine

src/util/generatePreview.js
alexpelan: good thinking x2

alexpelan:
I don't think we should ditch them for no-op - I think it's useful feedback in preview to see the alerts showing up / changing while you are coding. Since prompt isn't addressed by this PR, maybe just no-op it for now?

src/util/generatePreview.js
alexpelan: Added a new option to the PreviewGenerator, whether we are in a popout view or not. If false, then we use sweetalert.

src/util/generatePreview.js
alexpelan: ESLint wants Reflect.defineProperty(), btw.

alexpelan:
Ok, ready for another look if you want to. I tried just 'npm shrinkwrap' but it looked like that added a bunch of artifacts, and wiped out the node version and stuff - so let me know if there are any instructions to follow to do that right, and if i need a certain node/npm version.

outoftime:
@alexpelan ah, yes, you’ll want `npm-shrinkwrap`, not `npm shrinkwrap`. Great build environment ; )

src/config/alwaysOnLibraries.js
outoftime: OK, SGTM

src/components/Preview.jsx
```diff
@@ -13,7 +13,7 @@ class Preview extends React.Component {
     bindAll(this, '_handlePopOutClick');
   }
 
-  _generateDocument() {
+  _generateDocument(isPopout = false) {
```
outoftime: Any reason to take this as a parameter, since it’ll always be `true` in the context of the `Preview` component?

package.json
outoftime: In other news… shouldn’t this be in the Bower config??

outoftime: Oh, no, because the config reads it from `node_modules`.

Hmm—for all other libraries that get vendored in to the preview environment, we use Bower, since Bower packages are explicitly intended to run in a non-module environment, whereas with npm packages there’s only a convention of including a packaged up version that exports something to `window`. All things being equal, I’m inclined to do the same here. What do you think?

src/util/generatePreview.js
```diff
@@ -38,6 +39,19 @@ const errorHandlerScript = `(${(() => {
   };
 }).toString()}());`;
 
+const alertReplacementScript = `(${(() => {
+  // Object.create so our reference
+  // doesn't get destroyed when we delete it on the window
+  const _swal = Object.create({swal: window.swal});
```
outoftime: Hmm, not sure I understand the need for the `Object.create` dance here—can’t we just declare a local variable and assign `window.swal` as its value?

src/util/generatePreview.js
```diff
@@ -38,6 +39,19 @@ const errorHandlerScript = `(${(() => {
...
+
+  Reflect.defineProperty(window, 'alert', {
```
outoftime: `Reflect` is cool, but it’s [not supported by IE or current Safari](http://kangax.github.io/compat-table/es6/#test-Reflect). Since this code is running in the preview environment and thus doesn’t have the benefit of our Babel polyfill, I think we’ll need to use `Object.defineProperty`

src/util/generatePreview.js
```diff
@@ -38,6 +39,19 @@ const errorHandlerScript = `(${(() => {
...
+    value: (message) => {
+      _swal.swal(message);
+    },
+  });
+  Reflect.deleteProperty(window, 'swal');
```
outoftime: Same `Reflect` issue here—believe we’ll need the `delete` keyword

src/components/Preview.jsx
```diff
@@ -26,7 +26,12 @@ class Preview extends React.Component {
 
     return generatePreview(
       project,
-      {targetBaseTop: true, propagateErrorsToParent: true, breakLoops: true}
+      {
+        targetBaseTop: true,
+        propagateErrorsToParent: true,
+        breakLoops: true,
+        isPopout,
```
outoftime: I like keeping the `generatePreview` method unaware of interface-level concepts like a popout—would be inclined to name this `nonBlockingAlerts` or somesuch

src/components/Preview.jsx
outoftime: WAIT I’m a moron.

We should actually be taking this param and using it for _all_ the options being passed to `generatePreview` below.

Still stand by the idea of keeping the popout semantics confined to this component tho

package.json
alexpelan: Agreed, I npm installed sweetalert first thing before I even dug into the code, so I didn't realize how the existing libraries were structured. Better for consistency, I agree.

alexpelan: Also, assuming if I go through bower then npm shrinkwrap doesn't need an update anymore

src/components/Preview.jsx
alexpelan: Not sure I follow - the popout action calls this with true, the in window preview calls it with false. The other 3 parameters should always be true, or at least that seems to be the behavior before this change that I'm making.

src/components/Preview.jsx
alexpelan: Sure, makes sense.

src/util/generatePreview.js
alexpelan: yeah....I swear it wasn't working when I was trying, but it totally works.

src/util/generatePreview.js
alexpelan: K, was just blindly following the linter, I'd never even heard of Reflect until it yelled at me.

alexpelan:
Alright - there was one comment of yours I wasn't sure about but hopefully we're almost ready to merge.

src/util/generatePreview.js
```diff
@@ -119,21 +137,36 @@ class PreviewGenerator {
...
-      if (javascript !== undefined) {
-        castArray(javascript).
-          forEach(this._attachJavascriptLibrary.bind(this));
-      }
+      this._attachLibrary(library);
+    });
+
+    Object.keys(alwaysOnLibraries).forEach((libraryKey) => {
+      const library = alwaysOnLibraries[libraryKey];
+      this._attachLibrary(library);
```
outoftime: Ack, sorry, I didn’t think of this before, **but**: I think we only want to load `sweetalert` if we’re also adding alert handling (i.e. neither should exist in the pop-out environment). So, rather than having the concept of “always on”, I think it’s more a situation of “internal libraries that are available to `generatePreview` to use as needed”. This makes the `_attachLibrary` method even more useful, and also cuts in favor of the approach here you took of keeping sweetalert separate from the user-toggled libraries. Thoughts?

src/components/Preview.jsx
outoftime: @alexpelan right, sorry, that was confusing. my first comment was just wrong. second comment is basically saying that the preexisting code is *also* wrong—we should be varying the values of *all* the options passed to `_generatePreview` from the `_generateDocument` method. However I am down to address that separately since it’s outside the scope of this change. Making a ticket now!

alexpelan:
Ok, it seemed like that was the only thing to address, and we were gonna leave the comment about the different option parameters passed to generatePreview to another change set.  I was kinda assuming you'd squash all these before merge, but let me know - however you want your git history, I can make it so.

outoftime:
Sick! Yes! Let’s do this!

alexpelan:
I don't have bandwidth to make another PR right now, but it's a pretty simple fix, just check if matchType !== "phone" before doing this: https://github.com/gregjacobs/Autolinker.js/blob/master/src/AnchorTagBuilder.js#L94

Good fix for someone new to OSS or new to the project

alexpelan:
Done

outoftime:
Here goooooooes

PR #158, opened 2016-08-22
Improved logic for when to show spinner while typing

outoftime:
If the user is in the middle of typing code, we don’t want to show validation errors about their half-typed code. So, instead, we show a spinner if the user is typing *and* there are currently errors in the source.

Previously, we would do that using the following logic:

* After validation, if there were errors, set a property in the store instructing the UI to delay displaying the errors
* Also after validation, invoke a debounced action dispatcher to set the error display delay flag back to false

This had a couple of disadvantages:

* A session of typing that ended in valid code would still show the spinner for a second before turning green and showing the preview
* When a project with invalid code was loaded from storage, the spinner appeared for a second before showing the errors. This is incoherent

So, we replace that flawed logic with a first-class concept of “currently typing”. This is tracked independently of error validation.

Specifically, the Workspace component dispatches a `userTyped()` whenever
it receives a callback from an `Editor` that the input has changed; and it
uses the `ui.editors.typing` state to determine whether to show a spinner
instead of errors.

This means that we will *only* show a spinner if there are validation
errors *and* the user has recently typed. Importantly, we will no longer
show validation errors if:

* The user has just finished typing, making a previously invalid source
 valid
* Code has just failed validation, but the user was not recently typing
 (e.g. when a project with invalid code is loaded from storage)


outoftime:
@alexpelan want to take a look? : D

.eslintrc
```diff
@@ -272,8 +272,7 @@
             2
         ],
         "no-magic-numbers": [
-            2,
-            {"ignore": [-1, 0, 1, 2, 1000]}
+            0
```
alexpelan: if they're sensible, they ain't magic 👍

alexpelan:
Looks good to me. As a verbose person myself, I appreciate the verbose commit messages.

outoftime:
@alexpelan thanks!!

PR #161, opened 2016-09-02
Minimal modification to pop-out preview

outoftime:
The pop-out preview should be as unadorned as possible—no need to set up a base tag, an error handling script, etc.


outoftime:
@alexpelan you down?

alexpelan:
LGTM

outoftime:
thanks @alexpelan!

PR #162, opened 2016-09-06
Fix disappearing projects in list after switching projects

outoftime:
Fixes bug where switching projects causes all other projects in the list to disappear. This is because I had reversed one of the conditions in the filter for which projects to retain after a project switch.

### DIFF
--- a/spec/examples/actions/projects.spec.js
+++ b/spec/examples/actions/projects.spec.js
@@ -5,7 +5,12 @@ import {assert} from 'chai';
 
 import createApplicationStore from '../../../src/createApplicationStore';
 
-import {createProject} from '../../../src/actions/projects';
+import {
+  createProject,
+  changeCurrentProject,
+} from '../../../src/actions/projects';
+
+import {toggleLibrary} from '../../../src/actions';
 
 import {
   getProjectKeys,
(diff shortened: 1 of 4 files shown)


spec/examples/actions/projects.spec.js
```diff
@@ -38,4 +43,26 @@ describe('projectActions', () => {
...
+    let projectKey;
+
+    beforeEach(() => {
+      store.dispatch(createProject());
+      projectKey = getCurrentProject(store.getState()).projectKey;
+      store.dispatch(toggleLibrary(projectKey, 'jquery'));
+      store.dispatch(createProject());
+      const secondProjectKey = getCurrentProject(store.getState()).projectKey;
+      store.dispatch(toggleLibrary(secondProjectKey, 'jquery'));
+      store.dispatch(changeCurrentProject(projectKey));
```
outoftime: @alexpelan what do you think of this sort of thing? Specifically, setting up a particular state for the store by firing off a bunch of actions to put it in that state? The main alternative on my mind would be directly specifying what the initial state should be (probably using an abstraction specifically built for tests)

outoftime:
Gonna go ahead and merge this since it fixes a substantial bug but definitely interested in your thoughts @alexpelan ⬆️

spec/examples/actions/projects.spec.js
alexpelan: I think I usually end up with a mixture of firing actions (or the equivalent) and some qa abstraction. In my experience super common things like creating a project, creating a user usually end up behind an abstraction, and less common cases we specify the state.

And I vote for firing actions versus directly specifying because then you get updates to those actions for free in your test suite when the actions update, versus potentially having a stale state.

outoftime:
> And I vote for firing actions versus directly specifying because then you get updates to those actions for free in your test suite when the actions update, versus potentially having a stale state.

Quite! OK, I’m convinced—let’s stick with building state using a series of actions, with aggressive abstraction for both keeping tests semantic and encapsulating repetitive setup work.

PR #163, opened 2016-09-07
Disallow blocking prompt in preview

alexpelan:
Not sure if you wanted to do anything fancier than this, it just ignores whatever is passed in, doesn't block, and returns the second parameter, or the empty string. This way users can continue to write code after prompts and do string methods, coerce types, etc, without getting errors (I'm sure there are cases where this is not true).


outoftime:
Sick! I totally did not know about the default value parameter for `prompt()`

PR #168, opened 2016-09-15
Improved support for old browsers

outoftime:
* Add a polyfill for ES6 Set, specifically to ensure that `Set.prototype.forEach` is available. Lodash relies on this, but some older browser versions that we can otherwise support don’t have it.
* Increase version requirements to IE11 and Firefox 28, which are respectively the first versions to fully support modern flexbox. Popcode’s layout heavily depends on flexbox.
* Pass list of supported browsers to `cssnext`/`autoprefixer`, so that our CSS will contain all appropriate vendor prefixes for the browser versions we support.

### DIFF
--- a/config/browsers.json
+++ b/config/browsers.json
@@ -1,0 +1,6 @@
+{
+  "msie": "11",
+  "firefox": "28",
+  "chrome": "30",
+  "safari": "7.1"
+}
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -12,6 +12,7 @@ import uglify from 'gulp-uglify';
 import cssnano from 'gulp-cssnano';
 import gutil from 'gulp-util';
 import memoize from 'lodash/memoize';
+import forOwn from 'lodash/forOwn';
 import brfs from 'brfs-babel';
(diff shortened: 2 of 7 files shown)


outoftime:
@alexpelan WDYT?

alexpelan:
Commenting on this late and after merge, but the code looks fine. Since it doesn't work on the browsers that are now listed as unsupported anyway, no big deal. And we have no indication from any of the ScriptEd staff that any classrooms are locked down on IE9 for some obscure reason? (I come from the land of enterprise healthcare, where we still support back to IE8 in 2016)

outoftime:
@alexpelan yep—for the most part the browsers that are getting caught are ones in which Popcode just doesn’t work at all, so we’re just making the best of a bad situation. There are a few marginal Firefox versions where it would generally work, but incomplete support for Flexbox breaks the layout of the dashboard; it didn’t seem worth trying to work around that. We are supporting FF versions released in 2013, after all.

That said, it seems like it would definitely be worth setting up some analytics to make sure a lot of students aren’t trying to use it with unsupported browsers.

PR #170, opened 2016-09-17
Handle click of error associated with minimized editor

outoftime:
In this situation:

* Type invalid code into an editor
* Minimize that editor
* Then click on the error message in the list

Expected behavior: The editor is maximized and focused, with the cursor at the position of the erroneous code.

Previous behavior: uncaught exception.

This pull request afforded an opportunity to finally fix one of my least favorite things about Popcode’s implementation, namely that the `Workspace` component held references to the `Editor` component instances that it owned. Worse, when an error was clicked, the `Workspace` would call a method *directly on the `Editor`*. This is not how components are supposed to work. It’s also the cause of the uncaught exception—with the editor minimized, the `Editor` component was not mounted, so the `Workspace` was trying to call a method on an undefined component reference.

I have had a lot of trouble coming up with a satisfying way to represent what feels like a completely transient event—“the user clicked an error”—and a completely transient reaction—“move the cursor in the appropriate editor to the position of the error”—in Flux state. Specifically, I *don’t* want to represent the cursor position and focus state of each editor in the Flux store—that would involve a ton of unnecessary data flow just so we can handle one fairly niche user interaction. In this pull request, I took a different approach, which I think is pretty good:

* There is a `ui.editors.requestedFocusedLine` state property. This is usually `null`, but when populated, it contains information about a line that the user wants the cursor to move to, but which we have not yet actually moved the cursor to.
* When the user clicks an error, we dispatch a `USER_REQUESTED_FOCUSED_LINE` action, which causes the aforementioned property to populate.
* The `Editor` component pays attention to this property; if it receives a value that corresponds to its language, it moves the cursor appropriately, and then immediately dispatches an `EDITOR_FOCUSED_REQUESTED_LINE` action.
* This action causes the `ui.editors.requestedFocusedLine` property go back to `null`.

Unsurprisingly, migrating this interaction to a pure one-way data flow makes it a lot easier to work with and reason about. In general, we now know that, at any given time, we’re in one of three situations:

1. The user hasn’t clicked an error.
2. The user clicked an error, but we haven’t moved the cursor yet.
3. The user clicked an error, and we moved the cursor.

States 1 and 3 end up looking identical in the Flux state tree.

This approach solves a specific problem with a naïve solution to the original bug. When an error is clicked, we need to maximize the editor.  This is done by issuing a `COMPONENT_MAXIMIZED` action. However, this won’t cause the component to mount synchronously; we need to wait for React to flush component state. That means, if we want to call a method directly on a component we’re holding a `ref` to, we need to do so in some kind of next-tick block, which is unsatisfying and fragile.

Instead, by using Flux state to communicate the user’s intent to focus a line, we stop having to worry about the exact sequence of rendering.  When the `Editor` component mounts, it checks to see if there is a pending focused-line request, and focuses the line if there is. It also, of course, checks for focused-line requests when props change. So, the right thing happens regardless of the order things happen in.

Generally, this approach seems to offer a promising pattern for situations where one-way data flow and components as functions of state seem ill-suited. By representing a seemingly transient event as a fleeting but real state—something needs to happen, but hasn’t happened yet—we are able to keep components loosely coupled, get more robust behavior, and have an implementation that’s easier to reason about.

### DIFF
--- a/spec/examples/actions/ui.spec.js
+++ b/spec/examples/actions/ui.spec.js
@@ -6,7 +6,12 @@ import {useFakeTimers} from 'sinon';
 import createApplicationStore from '../../../src/createApplicationStore';
 import waitFor from '../../helpers/waitFor';
 
-import {TYPING_DEBOUNCE_DELAY, userTyped} from '../../../src/actions/ui';
+import {
+  TYPING_DEBOUNCE_DELAY,
+  editorFocusedRequestedLine,
+  userTyped,
+  userRequestedFocusedLine,
+} from '../../../src/actions/ui';
 
 const timeInterval = 1000 * 60 * 60 * 24;
 
@@ -54,5 +59,31 @@ describe('interfaceStateActions', () => {
         })
       )
     );
+  });
+
+  describe('userRequestedFocusedLine', () => {
+    beforeEach(
+      () => store.dispatch(userRequestedFocusedLine('javascript', 4, 2))
+    );
+
+    it('sets requestedFocusedLine to given value', () => {
+      assert.deepEqual(
+        store.getState().ui.getIn(['editors', 'requestedFocusedLine']).toJS(),
+        {language: 'javascript', line: 4, column: 2}
+      );
+    });
+  });
+
+  describe('editorFocusedRequestedLine', () => {
+    beforeEach(() => {
+      store.dispatch(userRequestedFocusedLine('javascript', 4, 2));
+      store.dispatch(editorFocusedRequestedLine());
+    });
+
+    it('sets requestedFocusedLine to null', () => {
+      assert.isNull(
+        store.getState().ui.getIn(['editors', 'requestedFocusedLine']),
+      );
+    });
   });
 });
--- a/src/components/Editor.jsx
+++ b/src/components/Editor.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import ACE from 'brace';
 import bindAll from 'lodash/bindAll';
+import get from 'lodash/get';
 import throttle from 'lodash/throttle';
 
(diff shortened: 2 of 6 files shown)


outoftime:
@alexpelan what upppppp

src/components/Editor.jsx
```diff
@@ -58,9 +62,18 @@ class Editor extends React.Component {
...
+  _focusRequestedLine(requestedFocusedLine) {
+    if (get(requestedFocusedLine, 'language') !== this.props.language) {
+      return;
+    }
+
+    this._editor.moveCursorTo(
+      requestedFocusedLine.line,
+      requestedFocusedLine.column
+    );
+    this._editor.clearSelection();
```
outoftime: I threw in another fix for a related bug, which is that if your cursor is in a later position in the editor than the error you’re jumping to, the intervening text will be selected. Now it’s not.

spec/examples/actions/ui.spec.js
```diff
@@ -55,4 +60,30 @@ describe('interfaceStateActions', () => {
...
+
+    it('sets requestedFocusedLine to given value', () => {
+      assert.deepEqual(
+        store.getState().ui.getIn(['editors', 'requestedFocusedLine']).toJS(),
+        {language: 'javascript', line: 4, column: 2}
+      );
+    });
+  });
+
+  describe('userRequestedFocusedLine', () => {
```
alexpelan: Is this correct? should this describe editorFocusedRequestedLine, and the second describe say that it nulls out the focused line state?

outoftime: @alexpelan ooh you caught me!!!

PR #174, opened 2016-09-21
Fix javascript error when redefining property

alexpelan:
Object.defineProperties by default uses configurable: true when
defining a property, which means that if you attempt to define that
property again later there will be a TypeError. This makes us fault
tolerant, which we want to be since the TypeError shows up in the UI.

### DIFF
--- a/src/util/generatePreview.js
+++ b/src/util/generatePreview.js
@@ -47,9 +47,11 @@ const alertAndPromptReplacementScript = `(${(() => {
       value: (message) => {
         _swal(message);
       },
+      configurable: true,
     },
     prompt: {
       value: (message, defaultValue = '') => defaultValue,
+      configurable: true,
     },
   });
 


alexpelan:
I wasn't able to reproduce this bug, but I saw it happening on multiple machines today in class (different chrome version? slower network connection?). I suppose a better fix would be to make sure we don't call this code twice, but this will at least make the error go away. Feel free to reject if you want me to take a deeper dive into what's actually happening, but allowing alert and prompt to be redefined by us seems harmless to me. 

relevant: http://stackoverflow.com/questions/25517989/why-cant-i-redefine-a-property-in-a-javascript-object

outoftime:
@alexpelan word up! I am curious about why it’s happening, *but* I’m also down with this change on its merits—in particular, the default `window.alert` is `configurable`, so for consistency our redefined one ought to be too.

That said, how confident are we that the error was the result of double-defining `alert`? Is it possible that they were using a version of Chrome that doesn’t allow redefining `window.alert` at all?

PR #175, opened 2016-09-21
Gracefully handle failed sign-in

outoftime:
We saw two classes of sign in error: one where the user intentionally cancels sign in (by closing the popup without choosing to authorize), and another vaguer failure.

In the first case, we’ll display a specific message instructing the user to try again and click the green button:

In the second case, we’ll display a more general message, and also notify Bugsnag of the error, including the error code that came back from Firebase.

This means generally setting up a framework for displaying application-level errors in red banners at the top of the screen, which can then be dismissed.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -21,6 +21,10 @@
     "message": "Popcode doesn’t work with your version of __name__.",
     "download": "Click here to download a new version."
   },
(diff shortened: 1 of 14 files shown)


outoftime:
@alexpelan wdyt?

PR #176, opened 2016-09-25
Use Webpack to bundle application

outoftime:
The editor just takes way too long to load right now, and by far the best thing we can do to fix it is to asynchronously load code. In particular, we should be able to asynchronously load all the linters, as well as all the external libraries, since neither is required for the editor to basically function.

So, Webpack. A pain to get set up, but asynchronous loading is dead simple (I already tried it, although it’s not in this pull request).

In order to get a good build system working with Webpack, I bit the bullet and merged in the [vigetlabs gulp starter](https://github.com/vigetlabs/gulp-starter). This involved some substantial rearrangement of the code structure to conform to the starter, but the old code structure was not based on any strong opinions of mine, so I figured I’d go with it.

One bonus is a much nicer development environment. For instance, the development server doesn’t crash when you have a syntax error. Also, hot CSS injection and code reloading!!!


outoftime:
@alexpelan this is a bit of a monster but not as bad as it looks (largely just moving files, and adding a bunch of gulp tasks)—any chance you could give it a skim and let me know what you think?

gulpfile.babel.js
```diff
@@ -1,161 +0,0 @@
...
-  extensions: ['.jsx'],
-  debug: true,
-  fullPaths: !gulp.env.production,
-};
-
-const buildBrowserifyCompiler = memoize(
-  (filename) => browserifyImpl(`src/${filename}`, browserifyOpts).
-    transform(brfs).
-    transform(babelify.configure({sourceMapRelative: __dirname})).
-    transform(envify)
```
outoftime: Seems like this might be missing from the new build system??

spec/helper.js
```diff
@@ -2,7 +2,5 @@ import chai from 'chai';
 import chaiImmutable from 'chai-immutable';
 import chaiAsPromised from 'chai-as-promised';
 
-import '../src/init';
```
outoftime: I want to get to the bottom of why this causes problems

package.json
```diff
@@ -69,63 +67,106 @@
...
+    "production": "gulp production",
+    "demo": "gulp production && gulp server",
+    "postinstall": "bower install",
     "pretest": "npm run lint",
-    "test": "./node_modules/.bin/karma start --single-run --no-auto-watch",
-    "autotest": "./node_modules/.bin/karma start --no-single-run --auto-watch",
-    "lint": "./node_modules/.bin/eslint --ext .js,.jsx --plugin react src"
+    "test": "gulp production && karma start --single-run --no-auto-watch",
+    "test:watch": "karma start --no-single-run --auto-watch",
+    "lint": "eslint --ext .js,.jsx --plugin react src"
```
outoftime: need to update the README

alexpelan:
So, knowing nothing about webpack, nothing jumped out at me. But it might be worth having someone with production webpack experience giving a second opinion.

outoftime:
@alexpelan more reasonable?

PR #177, opened 2016-09-28
Reduce entry bundle size

outoftime:
This reduces the size of the entry bundle by about 50%, from ~1.6MB to about 750K gzipped. Still have some work to do for sure, but it’s a real improvement.

Three changes:

* Linters are loaded asynchronously. The linters make up a big chunk of our total vendor code, and they’re generally not needed exactly at load time. So, we load the linters asynchronously, though still greedily, which is to say we initiate the async request as soon as the main bundle loads. This means the linters should still be available by the time the user actually needs to validate something, but downloading them is not blocking load of the initial UI.
* Remove libraries that no one is using. We keep jQuery, Lodash, Mustache, and Bootstrap, and ditch everything else. At some point we can bring them back, but I don’t think there’s any rush.
* Exclude ACE worker code from the bundle. ACE has a validation system of its own, which runs in a web worker. However, we don’t use this system, and we explicitly disable it. The language modes we *do* use (for syntax highlight and such) dynamically require the workers in a code path that never actually runs in our application; however, Webpack has no way of knowing that, so it still bundles the code in. We now spike the worker code for the bundle by applying a null loader to that directory.

There’s a bunch more we can do, including:

* Figure out why it seems that all of `lodash` is getting bundled, even though we specifically require only the modules we need.
* Look into `moment` and `text-encoding`—these are big contributors to bundle size for fairly minor usefulness.
* Asynchronously load the libraries that we kept.
* Perhaps asynchronously load more vendor code—e.g., we don’t need the GitHub or Firebase clients to render the UI either.
* Split out a vendor entry for better caching behavior.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,6687 @@
+{
+  "name": "popcode",
+  "version": "0.9.0",
+  "npm-shrinkwrap-version": "200.5.1",
+  "node-version": "v6.3.1",
+  "dependencies": {
+    "Base64": {
+      "version": "0.2.1",
+      "resolved": "https://registry.npmjs.org/Base64/-/Base64-0.2.1.tgz"
+    },
+    "PrettyCSS": {
+      "version": "0.3.12",
+      "resolved": "https://registry.npmjs.org/PrettyCSS/-/PrettyCSS-0.3.12.tgz"
+    },
+    "abbrev": {
+      "version": "1.0.9",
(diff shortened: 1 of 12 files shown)


outoftime:
@alexpelan WDYT?

outoftime:
@alexpelan gonna go ahead and fire on this, but still interested to hear your thoughts / down to follow up with changes if you think they are needed!

alexpelan:
Looks fine to me - this is gonna make a huge difference

PR #196, opened 2016-10-12
Ensure we can read gist from GitHub API before attempting to open it

outoftime:
Various classes have observed situations in which newly-created gists, when opened in the new tab, yield a GitHub 404 page. This is odd behavior, considering we are just using the `html_url` returned by the GitHub API.

Bit of a shot in the dark here, since the error isn’t readily reproduced, but my theory right now is that we’re hitting an eventual consistency situation or a slightly laggy database replica on GitHub’s end.

To try to combat this, after creating (and possibly updating) the gist, we send an API request to fetch the newly created gist. If the request fails, we’ll retry with exponential backoff several times.

The hope is that a 404 on the web URL will correlate with a 404 in the API. There is certainly no guarantee of this, but it’s worth a shot.

### DIFF
--- a/src/services/Gists.js
+++ b/src/services/Gists.js
@@ -1,6 +1,7 @@
 import GitHub from 'github-api';
 import trim from 'lodash/trim';
 import isEmpty from 'lodash/isEmpty';
+import promiseRetry from 'promise-retry';
 const anonymousGithub = new GitHub({});
 
 export function EmptyGistError(message) {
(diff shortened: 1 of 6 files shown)


src/services/Gists.js
```diff
@@ -63,22 +62,7 @@ function updateGistWithImportUrl(github, gistData) {
...
-function notifyAndRejectApiError(error) {
-  if (error.response) {
-    Bugsnag.notifyException(
-      error,
-      'GistError',
-      {failedRequest: pick(error, ['request', 'response', 'status'])}
-    );
-  }
-
-  return Promise.reject(error);
```
outoftime: We don’t really need this anymore, because the newer version of the GitHub client now gives us useful error information.

alexpelan:
lgtm

outoftime:
@alexpelan thanks!

PR #199, opened 2016-10-12
simple fix to bring dashboard items to the front

carolchau:
Simple fix to the problem where in smaller window sizes, the Pop svg blocks the click target for the dashboard menu items. By bringing the dashboard menu items to the front, the Pop svg will not block them.

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -236,6 +236,7 @@ body {
   margin: 1px 0;
   padding: 0.2rem 0.5rem;
   text-decoration: none;
+  z-index: 1;
 }
 
 .dashboard-menu-item--grid {


outoftime:
@carolchau thank you! I’m down with this as a quick fix, although I’m going to leave the issue open because I’d really like to get Pop to resize correctly in this situation. Unfortunately CSS is not my strong suit so I haven’t quite figured out how to do that yet even though it seems like it should be pretty straightforward.

carolchau:
@outoftime yeah, no problem -- that sounds like a good plan! It does seem like it should be pretty straightforward. Good luck and go team Popcode!

Issue #195, opened 2016-10-11
User Feedback


outoftime:
Submitter: Vaibhav Verma
Email: vaibhav2614@gmail.com
It would be cool if Libraries were remember when I exported to gist, and loaded popcode from a gist like so:

https://popcode.org?gist=593577ab539fa607496c206bfc2d0696

outoftime:
@v agreed! in fact it’s [already on the roadmap](https://trello.com/c/8407ODMg/175-include-enabled-libraries-when-exporting-importing-gist) : )

PR #200, opened 2016-10-13
Export libraries to popcode.json and reload when importing

alexpelan:
popcode.json could later be extended to save any other metadata that
we want to.

Fixes: #195

### DIFF
--- a/spec/examples/actions/index.spec.js
+++ b/spec/examples/actions/index.spec.js
@@ -1,0 +1,73 @@
+/* eslint-env mocha */
+/* global sinon */
+
+import '../../helper';
+import {assert} from 'chai';
+
+import createApplicationStore from '../../../src/createApplicationStore';
+
+import {
+  createProject,
+} from '../../../src/actions/projects';
+
+import {
+  importProjectFromGist,
+} from '../../../src/actions/index';
+
+import {
+  getProjectKeys,
+  getCurrentProject,
+} from '../../../src/util/projectUtils';
+
+describe('indexActions', () => {
+  let store;
+
+  beforeEach(() => {
+    store = createApplicationStore();
+  });
+
+  describe('importProjectFromGist', () => {
+    it('sets enabledLibraries to default if no popcode.json exists', () => {
+      const gistData = {
+        files: {
+          "index.html":
+            {
+              language: 'html',
+              content: '<h1>Hello, world!</h1>',
+            },
+        },
+      };
+      store.dispatch(createProject());
+      const currentProject = getCurrentProject(store.getState());
+      store.dispatch(importProjectFromGist(currentProject.projectKey, gistData))
+      const updatedProject = getCurrentProject(store.getState());
+      assert.equal(updatedProject.enabledLibraries.length, 0);
+    });
+
+    it('sets enabledLibraries to the enabledLibarires property in popcode.json', () => {
+      const gistData = {
+        files: {
+          "index.html": 
+            {
+              language: 'html',
+              content: '<h1>Hello, world!</h1>',
+            },
+          "popcode.json":
+            {
+              language: 'json',
+              filename: 'popcode.json',
+              content: '{"enabledLibraries":["jquery"]}',
+            },
+        },
+      };
+      store.dispatch(createProject());
+      const currentProject = getCurrentProject(store.getState());
+      store.dispatch(importProjectFromGist(currentProject.projectKey, gistData))
+      const updatedProject = getCurrentProject(store.getState());
+      assert.equal(updatedProject.enabledLibraries.length, 1);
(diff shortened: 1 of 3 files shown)


src/actions/index.js
```diff
@@ -249,6 +250,14 @@ function importProjectFromGist(projectKey, gistData) {
   };
 }
 
+function parsePopcodeJson(files) {
+  const popcodeJsonFile = find(files, {filename: 'popcode.json'});
+  if (!popcodeJsonFile) {
+    return {};
+  }
+  return JSON.parse(get(popcodeJsonFile, 'content', '{}'));
```
alexpelan: I think this is safe from XSS / people trying to inject bad stuff into JSON since we don't render this anywhere and the libraries are whitelisted, but please try to think of any vectors of attack I haven't thought of.

outoftime:
Yes!!!!!

spec/examples/actions/index.spec.js
```diff
@@ -0,0 +1,75 @@
...
+describe('indexActions', () => {
+  let store;
+
+  beforeEach(() => {
+    store = createApplicationStore();
+  });
+
+  describe('importProjectFromGist', () => {
+
+
```
outoftime: Don’t need this whitespace here?

spec/examples/actions/index.spec.js
```diff
@@ -0,0 +1,75 @@
...
+          "index.html":
+            {
+              language: 'html',
+              content: '<h1>Hello, world!</h1>',
+            },
+        },
+      };
+      store.dispatch(createProject());
+      const currentProject = getCurrentProject(store.getState());
+      store.dispatch(importProjectFromGist(currentProject.projectKey, gistData))
```
outoftime: Hmm, so the fact that we need to pass in a project key here exemplifies my primary discomfort with the testing situation in this PR, namely that `importProjectFromGist` is not part of the action creator module’s public API (until we made it so to test it). This isn’t really the fault of this PR, but rather the fault of some unfinished business with respect to the way the action creators are structured and tested, described in a disorganized fashion below:

* It was a mistake to dump all of the action creators into one big `index` module
* To that end, I’ve been creating new action creators in more narrowly defined modules (e.g. `actions/ui`)
* I’ve also been trying to migrate action creators from `index` into the single-purpose modules
* That refactor has been hand-in-hand with fleshing out the test suite for the Redux store
* To make it easy to track the progress of that refactor, I’ve been maintaining an invariant that action creators are either “tested and in single-purpose modules” or “untested and in index.js”
* The refactor has also exposed a web of tight coupling in the `index` action creators, which makes it non-trivial (but still ultimately good) to extract them
* Further, there is a deep tension between simple action creators and complex ones
  * “Simple” in this case meaning that it takes some inputs and returns an action
  * “Complex” meaning that it interacts with other modules and probably returns a thunk
* Simple actions are easy to test, but tend not to be what we want to expose as a public API—`importProjectFromGist` being a perfect example. We need something that generates a project key for us, but a simple action creator can’t do that (because it has to be a pure function)
* Complex action creators tend to be much harder to test, because they interact with outside modules that generally need to be mocked out
  * E.g. the `bootstrap` action creator that interacts directly with the `Gists` client (as well as the Firebase client) and does a bunch of complex logic
    * Also the `bootstrap` action creator is generally wack. Coordinating those different application bootstrap tasks should be the job of the `Workspace` view controller
    * But we would still need something along the lines of an `importGistFromId` action creator that interacted with the `Gists` client
* So far there is no affordance in the test suite for DI. I’ve been waiting for the refactor to force my hand (e.g. the first time I need to test an complex action creator that interacts with other modules). I’m thinking [bottle](https://github.com/young-steveo/bottlejs)
* Ergo, while the ideal version of this test would test a currently nonexistent `importGistFromId` action creator and use a mocked out `Gists` client that got dependency-injected into the action creators, the changes required to get to that point would be much bigger than the total surface area of this PR as written

So it seems like we have several unappealing options:

1. Merge this as-is, and I will commit to trying to put the structures in place so that this can be tested in a more appealing way ASAP
2. Remove the tests
3. Hold this PR until the test suite can accommodate the structure we want

I lean heavily toward option 1, but I’m interested to hear your thoughts!

spec/examples/actions/index.spec.js
alexpelan: Yeah, will fix at home later. Should we be linting tests too?

alexpelan: (fixed and force-pushed)

src/actions/index.js
outoftime: Good to think about, but yep, I can’t think of anything bad here either.

outoftime:
@alexpelan I will release this this evening!!

PR #206, opened 2016-10-15
Error message when a tag is somewhere it does not belong

outoftime:
For example, the following HTML:

```html
<!doctype html>
<html>
  <body>
    <li>List Item</li>
  </body>
</html>
```

Now generates an error, because the `<li>` isn’t inside a `<ul>`.

image

This involves integrating a new linter, [`html-inspector`](https://github.com/philipwalton/html-inspector), which detects this kind of error.

An interesting challenge here is that `html-inspector` operates on already-parsed DOM trees, rather than on source HTML. For that reason, the error payload does not give us any information about the location of the error in the source; we just get the DOM node that generated the error.

To find the source location, then, we create a `Range` that starts at the beginning of the document and ends just before the offending node.  By counting the number of newlines in the text content of that range, we can find out what line of the source the node came from. By counting the number of characters after the last newline, we get the column within that line.

Well, almost. As it turns out, browsers don’t generate text nodes for whitespace found before the `<head>` element. As far as I can tell, as soon as the browser parses the `<head>`, it keeps all subsequent whitespace text nodes in the DOM.

To work around this, we simply compare the number of newlines in the actual source with the number of newlines in the `outerHTML` of the `documentElement`. The difference is the number of lines that existed in the source but did not make it into the DOM. This approach seems to work reliably in Chrome, Firefox, and Safari.

The other minor challenge was getting `html-inspector` to be a nice module, since it’s primarily designed to run in the browser, and attaches itself to the global context. Happily, this is pretty straightforward with Webpack, using the `import` and `export` loaders.


outoftime:
@alexpelan WDYT?

alexpelan:
lgtm. figuring out that range finding must have been fun.

outoftime:
@alexpelan yes it was. definitely some troughs of despair before the idea of comparing newline counts occurred to me. surprised how well that seems to work actually.

alexpelan:
@outoftime gonna call my autobiography troughs of despair

PR #207, opened 2016-10-15
Acceptance tests for validators

outoftime:
Downloaded solution code for all first-semester ScriptEd projects, and make sure that none throws any errors. Good way to make sure that we don’t change the validator in a way that unintentionally breaks existing projects.

The basic structure is to download individual source files into `spec/data/acceptance`, and then use the script `script/update-validation-acceptance` to merge these into a single JSON file that is then read into tests. There is a test macro that generates an example for each source file and asserts that it is considered valid.


outoftime:
@alexpelan WDYT? This makes me less nervous about adding a completely novel validator ; )

outoftime:
Gonna fire on this bc YOLO / doesn’t actually touch application code

PR #208, opened 2016-10-16
Added compatibility for non-MacOS X developers

jwang1919:
Fixes bugs involved with developing this project on Windows. These fixes should also apply for Linux developers but can't confirm it.

**Bugs Fixed:**
`npm install` will fail when attempting to install fsevents dependency
See: https://github.com/npm/npm/issues/14042

`npm test` will fail because of CRLF linebreaks
Bug looks like this: https://github.com/sindresorhus/matcher/issues/4
Fixed using .gitattributes: https://help.github.com/articles/dealing-with-line-endings/

Also added `*.idea` for IntelliJ developers =)


.gitattributes
```diff
@@ -0,0 +1,6 @@
+* text eol=lf
```
outoftime: cool!

.gitignore
```diff
@@ -2,3 +2,4 @@ static/compiled
 node_modules
 npm-debug.log
 bower_components
+*.idea
```
outoftime: I think it’s best to use a [global excludes file](https://help.github.com/articles/ignoring-files/) to keep out artifacts generated by the editor of your choice—hence no `.swp` etc. in here even though I’m a vim user. Would that work for you?

README.md
```diff
@@ -130,6 +130,7 @@ for all the sordid details.
 * [Alejandro AR](https://github.com/kinduff)
 * [Vaibhav Verma](https://github.com/v)
 * [Alex Pelan](https://github.com/alexpelan)
+* [Jesse Wang](https://github.com/jwang1919)
```
outoftime: Be sure to add yourself to the `package.json` as well!

.gitignore
jwang1919: Sure can do.

README.md
jwang1919: Ahh didn't there was another contributors section!

PR #209, opened 2016-10-16
Graceful handling of closed gist windows

### LINKED ISSUE #263: Graceful handling of closed gist windows
https://trello.com/c/3hzWrcSI/264-graceful-handling-of-closed-gist-windows

If the gist export takes a long time, the student may close the window before it completes (or fails). We want to handle this gracefully:

* If the export completes and the student has closed the window, we display an application error instructing the student to try again.
* If the export fails and the student has closed the window, we don’t try to close the window again.

Also used this as an opportunity to move the meat of the gist export logic out of the `Dashboard` and into the `Workspace`, since the latter as the view controller should take care of (or delegate to the action creators) pretty much anything interesting.

Threw in another small fix which is to always open a new tab for the exported gist, rather than reusing the same tab for successive exports (which is 100% confusing).

Finally, I noticed that there was a pretty useless `_handleNewProject` method on `Dashboard`, which does nothing more than wrap the `onNewProject` prop the `Dashboard` is given, so I unrolled that.

https://github.com/popcodeorg/popcode/pull/209

outoftime:
If the gist export takes a long time, the student may close the window before it completes (or fails). We want to handle this gracefully:

* If the export completes and the student has closed the window, we display an application error instructing the student to try again.
* If the export fails and the student has closed the window, we don’t try to close the window again.

Also used this as an opportunity to move the meat of the gist export logic out of the `Dashboard` and into the `Workspace`, since the latter as the view controller should take care of (or delegate to the action creators) pretty much anything interesting.

Threw in another small fix which is to always open a new tab for the exported gist, rather than reusing the same tab for successive exports (which is 100% confusing).

Finally, I noticed that there was a pretty useless `_handleNewProject` method on `Dashboard`, which does nothing more than wrap the `onNewProject` prop the `Dashboard` is given, so I unrolled that.

Closes #263

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -11,8 +11,7 @@
       "libraries": "Libraries",
       "export-gist": "Export Gist",
       "send-feedback": "Send Feedback"
-    },
-    "anonymous-gist-export": "Are you sure you want to export this gist anonymously? If you sign in before exporting, the gist will be added to your GitHub account."
+    }
   },
   "editors": {
     "helpText": "Type in your $t(languages.__language__) code here."
@@ -26,6 +25,7 @@
     "auth-error": "Something went wrong trying to sign in. Please try again!",
     "empty-gist": "You need some code in your project to export a gist!",
     "gist-export-error": "Something went wrong trying to create that gist. Please try again.",
+    "gist-export-window-closed": "Looks like you closed your gist tab before it finished exporting. Please try again.",
     "gist-import-not-found": "Looks like that gist doesn’t exist. Check the URL and try again."
   },
   "languages": {
@@ -34,7 +34,10 @@
     "javascript": "JavaScript"
   },
   "workspace": {
-    "confirm-unload": "If you leave the page, you will lose your work. To save your work, you must log in using your GitHub account.",
+    "confirmations": {
+      "unload-unsaved": "If you leave the page, you will lose your work. To save your work, you must log in using your GitHub account.",
+      "anonymous-gist-export": "Are you sure you want to export this gist anonymously? If you sign in before exporting, the gist will be added to your GitHub account."
+    },
     "components": {
       "editor": {
         "html": "HTML",
--- a/src/components/Dashboard.jsx
+++ b/src/components/Dashboard.jsx
@@ -1,35 +1,17 @@
 import React from 'react';
-import fs from 'fs';
-import path from 'path';
 import Isvg from 'react-inlinesvg';
-import base64 from 'base64-js';
 import i18n from 'i18next-client';
 import bindAll from 'lodash/bindAll';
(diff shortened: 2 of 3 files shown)


outoftime:
@alexpelan you down?

alexpelan:
I tested this one out because I wanted to see exactly how it worked. At least in my dev environment the gist tab seems to be auto-closing every time.

Not caused by this, but it looks like the 'x' icon for the banner that says "Looks like you closed your gist tab before it finished exporting. Please try again" could be moved up a pixel or 2, it's not quite vertically centered

outoftime:
@alexpelan wait, it’s auto-closing even on what should be a successful export!? definitely not doing that for me—is there an error/uncaught rejection in the console?

alexpelan:
Yeah, there's nothing in the popcode console. And I can't do anything with
the tab that opens, it basically insta-closes

outoftime:
@alexpelan that is crazy!!

Can you sprinkle some breakpoints [in here](https://github.com/outoftime/popcode/blob/eafe4f799ec32d8987bfd701b7c08262cc9062f4/src/components/Workspace.jsx#L310-L334) and see if it sheds any light on it?

Definitely gonna hold this PR until we get to the bottom of it…

outoftime:
@alexpelan also I just made a small but possibly critical change so be sure to pull the latest!

alexpelan:
I can't make any sense of it - don't have a ton more time to look into it tonight, but the api response looks totally normal (the gist still gets created) it just automatically closes as soon as it finishes loading it seems. Doesn't do so on master.

outoftime:
@alexpelan no rush, I’m down to hold this until you have a chance to poke at it further. The actual error this was intended to fix turns out to have nothing to do with gist export at all, so while I still think this is overall an improvement, I feel no urgency about it.

outoftime:
well @alexpelan looks like https://github.com/popcodeorg/popcode/pull/214 points to the cause of what you were seeing (it worked), although I’m still very confused as to why, and why i don’t see it locally

do you access your local instance on something other than `http://localhost:3000` by any chance?

alexpelan:
No, I use localhost:3000

outoftime:
@alexpelan most curious… well, I think I know how to fix this PR for you anyway

outoftime:
@alexpelan I’ve updated this based on my understanding of the problem you were seeing. Mind taking another gander locally?

alexpelan:
I'm still not seeing this fixed, should the fix from #507 be included?

outoftime:
@alexpelan hmm, even the gist window still insta-closes? #507 fixes the pop out, but gist should work here…

alexpelan:
Good now. My bad, it was fixed.

Issue #190, opened 2016-10-05
CSS editor does not support webkit filter/filter


jwang1919:
Popcode reads -webkit-filter and filter as invalid properties.

outoftime:
@jwang1919 so, I have learned some things:

* `filter` is not part of a stable CSS spec. It’s a [W3C working draft](https://drafts.fxtf.org/filters/#FilterProperty)
* It’s [not supported whatsoever by IE](http://caniuse.com/#feat=css-filters)
* Only the [most recent release of Chrome (end of August) started supporting it unprefixed](http://caniuse.com/#feat=css-filters)

With that in mind, do you think Popcode should support it?

jwang1919:
Yes because Cloud9 supports use of the `filter` property.
EX: https://raw.githubusercontent.com/jwang1919/ScriptEdBasketball/master/screenshot.png

Ultimately if we're going to transition students from Popcode to Cloud9, any CSS valid in Cloud9 should be valid in Popcode. My thought is that (in our past experiences) some students like to prototype their code in JSBin before deploying to Cloud9. If we invalidate the `filter` property in Popcode, students might assume it won't work in Cloud9.

outoftime:
@jwang1919 But Cloud9 isn’t a learning tool—it’s just a coding environment. it also “supports” CSS properties like `silliness` and `marshmallows`, in the sense that it doesn’t complain if you use them, right? Ultimately students can use non-standard properties in Cloud9 as long as the browser they’re using supports them, but I think we need to figure out independently where Popcode should draw the line (for instance, I can guarantee lots of students at GWHS are using a version of Chrome that doesn’t support `filter`).

outoftime:
Actually, I stand corrected that Cloud9 treats `filter` and `marshmallows` the same—it does pop up an “invalid property” warning for the latter.

My point remains, though, that depending on what laptop they use that day, students may still find that `filter` does not work for them…

outoftime:
OK, giving it some more thought—the point of Popcode is to notify students of *mistakes* in their code. Using `filter` isn’t a mistake—it might not work in some browsers, but it’s not a mistake. So, Popcode should support it.

jwang1919:
@outoftime The point about the outdated Chrome versions does make sense. I would still argue in favor of `filter` because it's more of a school IT issue than a Popcode issue. Also older versions of Chrome do support the `-webkit-filter` property,

[...]

PR #215, opened 2016-10-19
Band-aid fix for Issue #190

jwang1919:
The current CSS linter will actually accept the `filter` CSS property but will completely invalidate any value against the `filter` property. The `-webkit-filter` property is invalid entirely. Oddly enough when I tweaked the code to accept `-webkit-filter`, ANY value worked on this property, no validations were being done, So instead I focused on adding support only for `filter`.

**Supported values:**
`blur`
`brightness`
`contrast`
`drop-shadow`
`grayscale`
`hue-rotate`
`invert`
`opacity`
`saturate`
`sepia`
`inherit`
`initial`
`unset`

The change will also support the use of multiple filters ie:
`filter: contrast(175%) brightness(3%);`

Still not perfect implementation because it cannot validate if a student enters something random in the parenthesis. Also mashing multiple filters improperly can't be validated ie:
`filter: contrast(17brightness(3%);`


jwang1919:
Also noticed, I should put these pull requests on separate branches... I'll do that next time.

outoftime:
Sick! Yes!

> Oddly enough when I tweaked the code to accept `-webkit-filter`, ANY value worked on this property, no validations were being done

This isn’t necessarily a reason not to do it, though—it’s better to miss errors than to throw errors on valid code (similarly to how this PR will accept invalid properties like `color: contrast(175%)` etc.)

That said, as I said in the Trello card, I’m fine with skipping `-webkit-filter`.

> Also noticed, I should put these pull requests on separate branches... I'll do that next time.

Definitely : )

spec/data/acceptance/css/u4p1.css
```diff
@@ -3,36 +3,36 @@ body{
...
+#about{
+    position:relative;
+    top:100px;
+    left:300px;
+}
+#topics{
+    position:relative;
+    top:300px;
+    left:400px;
+}
```
outoftime: This isn’t at all clear, but the stuff in `spec/data/acceptance` is lifted verbatim from the ScriptEd curriculum; the idea is to make sure that changes to the validations don’t inadvertently break something that ScriptEd thinks of as valid.

But I’d love it if you would add a new test case to the [CSS spec](https://github.com/popcodeorg/popcode/blob/master/spec/examples/validations/css.spec.js)!

src/validations/css/prettycss.js
```diff
@@ -5,9 +5,34 @@ import endsWith from 'lodash/endsWith';
...
+      'drop-shadow\\(|' +
+      'grayscale\\(|' +
+      'hue-rotate\\(|' +
+      'invert\\(|' +
+      'opacity\\(|' +
+      'saturate\\(' +
+      'sepia\\(|' +
+      'inherit|' +
+      'initial|' +
+      'unset'].join(''));
```
outoftime: Hmm, doesn’t seem like you’re getting much out of the `join` here—why not make a multi-element array without the pipes `['blur\\(', 'brightness\\(', …]` and then `join('|')`?

spec/data/acceptance/css/u4p1.css
jwang1919: I only changed spacing. Part of that css file was oddly tabbed out.
I also added the `filter` property on line 22.

Sure I can also add it in the CSS spec.

jwang1919: Nevermind, reverting changes.

spec/examples/validations/css.spec.js
```diff
@@ -30,6 +30,14 @@ describe('css', function() {
     `)
   );
 
+  it('allows valid filter', () =>
+      assertPassesValidation(css, `
```
outoftime: I hate to nitpick here, but the indentation convention in the project is two spaces—here and a few other places you’ve got four. Mind updating? Thanks/sorry!

src/validations/css/prettycss.js
```diff
@@ -5,9 +5,34 @@ import endsWith from 'lodash/endsWith';
...
+      'opacity\\(',
+      'saturate\\(',
+      'sepia\\(',
+      'inherit',
+      'initial',
+      'unset'].join('|'));
+
+function isIncorrectlyRejectedValue(value) {
+  return isIncorrectlyRejectedRadialGradientValue(value) ||
+          isIncorrectlyRejectedFilterValue(value);
```
outoftime: Along the same lines, should be *one* more space indenting this line ; )

jwang1919:
Spaces smaces. I'll make the fixes when I get home tonight.

jwang1919:
Revised my IDE to indent with 2 spaces instead of 4. The reformat code action inadvertently modified other parts of the code that I committed hence the multiple reverting.

src/validations/css/prettycss.js
```diff
@@ -5,9 +5,34 @@ import endsWith from 'lodash/endsWith';
...
+    'drop-shadow\\(',
+    'grayscale\\(',
+    'hue-rotate\\(',
+    'invert\\(',
+    'opacity\\(',
+    'saturate\\(',
+    'sepia\\(',
+    'inherit',
+    'initial',
+    'unset'].join('|'));
```
outoftime: @jwang1919 I am sorry I didn’t notice this before! But I think this expression is a bit too permissive, because it doesn’t have an anchor. For instance, it will allow something like:

```javascript
filter: verysillysepia(
```

Can we put a `^` anchor at the beginning?

outoftime: Also—`PrettyCSS` generally doesn’t support `initial` and `unset`; and I don’t think anyone ever uses them, so I would say let’s drop them here as well

jwang1919: Anchor makes sense. I was first concerned that this would break stacking filter values ie:
`filter: contrast(175%) brightness(3%);`, but I was wrong.

`initial` and `unset` is noted in the MDN filter documentation: https://developer.mozilla.org/en-US/docs/Web/CSS/filter#Syntax. I've never really used them either so I'm okay with dropping them.

PR #507, opened 2016-10-23
Use location.href workaround to work around Chrome closing bug in popout

### LINKED ISSUE #505: Popout webpage no longer works
When students would like to view their creations in full screen that can click on the icon on the top right of the webpage render.

On the current build of Popcode clicking this button will immediately cause the window to close.

outoftime:
It appears that a recent version of Chrome has introduced a regression that causes newly-opened windows to immediately close if they are opened with a data URI. This behavior definitely used to work fine, and it also works as expected in Firefox. [Here is a reduction of the bug](https://outoftime.github.io/presentations/openbug/index.html).

Even more oddly, I do not see this behavior in development, only production. Others have seen it in development, though.

Happily, opening `about:blank` and then immediately setting the `location.href` of the new window to the desired location (data URI) seems to work fine.

Fixes #505

### DIFF
--- a/src/components/Dashboard.jsx
+++ b/src/components/Dashboard.jsx
@@ -12,6 +12,7 @@ import Gists from '../services/Gists';
 import {EmptyGistError} from '../services/Gists';
 import ProjectList from './ProjectList';
 import LibraryPicker from './LibraryPicker';
+import {openWindowWithWorkaroundForChromeClosingBug} from '../util';
 import config from '../config';
 
 const spinnerPage = base64.fromByteArray(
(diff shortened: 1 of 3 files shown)


outoftime:
@jwang1919 @alexpelan WDYT?

jwang1919:
@outoftime Good idea checking for browser compatibility! I didn't think of that before. When I found the original bug, I was using Chrome Canary.

On Chrome Canary and Chrome it's working now.
I checked with IE11 and Microsoft Edge and it's failing for those browsers.

outoftime:
@jwang1919 you mean the code in this PR doesn’t work in IE??

jwang1919:
@outoftime Yes, but it didn't work in a prior commit either.

I think it's IE problem when handling data URI schemes in general: https://stackoverflow.com/questions/7405345/data-uri-scheme-and-internet-explorer-9-errors

PR #509, opened 2016-10-24
Minor update to README.md

jwang1919:
Updated link in road map section to link to Github Project Board since we're not using Trello anymore.
Updated some text that could use some links, such as reference to text editors as well as contact information.


outoftime:
Nice, thanks @jwang1919 !

Issue #510, opened 2016-10-25
Generated HTML page does not have a doctype


outoftime:
The preview document generated from the student’s code does not have a `<!doctype>` declaration. This means it will render in [quirks mode](https://developer.mozilla.org/en-US/docs/Mozilla/Mozilla_quirks_mode_behavior), which is bad!

PR #520, opened 2016-10-26
Force html doctype in generated preview

alexpelan:
I just hardcoded to html doctype, since that's what our html validators force anyway. I couldn't find any way to programmatically get the doctype from the DOMParser output, and my other thought was to grab the current document's doctype (in case times change), but that involved some nasty parsing  (http://stackoverflow.com/a/10162353/265714 )

Fixes: #510

### DIFF
--- a/src/components/Preview.jsx
+++ b/src/components/Preview.jsx
@@ -33,7 +33,7 @@ class Preview extends React.Component {
         breakLoops: isLivePreview,
         nonBlockingAlertsAndPrompts: isLivePreview,
       }
-    ).documentElement.outerHTML;
+    );
   }
 
   _handlePopOutClick() {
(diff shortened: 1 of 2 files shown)


src/components/Preview.jsx
```diff
@@ -34,6 +34,8 @@ class Preview extends React.Component {
         nonBlockingAlertsAndPrompts: isLivePreview,
       }
     ).documentElement.outerHTML;
+
```
outoftime: Nice! But wouldn’t this fit better in the `generatePreview` function?

outoftime: Yup gc, there was no deep intent behind the fact that `generatePreview` was returning a DOM, makes more sense the way you’ve got it now.

Issue #371, opened 2016-10-22
Click target for sidebar arrow is too small


outoftime:
You can actually click anywhere in the sidebar to open/close the dashboard, but only the arrow has a cursor: pointer. The whole dashboard should indicate that it is clickable.

PR #539, opened 2016-11-04
Set cursor: pointer for entire sidebar

alexpelan:
Fixes: #371


alexpelan:
Seemed like low-hanging fruit

PR #542, opened 2016-11-05
Link to exported gist if student closed the window

outoftime:
Gist exports can often take a long time in bad network environments.  Students frequently get frustrated and close the window before the export can complete. We now prevent students from trying again until the first export finishes, to avoid triggering GitHub’s bot-detection algorithms.

However, the previous behavior was simply to show an error message if the window is no longer open when the gist export completes. Instead, show a success message with a link to the gist. This message only appears if the export window is no longer open.


spec/examples/actions/ui.spec.js
```diff
@@ -89,28 +89,30 @@ describe('interfaceStateActions', () => {
...
 
-  describe('applicationErrorTriggered', () => {
+  describe('notificationTriggered', () => {
     beforeEach(() => {
-      store.dispatch(applicationErrorTriggered('some-error'));
+      store.dispatch(notificationTriggered('some-error', 'error'));
     });
 
-    it('sets applicationErrors to contain argument', () => {
+    it('sets notifications to contain argument', () => {
```
alexpelan: Pedantic, but maybe should be 'sets notifications to contain arguments'?

spec/examples/actions/ui.spec.js
```diff
@@ -89,28 +89,30 @@ describe('interfaceStateActions', () => {
...
+  describe('userDismissedNotification', () => {
     beforeEach(() => {
-      store.dispatch(applicationErrorTriggered('some-error'));
-      store.dispatch(userDismissedApplicationError('some-error'));
+      store.dispatch(notificationTriggered('some-error', 'error'));
+      store.dispatch(userDismissedNotification('some-error'));
     });
 
-    it('removes argument from applicationErrors', () => {
+    it('removes argument from notifications', () => {
```
alexpelan: Same here, if you make the change above

spec/examples/actions/ui.spec.js
outoftime: GC, also a good reminder to add an explicit spec for constructing an action with a `payload`

Issue #501, opened 2016-10-22
When error is clicked, editor should scroll to it


outoftime:
When the student clicks an error in the list, the cursor focuses on the location of the error in the editor. However, if the error is scrolled out of view, it will stay that way. We should scroll to the error as well as moving the cursor.

PR #543, opened 2016-11-06
Scroll to line when error is clicked

alexpelan:
ACE exposes a scrollToLine method so this ended up being trivial.

Fixes: #501

### DIFF
--- a/src/components/Editor.jsx
+++ b/src/components/Editor.jsx
@@ -3,6 +3,7 @@ import ACE from 'brace';
 import bindAll from 'lodash/bindAll';
 import get from 'lodash/get';
 import throttle from 'lodash/throttle';
+import noop from 'lodash/noop';
 
 import 'brace/mode/html';
 import 'brace/mode/css';
@@ -71,6 +72,9 @@ class Editor extends React.Component {
       requestedFocusedLine.line,
       requestedFocusedLine.column
     );
+
+    this._scrollToLine(requestedFocusedLine.line);
+
     this._editor.clearSelection();
     this._editor.focus();
     this.props.onRequestedLineFocused();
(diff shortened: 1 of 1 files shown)


src/components/Editor.jsx
```diff
@@ -71,6 +71,9 @@ class Editor extends React.Component {
       requestedFocusedLine.line,
       requestedFocusedLine.column
     );
+
+    this._scrollToLine(requestedFocusedLine.line);
```
alexpelan: I decided to keep this in the focus line method/action rather than creating a separate one because it seems like expected behavior to scroll to the line when focusing the cursor

src/components/Editor.jsx
```diff
@@ -80,6 +83,17 @@ class Editor extends React.Component {
     this._editor.resize();
   }
 
+  _scrollToLine(lineNumber) {
+    const shouldCenterVertically = true;
+    const shouldAnimate = true;
```
outoftime: pro self-documentation!

src/components/Editor.jsx
```diff
@@ -80,6 +83,17 @@ class Editor extends React.Component {
...
+    this._editor.scrollToLine(
+      lineNumber,
+      shouldCenterVertically,
+      shouldAnimate,
+      () => {}
```
outoftime: i assume this is a required argument? might be nicer to pass lodash’s `noop`

src/components/Editor.jsx
outoftime: 💯

src/components/Editor.jsx
alexpelan: Nice, did not know about that one.

alexpelan:
@outoftime made requested changes

Issue #515, opened 2016-10-25
Error message for improperly nested <li> is confusing the students


filapinoaquino:
If `<li>` is not properly nested the error message reads "The `<li>` tag can’t go inside the `<body>` tag."  Maybe we can do a poll to figure out a more helpful error message.

`<!DOCTYPE html>
<html>
    <head>
        <title>Page Title</title>
    </head>
    <body>
        <li></li>
    </body>
</html>`

outoftime:
Word! Here’s what I wrote in another ticket:

> E.g. when there is an `<li>` that isn’t in a `<ul>`/`<ol>`, the error message should say as much, rather than just complaining that it’s inside a `<div>` or whatever.

> I think we would just special-case specific tags that are particular about their parents, and use the current message as the generic one.

alexpelan:
Gonna take a look at this one next

outoftime:
@alexpelan all yours!

PR #545, opened 2016-11-06
Special-case html validation for <li> without <ul>, <ol>, or <menu>

alexpelan:
Fixes: #515

### DIFF
--- a/src/util/arrayToSentence.js
+++ b/src/util/arrayToSentence.js
@@ -1,0 +1,6 @@
+import i18n from 'i18next-client';
+import arrayToSentence from 'array-to-sentence';
+
+export function localizedArrayToSentence(array) {
+  return arrayToSentence(array, {lastSeparator: i18n.t('utility.or')});
+}
--- a/src/validations/html/htmlInspector.js
+++ b/src/validations/html/htmlInspector.js
@@ -2,14 +2,40 @@ import last from 'lodash/last';
 import isNull from 'lodash/isNull';
 import Validator from '../Validator';
 
+const specialCases = {
+  li: {
+    reason: 'invalid-tag-parent',
+    parent: '<ul>, <ol>, <menu>',
+  },
+  title: {
+    reason: 'invalid-tag-parent',
+    parent: '<head>',
+  },
+};
+
 const errorMap = {
-  'validate-element-location': (error) => ({
-    reason: 'invalid-tag-location',
-    payload: {
-      tag: error.context.tagName.toLowerCase(),
-      parent: error.context.parentNode.tagName.toLowerCase(),
-    },
-  }),
+  'validate-element-location': (error) => {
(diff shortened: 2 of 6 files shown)


src/validations/html/htmlInspector.js
```diff
@@ -2,14 +2,36 @@ import last from 'lodash/last';
 import isNull from 'lodash/isNull';
 import Validator from '../Validator';
 
+const specialCases = {
+  li: {
+    reason: 'invalid-tag-parent',
+    parent: '<ul>, <ol>, or <menu>',
```
outoftime: Can we get the “or” of it all into the translation? Maybe `i18next` has a clever way of allowing you to construct a natural-language list out of an array?

src/validations/html/htmlInspector.js
```diff
@@ -2,14 +2,36 @@ import last from 'lodash/last';
 import isNull from 'lodash/isNull';
 import Validator from '../Validator';
 
+const specialCases = {
+  li: {
```
outoftime: Any others come to mind? `<title>` inside `<head>` seems like a good one…

outoftime: (I’m down for that to come in later pull requests though)

src/validations/html/htmlInspector.js
alexpelan: Couldn't figure out an easy way to do this, so I changed it to the more robotic sounding "The <li> tag needs to be inside one of the following tags: <ul>, <ol>, <menu>."

src/validations/html/htmlInspector.js
alexpelan: Added title as a special case as well.

alexpelan:
Rebased and pushed again

src/validations/html/htmlInspector.js
jessewang-arvatosystems: Perhaps `<input>` inside `<form>` as well?

outoftime:
@alexpelan what if we used something like [`arrayToSentence`](https://github.com/shinnn/array-to-sentence), and wrapped it in a function that passed in `options` based on a special key in the translation file?

alexpelan:
I'm a little confused what you're looking for - so like, in Validator._convertErrorToAnnotation, pass a key in the payload to use the 'or' sentence logic?

alexpelan:
@outoftime fixed the node version on npm-shrinkwrap. Should be ready for another look.

src/util/arrayToSentence.js
```diff
@@ -0,0 +1,5 @@
+import arrayToSentence from 'array-to-sentence';
+
+export function arrayToSentenceWithOr(array) {
+  return arrayToSentence(array, {lastSeparator: ' or '});
```
outoftime: Can we pull `' or '` out of the translation file, and call this something like `localizedArrayToSentence`?

Issue #548, opened 2016-11-08
Long url in <img> completely hides HTML/CSS/Javascript panels during error


purplecones:
I came across this issue while working with a student a few days ago. The gist in question that causes this popcode issue can be seen [here](https://gist.github.com/Margauxmantrand/d9fb840273e03d9a523fdf2a3060fccc). 

The cause appears to be the length of url. When the error description is shown, the error pane pushes the HTML/CSS/Javascript until the panes are completely hidden. Perhaps long continuous strings should have a max length and wrap accordingly.

```
<img src:"https://www.google.com/search?q=mihs&safe=strict&espv=2&biw=1632&bih=905&source=lnms&tbm=isch&sa=X&ved=0ahUKEwijrLelzPjPAhVMOz4KHbfTCx0Q_AUICCgD#safe=strict&tbm=isch&q=manhattan+international+high+school&imgrc=jTYtEPRLxG2RAM%3A
```

In this particular case, the student used the google search URL instead of the image URL by mistake. Fortunately we were able to export the project so no work was lost.

![popcode issue](https://cloud.githubusercontent.com/assets/136654/20106709/1ab2439a-a5a4-11e6-933f-eac67dcf1de0.gif)

outoftime:
Yikes! Good catch!

PR #549, opened 2016-11-08
Resolved long word wrap issue in error pane

purplecones:
Added styling to allow long 'words' to wrap in the error pane. This should fix issue https://github.com/popcodeorg/popcode/issues/548.

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -473,6 +473,20 @@ body {
   cursor: pointer;
 }
 
+.errorList-error-message {
+  overflow-wrap: break-word;
+  word-wrap: break-word;
+
+  -ms-word-break: break-all;
+  word-break: break-all;
+  word-break: break-word;
+
+  -ms-hyphens: auto;
+  -moz-hyphens: auto;
+  -webkit-hyphens: auto;
+  hyphens: auto;
+}
+
 .notificationList-notification {
   text-align: center;
   margin-bottom: 0.2rem;


src/css/application.css
```diff
@@ -473,6 +473,20 @@ body {
   cursor: pointer;
 }
 
+.errorList-error-message {
+  overflow-wrap: break-word;
+  word-wrap: break-word;
+
+  -ms-word-break: break-all;
+  word-break: break-all;
+  word-break: break-word;
```
outoftime: I’m not an expert, but I don’t think this is necessary—I think all the [browser versions Popcode supports](https://github.com/popcodeorg/popcode/blob/master/config/browsers.json) can deal with `break-all` just fine?

src/css/application.css
```diff
@@ -473,6 +473,20 @@ body {
   cursor: pointer;
 }
 
+.errorList-error-message {
+  overflow-wrap: break-word;
+  word-wrap: break-word;
```
outoftime: What are your thoughts on *just* using `word-wrap`? As I understand it, `word-wrap` is officially considered an alias for `overflow-wrap`, but has better browser compatibility. Better to keep it simple and just use `word-wrap`, which is widely supported?

src/css/application.css
```diff
@@ -473,6 +473,20 @@ body {
...
+
+  -ms-word-break: break-all;
+  word-break: break-all;
```
outoftime: I don’t think we want this property—won’t it cause words to break even if they’re not one long continuous word? Would rather have normal whitespace wrapping unless it is the specific case you’re targeting here (one very long unbroken “word”) which I think should be handled by `word-wrap`/`overflow-wrap`.

purplecones:
Hi @outoftime, agree with all your points. However, it seems that really the only property that helps us here is `word-break: break-all;`.  `word-wrap: break-word;` alone doesn't fix it.

Would you be open to merging just `word-break: break-all;` ?

outoftime:
@purplecones don’t want you to think I’ve forgotten about this: short answer is, I’m hesitant to use `word-break`, because it will cause the browser to break words all over the place, even when not really needed, in a way that will make error messages harder to read.

My gut is that there is a way to get `word-wrap` to do what we want—maybe combined with `overflow`?—but I need to play around with it a bit. I’m hoping to get to that very soon!

ericandrewlewis:
Glad you two are looking at this 😄 

After reading [the CSS Tricks notes on forcing breaks](https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/) (a nice read), I have a few thoughts.

`word-break: break-word;` fixes the problem in Chrome 54 OS X and Safari 10. It does not fix the problem in Firefox, which makes sense as `break-word` is a non-standard property for `word-break`.

Surprisingly, neither `word-wrap: break-word;` nor `overflow-wrap: break-word;` fix this problem in any browser. I have a feeling this is because of the internals of flexbox's layout algorithm, which for some reason ignores these properties while deciding how line breaks are performed on the contents of a flex element, while `word-break` is taken into account. 

An incomplete solution (won't fix FF or perhaps browsers I didn't test) is to introduce `word-break: break-word;` on `.error-list__error`. We should probably just do this, as it fixes the problem in some cases. `word-wrap` and `overflow-wrap` rules could be added too, but it's not clear what that would do.

We could consider alternative layout for `.environment` or research how flexbox's layout relates to these properties, but those would be more involved tasks.

I think we can ignore hyphenation here. URLs seem to be the common cause of this problem, and adding hyphens inside URLs could confuse users.

Issue #191, opened 2016-10-05
No error prompt on closing curly brace in CSS section


jwang1919:
Popcode will read the following CSS as valid with no error prompt.
`p {
    color: black;
`

outoftime:
I think the fix for this is to incorporate [`stylelint`](https://github.com/stylelint/stylelint), which does report this as an error.

outoftime:
@jwang1919 this would be useful and possibly juicy and fun if you are looking for a meaty contribution!

jwang1919:
I'll work on it the week after. I got other commitments this weekend 😄 .

jwang1919:
@outoftime please assign this to me. I can't find any option to do so myself. I'll start doing some basic digging into this issue.

outoftime:
@jwang1919 all yours!!

outoftime:
Fixed by #556

PR #553, opened 2016-11-12
#191 incorporate stylelint

jwang1919:
Addresses issue: #191 

Adds stylelint to the suite of CSS validators.
stylelint picks up on the errors that PrettyCSS can't detect.

Currently only checks for two scenarios:
1. Missing closing braces
```
p {
    width: 50px;

```
2. Missing semi-colons (with one property defined)
```
p {
    width: 50px
}
```

*Note in this scenario the missing semi-colon is properly detected in PrettyCSS:*
```
p {
    width: 50px
    height: 50px;
}
```

Additional rules can be implemented on an ad-hoc basis.


src/util/minimalStylelint.js
```diff
@@ -0,0 +1,33 @@
...
+
+  isPathIgnored() {
+    return Promise.resolve(false);
+  },
+
+  _getPostcssResult: partial(getPostcssResult, stylelint),
+});
+
+export default (code) => {
+  // Prevent postcssResultCache from caching results
```
outoftime: I’m generally against inline comments—if the code doesn’t speak for itself, it can be made clearer! In this case, I think it’s reasonably obvious what’s going on, but if you think it needs further explanation, you could extract a `preventPostcssResultCaching()` function and call it.

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,65 @@
+import Validator from '../Validator';
+
+const errorMap = {
+  // CSS Syntax Errors
+  'Unclosed block': (error) => {
```
outoftime: Since we can get either errors or lint violations, I think we should namespace the keys here—e.g. `errors/Unclosed block` or `lint/declaration-block-trailing-semicolon`. Then `keyForError()` can construct the key appropriately.

outoftime: (btw I am not attached to the exact namespacing scheme above, it’s just an illustration)

outoftime: Also I think that would remove the need for the comments : )

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,65 @@
...
+    const token = error.source;
+
+    return {
+      reason: 'missing-closing-curly',
```
outoftime: We need translations for these, right?

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,65 @@
...
+      payload: {error: token},
```
outoftime: I’m guessing we probably _don’t_ need a payload here? (E.g. I don’t think we need to pass any interpolation arguments to the translation)

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,65 @@
...
+
+  _getRawErrors() {
+    return System.import('../linters').then(
+      ({stylelint}) => stylelint(
+        this._source
+      ).then(
+        (result) => ({result})
+      ).catch(
+        (syntaxError) => ({syntaxError})
+      )
```
outoftime: This is subtle, but in this situation it’s preferable to pass both success and error callbacks to `then`, like so:

```js
      stylelint(this._source).then(
        (result) => ({result}),
        (syntaxError) => ({syntaxError})
      )
```

The reason is that we don’t want our `catch` to catch rejections generated *in the `then` callback*—we only want to catch rejections from the initial call to `stylelint(this._source)`.

outoftime: Also, I think this may work by something of a coincidence—`_getRawErrors()` is expected to return a promise that resolves to an array, not to an object.

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  _keyForError(error) {
+    // CSS Syntax Error or Custom Rule
+    return error.reason || (error.messages[0] && error.messages[0].rule);
+  }
+
+  _locationForError(error) {
+    // CSS Syntax Error
+    if (error.reason) {
```
outoftime: I think it would be worth extracting an `isSyntaxError` method, and using it both here and in `_keyForError`

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,65 @@
...
+
+    // Custom Rule
+    const customRule = error.messages[0];
+    if (customRule) {
+      return {
+        row: customRule.line - 1, column: customRule.column,
+      };
+    }
+
+    return {row: 0, column: 0};
```
outoftime: Do we expect the code to reach this point?

spec/examples/validations/css.spec.js
```diff
@@ -73,6 +73,15 @@ describe('css', function() {
...
 
+  it('gives a useful error when there is no closing curly brace', () =>
+    assertFailsValidationWith(
+      css,
+      `p {
+        display: block;`,
+      'missing-closing-curly'
+    )
+  );
+
```
jwang1919: I first thought the the missing semicolon test between lines 109-124 already validates for this error. After looking into it again, PrettyCSS seems to only fail this test as long as the missing semicolon isn't placed on the last property declaration.

How should I approach this? Should I simply add another `it()` test or add on to the `context('missing semicolon') piece?

src/validations/css/stylelint.js
jwang1919: Was already defined in en.json so didn't need to touch a thing.

src/validations/css/stylelint.js
jwang1919: We don't. Was following the same procedure as `prettycss.js`.

src/validations/css/stylelint.js
jwang1919: Yup. Had to wrap `result` and `syntaxError` as objects because if I passed them like is, they get treated as arrays and each individual attribute gets passed into `_keyForError`.

src/validations/css/stylelint.js
jwang1919: No, npm test was complaining I wasn't returning anything so had to add it. Do you prefer throwing an error here instead?

jwang1919:
Thanks for the suggestions @outoftime! I made the changes, except for two, one regarding CSS spec and the other regarding the `return {row: 0, column: 0}` part. Please read the comments.

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,57 @@
...
+    super(source, 'css', errorMap);
+  }
+
+  _getRawErrors() {
+    return System.import('../linters').then(
+      ({stylelint}) => stylelint(
+        this._source
+      ).then(
+        (result) => ({result}),
+        (syntaxError) => ({syntaxError})
```
outoftime: As a corollary to the above, I believe this should just read `(syntaxError) => [syntaxError]`. Since we’re only ever going to get a single error, we should just return a one-element array.

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,57 @@
...
+      )
+    );
+  }
+
+  _keyForError(error) {
+    if (isSyntaxError(error)) {
+      return `syntaxError/${error.reason}`;
+    }
+
+    return error.messages[0] && `lintRule/${error.messages[0].rule}`;
```
outoftime: One line that I think will look better once this method is processing a single discrete lint error : )

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    const lintRule = error.messages[0];
+    if (lintRule) {
+      return {
+        row: lintRule.line - 1, column: lintRule.column,
+      };
+    }
+
+    return {row: 0, column: 0};
```
outoftime: I also don’t think we’ll need to handle this degenerate case here anymore—`_locationForError` will only be called with something that is a syntax error, or something that is a discrete lint violation.

jwang1919:
Made the changes. Yep after following changes regarding what `stylelint` returns it made the rest of the code more clean.

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,46 @@
...
+  constructor(source) {
+    super(source, 'css', errorMap);
+  }
+
+  _getRawErrors() {
+    return System.import('../linters').then(
+      ({stylelint}) => stylelint(
+        this._source
+      ).then(
+        (result) => (result.messages),
```
outoftime: Tiny nitpick, but the parens around `result.messages` are superfluous—you need parens around an object literal to disambiguate curly brackets for a function body, but otherwise you can just write `=> result.messages` etc.

src/validations/css/stylelint.js
```diff
@@ -0,0 +1,46 @@
...
+        (syntaxError) => ([syntaxError])
```
outoftime: Same deal with the parens here : )

jwang1919:
Yeah i'm fine with rebasing where do I need to rebase to?

Issue #514, opened 2016-10-25
Make it possible to minimize the preview pane


outoftime:
Particularly when demonstrating code on the board, it might be helpful to have the preview pane minimized to make more space for the editors. We can probably just put a minimize button next to the pop-out button, and have it minimize to the sidebar in the same way the editors currently do.

jwang1919:
While PR #131 is still pending, I'll take a look at this. This looks like a simple task with a big upside.

outoftime:
Fixed by #554

PR #554, opened 2016-11-13
#514 Make it possible to minimize the preview pane

jwang1919:
Addresses issue #514

Matched functionality that hides the HTML, CSS, JS panes to the Preview pane.

Slightly modified editor code so that Preview pane will match page width when HTML, CSS and JS panes are hidden.

### DIFF
--- a/src/components/Output.jsx
+++ b/src/components/Output.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import i18n from 'i18next-client';
 import isEmpty from 'lodash/isEmpty';
 import ErrorList from './ErrorList';
 import Preview from './Preview';
@@ -23,6 +24,7 @@ class Output extends React.Component {
         isValid={this.props.validationState === 'passed'}
         project={this.props.project}
         onClearRuntimeErrors={this.props.onClearRuntimeErrors}
+        onMinimize={this.props.onMinimize}
         onRuntimeError={this.props.onRuntimeError}
       />
     );
@@ -59,6 +61,12 @@ class Output extends React.Component {
   render() {
     return (
       <div className="environment-column output">
+        <div
+          className="preview-label container-label"
+          onClick={this.props.onMinimize}
+        >
+          {i18n.t('workspace.components.preview')}
+        </div>
         {this._renderErrors()}
         {this._renderPreview()}
         {this._renderRuntimeErrorList()}
@@ -74,6 +82,7 @@ Output.propTypes = {
   validationState: React.PropTypes.string,
   onClearRuntimeErrors: React.PropTypes.func.isRequired,
   onErrorClick: React.PropTypes.func.isRequired,
+  onMinimize: React.PropTypes.func.isRequired,
   onRuntimeError: React.PropTypes.func.isRequired,
 };
 
--- a/src/components/Output.jsx
+++ b/src/components/Output.jsx
@@ -24,7 +24,6 @@ class Output extends React.Component {
         isValid={this.props.validationState === 'passed'}
         project={this.props.project}
         onClearRuntimeErrors={this.props.onClearRuntimeErrors}
-        onMinimize={this.props.onMinimize}
         onRuntimeError={this.props.onRuntimeError}
       />
     );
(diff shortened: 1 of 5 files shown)


src/css/application.css
```diff
@@ -329,14 +329,32 @@ body {
...
   display: flex;
   flex-direction: row;
+  height: 100vh;
 }
 
 .environment-column {
   flex: 1 1 50%;
 }
 
+.container-label {
```
outoftime: @jwang1919 Popcode’s CSS uses a slightly modified version of the [Block Element Modifier convention](https://en.bem.info/methodology/quick-start/)—only difference is that it uses single dash `-` instead of double underscore `__` (honestly I should probably just refactor it to use BEM exactly, in order to be more approachable).

Anyway, under BEM, the existence of a `container-label` element implies the existence of a `container` top-level block component, which is not the case.

I think what probably makes the most sense is to define a top-level block component (`label` maybe? seems a little unspecific) that defines the non-positioning-related stuff here (background, color, padding, etc.) and then an `environment-paneLabel` that contains the positioning information (position, z-index, etc.). This follows the [“Mix” technique](https://en.bem.info/methodology/quick-start/#mix) of BEM. WDYT?

src/css/application.css
```diff
@@ -329,14 +329,32 @@ body {
   flex: 1 1 auto;
   display: flex;
   flex-direction: row;
+  height: 100vh;
```
outoftime: I’m curious to hear your reasoning here, but I would have been inclined to put this on the `environment-column` element rather than the `environment` block. No really firm reason for this, just that it makes more sense for elements to “push out” their block, rather than expanding to fill it.

src/components/Workspace.jsx
```diff
@@ -210,6 +211,10 @@ class Workspace extends React.Component {
   }
 
   _renderOutput() {
+    if (includes(this.props.ui.minimizedComponents, 'preview')) {
```
outoftime: Since this is minimizing the output, rather than specifically the preview, I think we should be explicit about that and use the string `'output'` to designate it in the UI state.

src/css/application.css
```diff
@@ -329,14 +329,32 @@ body {
...
+.container-label {
+  position: absolute;
+  bottom: 1em;
+  right: 1em;
+  background-color: black;
+  color: white;
+  z-index: 1;
+  padding: 0.2rem;
+  opacity: 0.2;
+  font-size: 0.8rem;
```
outoftime: We should also specify an explicit font family here—right now the editor labels are inheriting a different font from the preview label, which is making them appear to be a different size.

src/components/Output.jsx
```diff
@@ -23,6 +24,7 @@ class Output extends React.Component {
         isValid={this.props.validationState === 'passed'}
         project={this.props.project}
         onClearRuntimeErrors={this.props.onClearRuntimeErrors}
+        onMinimize={this.props.onMinimize}
```
outoftime: Do we need to pass this prop to `<Preview>`? I don’t think it’s used?

jwang1919: Oops, I forgot to remove that from the latest refactor.

src/components/Workspace.jsx
jwang1919: Sure. I was just following the previous naming convention where the output component was named `'preview'`

src/css/application.css
jwang1919: Okay, assigning it with font family `Inconsolata`

src/css/application.css
jwang1919: If `height: 100vh` is placed in `environment-column` instead, when all containers are closed, the site collapses onto the `POPCODE` sidebar and doesn't look that great.

jwang1919:
@outoftime made the changes. Not sure if it's BEM compliant though. The longer I stared at it, the more confused I got.

What I did was create `environment-column-label` to hold positioning and `label` to hold content styling.

src/css/application.css
outoftime: @jwang1919 OK, I’m convinced—have you checked on the behavior in Firefox and Safari? This does seem like an area where the browsers’ subtly differing flex behavior might bite us…

jwang1919:
Works fine in Firefox. I don't have a Mac so can't test Safari.

outoftime:
@jwang1919 just took it for a spin in Safari—lookin’ good! Here goes!

outoftime:
Actually gonna wait to hear your thoughts on the rebasing question in the other PR : )

jwang1919:
Okay @outoftime, after some wrangling I got the commits squashed!

PR #571, opened 2016-11-27
Show spinner until gist/previous project is loaded

### LINKED ISSUE #346: SVGs are not getting inlined
We use `react-inline-svg`, which is supposed to do what the name would imply. But the browser is loading SVG assets discretely. This is bad for UX, especially on slow connections, in which there will be a long delay before the SVGs actually appear.

### LINKED ISSUE #506: When loading a project from gist, Popcode renders a blank project until the gist loads
We should show a spinner instead.

outoftime:
During application bootstrap, we should not display the editing environment until the correct current project is available. The user-facing logic for this is fairly simple:

1. If we’re importing a gist, wait for the gist to load
2. Otherwise, if we’re logged in, wait for the most recent project we were working on to load
3. Otherwise, initialize a new project

The previous behavior was to always show a new project when the environment first loads, which is quite disorienting on a slow network connection if we’re ultimately loading a gist or a project from Firebase.

Though the behavior is fairly simple, the code changes are substantial, mostly because I wanted to make sure all the code I touched is tested and well-structured. Here’s a “chapter summary” of the changes herein:

* Create a new `bootstrap` action dispatcher, which is its own module, rather than living in `actions/index` (this is part of an overall refactor to have `index` only republish dispatchers from nested modules)
  * One big change is that the `bootstrap` dispatcher sets up a handler to get the initial auth state from Firebase, but then **detaches the handler** and does not listen for further changes to auth state.  Changes to auth state that are initiated by user actions (logging in, logging out) are handled directly via the promise that the Firebase auth/unauth methods return. The advantage here is that we don’t need a single code path that handles all possible scenarios of auth/unauth, which are pretty different during bootstrap vs. as a result of a user action. The disadvantage is that there is some minor code duplication
  * I extracted shared logic related to the auth process wherever possible, in order to minimize the duplication between initial auth state handling and user-initiated auth state change handling
  * I simplified the action flow associated with loading projects into the system. Rather than defining single-purpose actions like `CURRENT_PROJECT_LOADED_FROM_STORAGE` and `PROJECT_IMPORTED`, there is simply a `PROJECT_LOADED` action that is invoked any time a project should be added to the store from an external source. This can be used in combination with the existing `CURRENT_PROJECT_CHANGED` action when the loaded project should also become the current project.
  * The `logIn` and `logOut` action dispatchers are now part of the public API of the `actions` module, and in fact are no longer used internally in that module.
  * Several helper functions and dispatchers are moved from `actions/index` to `actions/projects`. As a corollary, establish the principle that only action dispatchers exported from `actions/index` are part of the public action dispatcher API; nested modules may export functions to be used by other action dispatcher submodules, but those should not be imported from outside the `actions` module. To make this manifest, change all the action specs so that they import from the top-level `actions` module.
* Updated the `Workspace` module to handle the promise returned by `auth`/`unauth` and dispatch the `logIn`/`logOut` actions directly
* Added a throbber which is the Pop “thinking” SVG and the word “Loading”; both subtly expand and contract. It’s pretty neat! And my first time writing a CSS animation
* Overhauled the way we ship SVGs
  * Switch to the `svg-react-loader` rather than `react-inlinesvg`, which is more friendly to the overall build environment
  * Also run inlined SVGs through `svgo`, which compresses them by removing unnecessary cruft
* Write tests for the `bootstrap` action to cover all the cases outlined above
  * This is the first time we’re testing action dispatchers that interact with external dependencies, namely Firebase and the GitHub API. After trying out several options—dependency injection, injection loader, etc.—the only practical solution I could come up with was directly stubbing the live instances of the external clients using Sinon
  * This is not terribly satisfying, but I did build stub harnesses that allow tests to define how they want the clients to behave semantically; this also means that changing to a different method of stubbing in the future wouldn’t necessarily require huge changes to the tests
  * I also moved the test cases for gist import behavior, previously in a test for the `index` module, into the `bootstrap` spec. This also means we can test behavior via a public API
  * I also added a line of code to the spec helper to load the linters into the main bundle, rather than code splitting them. ~~Without this change, we have no way to know when the validations have completed, since it involves an actual web request (via the browser via Karma)~~. Edit: this still doesn’t happen reliably in a CI situation. For now, I’m punting on figuring out how to test the side effects of asynchronous code outside of simply deferred promise resolution handlers.
  * That said, the way we’re dealing with asynchronicity in tests feels fragile. The problem stems from the basic workflow of these tests: dispatch an action; wait for a bunch of promises to resolve and their handlers to complete; check the state of the store. The promises and handlers are generally completely internal to the implementation, so the tests can’t directly introspect on when they have fired. So we resort to just resolving a bunch of promises in series before making any assertions, which is a sort of blunt-instrument way to be fairly confident that all promises have fired.

Fixes #506 
Fixes #346


jwang1919:
I don't understand the Popcode architecture well enough to give a concrete opinion. All I can say is basic smoke testing looks good from my end.

PR #572, opened 2016-11-27
Construct path-matching patterns to be platform-independent

outoftime:
On Windows, we’ll get paths with the `\` segment separator. So, our regexps can’t assume the separator will be `/`. Use `path.join` to construct the partial paths that go into the RegExp; this will also change any `/` passed in the `modulePath` argument to `\`, via the `path.normalize` method.


jwang1919:
Can confirm this is working and fixes the problem that currently exists on master branch.

outoftime:
@jwang1919 yesssss

PR #593, opened 2016-12-05
Retry asynchronous module load if first time fails, take 2

### LINKED ISSUE #541: Loading chunk 0 failed
## Error in Popcode

**Error** in **/**
Loading chunk 0 failed.

[View on Bugsnag](https://app.bugsnag.com/popcode/popcode/errors/581cc26842a024dfe3331480?event_id=581cc268a821fb1000e262b7)

## Stacktrace

    webpack:/webpack/bootstrap efda956c0d34a03e003d:104 - HTMLScriptElement.e

[View full stacktrace](https://app.bugsnag.com/popcode/popcode/errors/581cc26842a024dfe3331480?event_id=581cc268a821fb1000e262b7)

outoftime:
Re-run of #587, which was reverted because of tons of uncaught errors getting thrown in production. Original PR description follows.

---

One of the most common uncaught errors reported for Popcode is that it fails to load the asynchronous module that contains the linter libraries used for validations. To guard against this, we introduce an `importLinters()` function, which has the following safeguards:

* Wrap the `System.import` call in a `promiseRetry`, so we retry loading several times with exponential backoff
* If even after retrying the operation has still totally failed, reject the promise, but also null it out so that, next time we need the linters, we’ll start the process all over again.

Fixes #541

Reverts popcodeorg/popcode#592


outoftime:
Closing in favor of a PR from my branch

PR #606, opened 2016-12-09
Flatten user data structure in store

outoftime:
Until now, we’ve been directly storing the profile information in the `github` key of the Firebase response in the `info` key of our own user representation in the store. This means that knowledge of the particular structure of the Firebase user representation leaks throughout the application.

Now, instead, the user reducer maps the Firebase structure on to a normalized, flat user representation for internal use. Most relevant information is now a key in the top-level user object; the only exception is `accessTokens`, which is a map of provider to access token, in the unlikely(?) event we’ll be supporting authentication with other providers in the future.

This is in preparation for the Firebase SDK upgrade, which will in fact give us a different structure of the Firebase user representation.

Also reads the `uid` from the top-level Firebase response rather than the `auth.uid` key, because this seems more correct (it’s a top-level fact about the Firebase user).

### DIFF
--- a/spec/examples/actions/user.spec.js
+++ b/spec/examples/actions/user.spec.js
@@ -3,6 +3,7 @@
 
 import '../../helper';
 import MockFirebase from '../../helpers/MockFirebase';
+import {createUser} from '../../helpers/MockFirebase';
 import {assert} from 'chai';
 import dispatchAndWait from '../../helpers/dispatchAndWait';
 import buildProject from '../../helpers/buildProject';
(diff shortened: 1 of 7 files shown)


spec/examples/actions/user.spec.js
```diff
@@ -118,28 +110,28 @@ describe('user actions', () => {
...
+          store.getState().user.get('username'),
           userData.github.username
         );
       });
 
-      it('should set imageURL', () => {
+      it('should set avatarUrl', () => {
         assert.equal(
-          store.getState().user.getIn(['info', 'imageURL']),
-          userData.github.imageURL
```
outoftime: oops `imageURL` was never the right key

PR #611, opened 2016-12-12
Change editors and error output containers from flexbox to width-base…

### LINKED ISSUE #548: Long url in <img> completely hides HTML/CSS/Javascript panels during error

ericandrewlewis:
…d layout.

Fixes #548.

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -343,8 +343,6 @@ body {
 .environment {
   box-sizing: border-box;
   flex: 1 1 auto;
-  display: flex;
-  flex-direction: row;
   height: 100vh;
 }
 
@@ -360,6 +358,9 @@ body {
 }
 
 .editors {
+  float: left;
+  width: 50%;
+  height: 100%;
   display: flex;
   flex-direction: column;
   font-family: 'Inconsolata';
@@ -412,6 +413,8 @@ body {
 }
 
 .output {
+  width: 50%;
+  height: 100%;
   display: flex;
   flex-direction: column;
   position: relative;
@@ -486,6 +489,7 @@ body {
   margin: 0;
   padding: 0.5rem 1rem;
   background-color: var(--color-gray);
+  word-wrap: break-word;
 }
 
 .error-list__error {


ericandrewlewis:
After reading a lot about flexbox the other night, I explored using width-based layout instead of flexbox to fix this bug in PR #611. This fixes the bug in Chrome, FF and Safari on my Mac.

The flex sizing algorithm will take long strings of unbroken text into account for figuring out how big a flex item should be. A percent width-based layout will keep the column constraints and work well with a `word-wrap: break-word;` rule.

ericandrewlewis:
> what happens when all the editors are minimized? or the output is minimized?

Ah, I overlooked this feature. The PR breaks this, and constrains both columns to 50% even when one is minimized.

If we'd like to continue down this path, we could set each column's width to 100% by default, and add a sibling selector to constrain them each to 50% when there is a sibling. Glad to amend the PR, or consider different solutions if that sounds awkward.

outoftime:
@ericandrewlewis ooh I think that’s clever! BUT at a higher level, moving from a flex-based to a float-based layout feels like a step backwards (`float` was never intended to do this; `flex` was). I still have a gut feeling that there is a combination of good `flex-basis` and `flex-shrink` values that will make this do the right thing (although if it can be definitively demonstrated that that is not the case, I think your solution is the next-best option for sure)

PR #614, opened 2016-12-14
Introduce an npm script for local development.

### LINKED ISSUE #610: Installing gulp is not mentioned in local dev instructions
The local development instructions say the user should run 

`$ gulp dev`

I didn't have this on my computer and then installed gulp globally via 

`$npm install -g gulp`

ericandrewlewis:
Fixes #610

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ $ yarn
 That'll pull down the dependencies. Then run:
 
 ```bash
-$ npm run dev
+$ yarn run dev
 ```
 
 This will start a local static server, and open it in your browser. The first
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -154,7 +154,7 @@ $ yarn
 That'll pull down the dependencies. Then run:
 
 ```bash
-$ gulp dev
+$ npm run dev
```
outoftime: Can you change this to `yarn run dev`? In theory it should be possible to develop Popcode without using `npm` at all.

ericandrewlewis: Whoops, good point! I've pushed another commit to fix this.

Issue #544, opened 2016-11-06
Allow undefined variables if HTML contains an external script


outoftime:
If students load an external script on to the page by putting a `<script>` tag in the HTML, we can no longer assume that all global variables are known to us. So, in that case, we should turn off the `jshint` check for undefined variables.

This would, for instance, allow Popcode to afford a Tiny Turtle project without explicitly supporting it.

alexpelan:
Think I might take a look at this one next - sound good?

outoftime:
@alexpelan sure!!

PR #619, opened 2016-12-21
Allow undefined variables if external script exists

alexpelan:
Guessing we are going to go back and forth on this a little bit so figured might as well send it your way for some feedback. Big changes:

1) Analyzers, modules that look a lot like validations but are meant for differing application behavior at runtime based on the contents of the Popcode source. It seemed like there are enough variations for custom parsing that we would want this to be generic, I know there's the ticket about changing Popcode title based on HTML document title, plus with this we could allow ignoring certain rules/lines like some linters do (//jshint ignore:line, etc). 
2) validationOverrides - now, we store in the state a list of validation rules that might be on or might be off, and defer to this part of the state when validating, dropping the error if the rule is disabled

Fixes: #544


outoftime:
@alexpelan I’m excited to get this out! Sorry for the long delay in feedback.

bitmoji

High-level, the thing I think we should explore approaching differently is the relationship between the analyzer(s) and the store. In particular, I think of Popcode’s architecture as following these basic principles:

* The store contains a fully normalized representation of the current state of the application
* Actions are dispatched in response to an external event that should change application state: a user action, the results of a database operation, etc
* Reducers change the state of the store in response to actions
* The user interface constructs itself based on the current state of the store, and calculates any second-order information from the store’s contents on the fly

If I’m correctly understanding what you’ve got here, it deviates from these principles because the information in the store is not normalized. In particular, the state of the `validationOverrides` can be fully derived from the state of `currentProject` and `projects`. As a corollary, the `analyzeSource` action is triggered as a second-order effect of an external action (changing a source), rather than being a direct effect of an external event.

The alternative, I think, would be to stick closer to the way `enabledLibraries` are currently passed to the validator, i.e. just figure out what to pass based on the `currentProject` at the time we are initiating validation. Obviously answering that question is pretty transparent in the case of `enabledLibraries`, whereas it definitely requires more calculation in the case of rule-skipping, but ultimately it’s a pure function of the current project in both cases.

In this scenario, I think the Analyzer idea might look like a class that is initiated with a raw project and can answer questions about it. I think it’d probably make sense to also use this for `enabledLibraries`, and probably for the logic that (rather poorly at the moment) generates the project preview used in the project list. Agreed that the same structure would make sense to reuse for using the project `<title>` as the Popcode window title.

I’d also propose moving more validator-specific knowledge into the validator itself, and out of the analyzer, e.g. I think something along the lines of the validator calling `analyzer.hasAnyExternalScripts()` and then deciding what validation rules, if any, to suppress.

An alternative structure might be to explore using [reselect](https://github.com/reactjs/reselect), which I’ve been wanting to integrate into Popcode across the board. That might be a more Flux-y way to create a module that can answer questions about the current project, rather than exporting a class.

Anticipating a pretty glaring counterpoint:

### The store is not fully normalized today! The entire `errors` structure is derived from the current project state.

This is true; the approach you take here is (I’m guessing consciously) analogous to the way we do validations. However, I can think of two reasons why errors should be represented in the store but validation overrides should not.

First, as a practical matter, validations are asynchronous. Some of the underlying linters provide only async interfaces, and all are loaded asynchronously, so this is irreducible. The approach of calculating all derived data in each render cycle only works if those calculations can be done synchronously. Of course, it’s not incredibly satisfying to carve out such a big exception to a core invariant of the data structure based on the fairly arbitrary question of whether derived state can be calculated synchronously.

At a more abstract level, though, one could make the argument that error state is not, in fact, derived purely from project state: instead, error state is a function of project state *and the set of rules in the validators*, the latter of which is external to application state. If you think of the validators as an external system (like a database), we don’t have to think of error state as an exception to the no-derived-state rule. Granted, this is a bit hand-wavy, as error state is still a pure function of current project state, but that pure function is the output of a large, complex, and totally separate subsystem of the application.

(To be honest, I do think there would be a good argument for taking error state out of the store, if it were possible to do all validations synchronously, and also if we used something like reselect to avoid repetitive expensive computations)

Okay! So! What do you think?

alexpelan:
Makes sense, but a little bit confused - it seems you recommend two different directions:

1) Do this like enabled libraries, in that case I'd imagine when we dispatch validateSource we'd pass analyzer.getValidationOverrides(currentProject), which would pass a map of languages to error codes, like what gets stored in the state now. In this case we have HTML source code that affects javascript validation so we'd have to send the whole map to every language.
2) Then you say it should be owned by the validator itself, in which case only the jshint validator would have to know about the overrides for now. It'll know exactly the question to ask. 

I prefer 1), since then other validators get the validation overrides for free should we choose to do similar things elsewhere.

I implemented the analyzers as asynchronous in case one day we wanted to include external tools like the linter, but in retrospect that seems a bit overengineered. 

Just want us to agree on an approach before I write the code, let me know.

outoftime:
@alexpelan sorry for the ambiguity! i would fall firmly in the **2** camp. in particular, as of now, only the JSHint validator itself has any knowledge of what a JSHint error code even *is*, and I think that’s a good thing. further, the validator could decide to change the configuration given to jshint rather than suppressing certain errors (which in fact may be a good idea in this case).

alexpelan:
Okay, take another look.

Now the validator asks the analyzer its question directly, and we send in different configuration to the linter versus maintaining a list of rules to drop / not drop the validations we receive back. This still feels a little bit awkward, mostly because the jshint validator relies on the html source for the analyzer - would it be better for the analyzer to fetch the html source at the moment that it is asked 'containsExternalScript' instead? Not sure how idiomatic it is to reach directly into the store from an otherwise  non-redux-related module like that.

src/validations/javascript/jshint.js
```diff
@@ -21,6 +21,18 @@ const jshintrc = {
   undef: true,
 };
 
+const jshintrcExternalScript = {
```
alexpelan: doing it this way is unsustainable in the long run, but I figured beats having mutable configuration for now?

outoftime:
@alexpelan this is looking great!!

The only meaningful change I would suggest is to remove the `enabledLibraries` parameter to the validation step, and fold that into the `Analyzer`. WDYT?

Also something to consider: have the `Analyzer` expose properties with custom getters, rather than methods, to expose information about the project. E.g., you’d say `analyzer.containsExternalScript` rather than `analyzer.containsExternalScript()`. This would allow the validation step (and other future clients of the `Analyzer`) to just treat it as a plain JavaScript object (perhaps you might even change the param name in the validators to something like `projectAttributes`, a role that just happens to be filled by an `Analyzer` instance). Usefully, it would mean we could pass plain JavaScript objects to the validators in tests.

Then I have some more line-level feedback as well but I want to make sure we’re on the same page about the overall structure before I get picky ; )

outoftime:
Also, in reply to your writeup:

> This still feels a little bit awkward, mostly because the jshint validator relies on the html source for the analyzer - would it be better for the analyzer to fetch the html source at the moment that it is asked 'containsExternalScript' instead?

I don’t think so—I like the idea of validators acting basically as pure functions, blissfully unaware of the higher-level application structure.

> Not sure how idiomatic it is to reach directly into the store from an otherwise non-redux-related module like that.

I’m no expert on Redux idioms but this would definitely be a red flag to me : )

alexpelan:
Okay, rolled enabledLibraries into analyzer and used getters (that made testing 1000x easier, nice idea). Didn't rename from analyzer, just missed that reading over your comment again - but if you feel strongly we can do that as part of line-level feedback.

src/validations/javascript/jshint.js
```diff
@@ -157,6 +169,13 @@ class JsHintValidator extends Validator {
     });
   }
 
+  _getConfig(analyzer) {
```
outoftime: Would be nifty to destructure here

src/validations/javascript/jshint.js
```diff
@@ -139,11 +151,11 @@ const errorMap = {
...
-    this._jshintOptions = defaults(clone(jshintrc), {predef: []});
-    enabledLibraries.forEach((libraryKey) => {
+  constructor(source, analyzer) {
+    super(source, 'javascript', errorMap, analyzer);
+    const jshintConfig = this._getConfig(analyzer);
+    this._jshintOptions = defaults(clone(jshintConfig), {predef: []});
+    analyzer.enabledLibraries.forEach((libraryKey) => {
       if (!(libraryKey in libraries)) {
         return;
       }
```
outoftime: To clarify, my proposal would be:

* Just define the one default `jshintrc`
* In the `_getConfig()` method/function, deep-copy that default object and then mutate the copy according to the project properties, meaning what’s there now plus the logic above.

src/actions/index.js
```diff
@@ -55,10 +57,10 @@ export function saveCurrentProject(state) {
   return false;
 }
 
-function validateSource(language, source, enabledLibraries) {
+function validateSource(language, source, analyzer) {
```
alexpelan: Okay, I renamed usage to projectAttributes while keeping the class name the same, which I think is what you meant.

outoftime: @alexpelan pretty much. To clarify, my proposal was that *in the context of validations*, we think of this as just being a collection of `projectAttributes`, with no particular knowledge/opinion about what is providing that collection of attributes. In contexts where we are actually constructing an analyzer instance (e.g. [here](https://github.com/popcodeorg/popcode/pull/619/files/80d41631fdc3137bad018bb4f6cae9e39bd3b002#diff-4e3037116dbc2bc69a473b7ace3d02abR84)), I think it still makes sense to call the variable `analyzer`.

However my intent definitely isn’t for you to make changes you don’t agree with / don’t make sense. Please let me know if that is the case here!

src/analyzers/index.js
```diff
@@ -0,0 +1,18 @@
+class Analyzer {
+  constructor(project) {
+    this._project = project;
+  }
+
+  get containsExternalScript() {
+    const docElement = this._project.get('sources').get('html');
+    const doc = new DOMParser().parseFromString(docElement, 'text/html');
```
alexpelan: Initialize both in the constructor - this'll get us some savings should we eventually ask any more questions of the html.

src/validations/javascript/jshint.js
alexpelan: Ok, defined changes, use `merge`

src/validations/javascript/jshint.js
alexpelan: Ok, now `_getConfig` calls through to two helpers, one for figuring out jshint configuration and one for mutating based on enabled libraries.

src/validations/javascript/jshint.js
alexpelan: Wasn't quite sure what you meant - I copied into local variables versus referring to the analyzer directly.

alexpelan:
this'll need a rebase to clean up the history before you merge it.

spec/examples/analyzers/html.spec.js
```diff
@@ -0,0 +1,39 @@
...
+    assert.isFalse(analyzer.containsExternalScript);
+  });
+
+  it('containsExternalScript returns true if the html does contain a <script> tag', () => {
+    const currentProject = buildImmutableProject({html: '<script src="https://script.com/script.js">'});
+    const analyzer = new Analyzer(currentProject);
+    assert.isTrue(analyzer.containsExternalScript);
+  });
+
+  it('enabledLibraries', () => {
```
alexpelan: `describe`

spec/helpers/buildImmutableProject.js
```diff
@@ -0,0 +1,11 @@
+import Immutable from 'immutable';
+
+export default function buildImmutableProject(sources = {html: "", css: "", javascript:""}) {
+  return new Immutable.Map({
+          sources: new Immutable.Map({
```
alexpelan: fixed

src/analyzers/index.js
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  get containsExternalScript() {
+    return this._doc.documentElement.innerHTML.includes('</script>');
+  }
+
+  get enabledLibraries() {
+    return this._project.get('enabledLibraries').toJS();
+  }
+
```
alexpelan: sorry, i am a lover of whitespace. it's a hard habit to break

src/validations/javascript/jshint.js
```diff
@@ -139,10 +144,34 @@ const errorMap = {
...
 
 class JsHintValidator extends Validator {
-  constructor(source, enabledLibraries) {
-    super(source, 'javascript', errorMap);
+  constructor(source, analyzer) {
+    super(source, 'javascript', errorMap, analyzer);
+    this._jshintOptions = this._getConfig(analyzer);
+  }
+
+  _getConfig(analyzer) {
```
alexpelan: fixed

src/validations/javascript/jshint.js
```diff
@@ -139,10 +144,34 @@ const errorMap = {
...
+    const containsExternalScript = analyzer.containsExternalScript;
+    const enabledLibraries = analyzer.enabledLibraries;
+
+    const jshintConfig = this._getJshintrc(containsExternalScript);
+
+    let options = defaults(clone(jshintConfig), {predef: []});
+    options = this._setEnabledLibrares(options, enabledLibraries);
+
+    return options;
```
alexpelan: Makes sense to me.

alexpelan:
Ok, I think that should be everything? Let me know so I can rewrite history a little bit

src/analyzers/index.js
```diff
@@ -0,0 +1,18 @@
+class Analyzer {
+  constructor(project) {
+    this._project = project;
+    this._domParser = new DOMParser();
+    const htmlString = this._project.getIn(['sources', 'html']);
+    this._doc = this._domParser.parseFromString(htmlString, 'text/html');
+  }
+
+  get containsExternalScript() {
+    return this._doc.documentElement.innerHTML.includes('</script>');
```
outoftime: I had a suspicion that using a simple `querySelector` rather than `innerHTML.includes` here would be faster, and a [jsperf test](https://jsperf.com/innerhtml-includes-vs-queryselector2) confirms that it is. Given that it’s also probably a bit more idiomatic, I think it might be worth using that approach instead?

src/validations/javascript/jshint.js
```diff
@@ -139,10 +139,20 @@ const errorMap = {
 };
 
 class JsHintValidator extends Validator {
-  constructor(source, enabledLibraries) {
-    super(source, 'javascript', errorMap);
+  constructor(source, analyzer) {
+    super(source, 'javascript', errorMap, analyzer);
+    this._jshintOptions = this._getConfig(
+      analyzer.containsExternalScript,
+      analyzer.enabledLibraries);
```
outoftime: Project style is to put the `)` on a new line (sorry!!!)

src/validations/javascript/jshint.js
```diff
@@ -139,10 +139,20 @@ const errorMap = {
...
+  }
+
+  _getConfig(containsExternalScript, enabledLibraries) {
```
outoftime: Looking great!

alexpelan:
ok, how's it look?

src/analyzers/index.js
```diff
@@ -0,0 +1,18 @@
+class Analyzer {
+  constructor(project) {
+    this._project = project;
+    this._domParser = new DOMParser();
+    const htmlString = this._project.getIn(['sources', 'html']);
+    this._doc = this._domParser.parseFromString(htmlString, 'text/html');
+  }
+
+  get containsExternalScript() {
+    return this._doc.querySelector('script');
```
outoftime: OK sorry to be such an obsessive freak here, but I think this should be a boolean… mind giving it the `Boolean()` treatment?

alexpelan: done

outoftime:
@alexpelan awesome!!

PR #626, opened 2017-01-21
Adds top-level and head-level html validation

gangstertim:
I've added a few rules here based on some of the most common violations I've seen in my class:

1. Only `<head>` and `<body>` elements may exist as direct children of <html>
2. Only `<title>`, `<script>`, `<link>`, and various `<meta>` tags may be children of `<head>`
3. The `<head>` must appear before the body>
4. There may only be one `<head>` and <body> tag per document.

These are all based on the content models for the `<head>` and `<html>` elements described in the w3 spec.

Not totally sure what the protocol is for yarn.lock files, does that belong in the PR?


locales/en.json
```diff
@@ -187,6 +191,6 @@
 
       "set-property-of-null_with-property": "You are trying to set the property __property__ on a null value.",
       "infinite-loop": "There is a loop on this line that looks like it might run forever. Check the loop to make sure it has the right stopping condition."
-    }
+   }
```
outoftime: Is this intentional? I think the previous indentation was correct.

outoftime:
> Not totally sure what the protocol is for yarn.lock files, does that belong in the PR?

Sure does!

outoftime:
@gangstertim I just realized that you added that functionality to `htmllint` yourself—I tip my hat to you!

locales/en.json
gangstertim: whoops! nope, looks like a bug. I'll fix it up.

outoftime:
@gangstertim I can’t wait to merge this! Did you see my previous comment about maybe adding some tests? 😬

gangstertim:
Yep! I have some locally, i'll get 'em up tonight

outoftime:
@gangstertim awesome

outoftime:
@gangstertim very excited to see your commits on this! i notice the build is failing because of a `yarn.lock` error—I saw on Slack that you were having some trouble with your environment, can you check which `yarn` version you’re using? Popcode uses `0.24.6` which is latest stable. Should be available via homebrew if that’s your thing. Happy to help if you run into trouble!

spec/examples/validations/html.spec.js
```diff
@@ -9,9 +9,12 @@ import assertPassesAcceptance from './assertPassesAcceptance';
 
 import html from '../../../src/validations/html';
 
-function htmlWithBody(body) {
+function htmlWithBody(body, head) {
```
outoftime: I don’t think there’s any need for the `head` argument—we don’t use it do we? I’m fine just always having that be hard-coded.

Also, hate to be the bearer of bad news, but… the test suite has moved! Can you rebase/merge master and throw these tests into `test/unit/validations`? I think you’ll find the actual test code quite nice, sorry for the churn though : (

yarn.lock
```diff
@@ -2391,9 +2391,15 @@ domhandler@2.3:
   dependencies:
     domelementtype "1"
 
+<<<<<<< HEAD
```
outoftime: Actually never mind the yarn version thing, looks like you’ve just got some merge conflict in here (I usually just check out `yarn.lock` from master, then re-run `yarn` to generate the right one in this situation).

locales/en/translation.json
```diff
@@ -107,6 +107,14 @@
       "unterminated-open-tag": "You forgot to put a > at the end of your <{{tag}}> tag.",
       "unterminated-close-tag": "You forgot to put a > at the end of your </{{tag}}> tag.",
       "unterminated-comment": "You have a comment that never ends. End the comment with -->",
+<<<<<<< HEAD:locales/en.json
```
outoftime: Also got a merge conflict artifact here

spec/examples/validations/html.spec.js
outoftime: Actually just realized you *did* rebase master, however it looks like your new tests got dropped altogether in the merge

outoftime:
@gangstertim i’d love to get this released before the school year starts—do you think you’ll have a chance to bring it up to date with master and get those tests back in? i can take it over if not, just let me know : )

gangstertim:
I probably will not—I haven't really been developing at all recently and
think it'd take me a while to get back in the swing of things. Thanks!

outoftime:
Got it—no worries, I’ll bring it home from here. Thanks!

Issue #629, opened 2017-01-26
JavaScript doesn’t run at at all when output is hidden


outoftime:
Submitter: Erica Wong
When you are using the console, you need to have the output open for it to run. This is problematic when you are projecting to a screen in front of the whole class. It would be nice if you could close the output so you can zoom in and see everything clearly!

outoftime:
@jwang1919 this is a juicy one! are you down to take a look? probably pretty easily fixed by `display: none`ing the output if it’s hidden, rather than removing it from the DOM entirely…

jessewang-arvatosystems:
@outoftime yeah sure. This doesn't look too bad.

outoftime:
Fixed by https://github.com/popcodeorg/popcode/pull/631

PR #631, opened 2017-01-27
#629 Hide output panel instead of removing it when Output button is clicked

jwang1919:
Addresses Issue #629.

Instead of going straight for `style: display-none`, decided to hide the panel using css class `output_hidden`. The `"_hidden"` string is appended to the class when the Output button is clicked.

### DIFF
--- a/src/components/Output.jsx
+++ b/src/components/Output.jsx
@@ -3,6 +3,7 @@ import i18n from 'i18next-client';
 import isEmpty from 'lodash/isEmpty';
 import ErrorList from './ErrorList';
 import Preview from './Preview';
+import classnames from 'classnames';
 
 class Output extends React.Component {
   _renderErrorList(props) {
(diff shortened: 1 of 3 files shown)


src/components/Output.jsx
```diff
@@ -59,7 +59,7 @@ class Output extends React.Component {
 
   render() {
     return (
-      <div className="environment__column output">
+      <div className={`environment__column output${this.props.isHidden}`}>
```
outoftime: Definitely like the overall approach here—main thing I’d like to see changed is that in [BEM](https://en.bem.info/methodology/naming-convention/), we always keep the top-level component name class (in this case `output`) and then *add* modifier classes (e.g. `output_hidden`). So, in the hidden case, the classnames of this element should be `environment__column output output_hidden`.

You will likely find the `classnames` helper quite helpful for this.

src/components/Output.jsx
```diff
@@ -76,6 +76,7 @@ class Output extends React.Component {
 
 Output.propTypes = {
   errors: React.PropTypes.object.isRequired,
+  isHidden: React.PropTypes.string,
```
outoftime: From the standpoint of the component’s interface, I would expect this to be a boolean (this will probably occur naturally as part of the change I proposed above, but even independently of that change, passing `"_hidden"` as the `isHidden` prop seems like a leaky implementation detail).

jwang1919:
@outoftime Made changes, take a look when you can. I'll be out for Saturday for Chinese New Year, will be back on Sunday the earliest.

PR #635, opened 2017-02-07
Keep project when logging out if locally modified

outoftime:
The following scenario is fairly common:

* Student logs in to Popcode
* Student does some work
* Student then realizes they are logged in to the wrong account (because the browser was logged in to a different GitHub)
* Student “loses” their work because logging out resets the workspace

While the work the student did was, in fact, still available in the account they were logged in to, they may not have access to it, and this is generally an inconvenient situation.

We thus change the logic to only reset the workspace on logout if the current project was not locally modified.

This is particularly important because the forthcoming Firebase upgrade will mean that Popcode sessions are shared between all browser tabs. So, logging out of one tab will log you out of all tabs. This would add a particular edge to the situation described above.

### DIFF
--- a/spec/examples/actions/user.spec.js
+++ b/spec/examples/actions/user.spec.js
@@ -163,8 +163,8 @@ describe('user actions', () => {
       assert.isUndefined(store.getState().getIn(['user', 'id']));
     });
 
-    it('should create a fresh project', () => {
-      assert.notEqual(
+    it('should retain current project', () => {
+      assert.equal(
         getCurrentProject(store.getState()).projectKey,
         loggedInProjectKey
       );
--- a/src/actions/user.js
+++ b/src/actions/user.js
@@ -1,14 +1,13 @@
 import {createAction} from 'redux-actions';
 import identity from 'lodash/identity';
 import {loadAllProjects, saveCurrentProject} from '.';
-import {createProject} from './projects';
 
 export const userAuthenticated = createAction(
   'USER_AUTHENTICATED',
   identity
 );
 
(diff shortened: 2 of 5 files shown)


outoftime:
@alexpelan @jwang1919 @gangstertim @mccluresc WDYT?

jwang1919:
I suggest to keep it simple and persist everything on logout regardless if anything is modified.
It's a bit weird but it helps address the student issue without adding too much complexity. Also doing so is more beginner friendly because it's not entirely obvious that logout === all work gone.

Lastly, I can't imagine a non-student user having to frequently login and logout.

outoftime:
As discussed IRL, here for posterity:

Let’s make two related changes, resulting in an invariant that **changing login state does not affect the project that’s currently on the screen**.

The two specific changes are:

1. When logging in, always keep the current project, even if it’s an unmodified blank project (currently, if you open Popcode and immediately log in, you’ll see the last project you were working on)
2. When logging out, always keep the current project, even if you haven’t modified it locally (i.e. @jwang1919’s proposal above)

This flow has a few advantages:

* It’s simple and easy to anticipate the behavior of the application
* The flow of log in &rarr; realize you’re in the wrong account &rarr; log out is fine, because you’ll just be looking at the same project the whole time
* When we incorporate Firebase 3, if you log in and then open a second Popcode tab, you’ll be logged into that tab too. In this case, we certainly want to open a new project, not the project you have open in the first tab

I’m going to open a separate pull request that makes change 1 above, and then rework this one to reflect change 2.

PR #640, opened 2017-02-10
Always keep current project on login

outoftime:
As the first step in introducing an invariant that changing login state does not affect the current project, remove the logic that under some circumstances changes the current project when logging in. In particular, if no changes had been made to the blank current project before logging in, Popcode would automatically load the last project the user had worked on as the current project.

Now, if you open Popcode and are already logged in, you will always see a blank project. If you log in explicitly, you will always see the project that was on the screen before logging in, even if it is an
untouched blank project.

This simplifies the behavior and code, and makes for an easier mental model of Popcode’s behavior.

See #635 for further discussion and motivation

### DIFF
--- a/TEST_PLAN.md
+++ b/TEST_PLAN.md
@@ -39,8 +39,8 @@ parts that seem relevant to your code changes.
 
 ### Login
 
-* Open a fresh Popcode environment, and log in. Verify that the project that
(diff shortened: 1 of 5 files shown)


outoftime:
@jwang1919 WDYT?

PR #744, opened 2017-03-31
Absolutely position column contents

### LINKED ISSUE #548: Long url in <img> completely hides HTML/CSS/Javascript panels during error

### LINKED ISSUE #742: User Feedback
Submitter: Erica
Email: erica@scripted.org
Student cannot see the HTML, CSS, or JavaScript window panes. She's using Chrome... not sure haven't seen this before!

outoftime:
If a very long token (e.g. a data URI) appears in an error message, [the column will expand to fit its contents, causing it to consume the entire width of the container, and causing the editors to disappear](https://github.com/popcodeorg/popcode/issues/548).

I couldn’t find a completely satisfactory solution using flex all the way down; the interaction between the contents size and the flex properties is not well-documented, but it is certainly sensitive to the size of the content.

By wrapping the contents of each column in an absolutely positioned div that fills its container horizontally and vertically, we can get exactly the behavior we want.  The use of absolute positioning here means that the contents of the columns cannot affect the layout of the columns; it breaks them out of the flow. Instead, the columns’ width is determined solely by the flex arrangement, and the contents are forced to stay within the confines of the columns.

With this in place, a simple `overflow-wrap: break-word` gives us the exact wrapping behavior we want for the long tokens.

image

Fixes #548 
Fixes #742 
Closes #549 
Closes #611

### DIFF
--- a/src/components/Output.jsx
+++ b/src/components/Output.jsx
@@ -62,22 +62,22 @@ class Output extends React.Component {
   render() {
     return (
       <div
-        className={
-          classnames(
-            'environment__column output',
-            {output_hidden: this.props.isHidden},
-          )
-        }
+        className={classnames(
+          'environment__column',
+          {u__hidden: this.props.isHidden},
+        )}
       >
-        <div
-          className="environment__label label"
-          onClick={this.props.onMinimize}
-        >
-          {t('workspace.components.output')}
+        <div className="environment__columnContents output">
+          <div
+            className="environment__label label"
+            onClick={this.props.onMinimize}
+          >
+            {t('workspace.components.output')}
+          </div>
+          {this._renderErrors()}
+          {this._renderPreview()}
+          {this._renderRuntimeErrorList()}
         </div>
-        {this._renderErrors()}
-        {this._renderPreview()}
-        {this._renderRuntimeErrorList()}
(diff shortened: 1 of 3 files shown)


outoftime:
@purplecones @ericandrewlewis relevant to your interests ; )

Issue #719, opened 2017-03-14
Error when functions are declared below function call


megknoll:
Popcode is reporting an error when functions are declared below where they are called.  
It should only enforce that rule for actual var declarations, not for function declarations, which specifically *are* okay to declare below calling.

alexpelan:
Here's a reduced gist: https://gist.github.com/alexpelan/59caa995e2694d41170623365711294f

schraj:
This looks to have to been a design decision by JSLint to consider use of "hoisting" a bad practice. JSHint has added the option to override it if you want to.  

http://jshint.com/docs/options/#latedef

I've added a pull request to override it.

outoftime:
Fixed by #761

PR #761, opened 2017-04-08
add jshint rule to allow function hoisting

schraj:
This fixes issue #719.

This looks to have to been a design decision by JSLint to consider use of "hoisting" a bad practice. JSHint has added the option to override it if you want to.  

http://jshint.com/docs/options/#latedef


outoftime:
@schraj awesome, great discovery! are you down to add a test as well?

also cc @megknoll relevant to your interests

outoftime:
(the test situation is in transition right now btw but `spec/examples/validations` is the place to be for this)

schraj:
ah, of course, sorry should have thought about that!

Issue #369, opened 2016-10-22
Better feedback for enabling/disabling libraries


outoftime:
Currently it’s kind of subtle. We should add a checkmark next to enabled libraries or something.

PR #762, opened 2017-04-08
add checkbox to libraries when they are enabled

schraj:
This addresses issue #369. Let me know if you'd like the look of this changed or tweaked at all.

Also, I don't see a dependency on enzyme or tests for the "jsx" components.  Should I add that and then add some initial tests including one for this?


schraj:
good idea. I'll try that.

src/css/application.css
```diff
@@ -266,6 +266,23 @@ body {
   background-color: var(--color-gray);
 }
 
+.librarypicker__item.dashboard__menu-item_active::before {
```
outoftime: So adding properties to a conjunction of two classes like this, I think, diverges from the spirit of [BEM](http://getbem.com/introduction/), which strenuously avoids any interdependencies between different block definitions.

I would propose instead adding a `dashboard__menu-item_checked` and corresponding `dashboard__menu-item_unchecked`, with the latter just adding the padding. We’d only use those modifiers in the `<LibraryPickerItem>` component.

That would also fix another issue, which is that currently the padding is applied to _all_ dashboard menu items, including ones that never take a checkbox:

image

src/css/application.css
```diff
@@ -266,6 +266,23 @@ body {
...
+    content: '';
+    position: absolute;
+    border-color: #000;
+    border-style: solid;
+    border-width: 0 0.15rem 0.15rem 0;
+    height: .7rem;
+    left: 0.6em;
+    -webkit-transform: rotate(45deg);
```
outoftime: I believe [`autoprefixer`](https://github.com/postcss/autoprefixer#autoprefixer-) should take care of this

src/css/application.css
```diff
@@ -266,6 +266,23 @@ body {
   background-color: var(--color-gray);
 }
 
+.librarypicker__item.dashboard__menu-item_active::before {
+    content: '';
+    position: absolute;
+    border-color: #000;
+    border-style: solid;
+    border-width: 0 0.15rem 0.15rem 0;
```
outoftime: I really like the way this looks, although when I originally suggested using a `:before` pseudo element, I assumed you’d just use the unicode checkbox in the `content` property. I’m totally down with what you’ve got here but also curious as to whether you tried the unicode option and found it wanting?

schraj:
Thanks for the BEM tip.  I haven't looked at that before.

As for the "before" pseudo-element, that is just a way I had done it previously.  Turns out to be simpler using unicode(and looks a little better, I think) so I switched to that.

Issue #517, opened 2016-10-25
Improve the IE error message


outoftime:
Popcode no longer supports IE or Edge at all, but we currently reuse the error message for unsupported browser _versions_, which is thoroughly confusing.

outoftime:
Fixed by #763

PR #763, opened 2017-04-08
add specific error message for ie or edge indicating they aren't supported

schraj:
This fixes issue #517.  Let me know if you'd like other text, etc.

Again, I didn't add tests for this component-level change because the dependencies aren't there, but will add them when a library is chosen.

sorry to bombard you with pull requests : ) Just trying to get more real-world react experience and this is a neat app.


outoftime:
No need to apologize for all the contributions, these look great and are very welcome!

outoftime:
This is lookin good, just tested it out in BrowserStack—will merge as soon as the build passes against the updated branch

PR #765, opened 2017-04-10
Fix broken link in README.md

alexpelan:
Probably related to the recent Github markdown parser changes - the URL and display text were reversed.


outoftime:
Seems more related to me being a dumbo ; )

Issue #769, opened 2017-04-11
does not work when deployed to sub-directory


freder:
URLs in [static/index.html](../tree/master/static/index.html) (and elsewhere) should be relative

outoftime:
@freder what’s the use case?

freder:
@outoftime I'd like to host popcode on `my.own.server/teaching/popcode`

outoftime:
@freder OK, fair enough—you’d set up your own Firebase account etc.?

Anywho, I’d definitely accept a PR that makes running it in a subdirectory work smoothly!

freder:
@outoftime  the main reason to host it myself would be for the students to find all material and resources in one place. another thing I might do is add more libraries, or remove some that I don't talk about in class.

I'm not super familiar with firebase yet. — would you like me to set up a separate account in this case?

outoftime:
@freder yes please—the default is a dev account that I’m totally fine with people using for local development but shouldn’t be used in a live environment (I may reset it at any time). In general I’d suggest looking at the configuration for the WebPack environment plugin for things you’ll need to configure yourself (it’s not a lot):

https://github.com/popcodeorg/popcode/blob/master/webpack.config.js#L174-L182

Feel free to ping me with any questions.

Also if you do move forward with your own deployment, it would be _great_ if you could keep notes on what you needed to do, as a resource for anyone else who wants to do this in the future!

outoftime:
Also, for stuff like adding more libraries, please keep me informed of where the canonical codebase isn’t sufficient. I’m definitely open to adding more to the current list.

outoftime:
Fixed by #770

PR #770, opened 2017-04-12
relative paths

freder:
see #769

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -135,7 +135,7 @@ gulp.task('browserSync', () => {
         {
           lazy: false,
           stats: 'errors-only',
-          publicPath: webpackConfiguration.output.publicPath,
+          publicPath: `/${webpackConfiguration.output.publicPath}`,
         }
       )],
     },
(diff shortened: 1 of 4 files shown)


outoftime:
@freder I don’t have any philosophical objection to the changes here, but it looks like they break the development environment—when I run `yarn run dev`, the application doesn’t properly load, and I see these errors:

image

I would have to imagine this is the result of the change in `webpack.config.js`.

freder:
weird, `yarn run dev` still works for me.

freder:
not entirely sure why, but https://github.com/popcodeorg/popcode/pull/770/commits/f4639fb8e2baad741d51f131bbfcb817bda35432 seems to do the trick.

outoftime:
@freder that makes sense—it offsets the change [here](https://github.com/popcodeorg/popcode/pull/770/commits/381f1de1b4a516226696fd817d885e02040b7a44#diff-11e9f7f953edc64ba14b0cc350ae7b9dR41) from BrowserSync’s standpoint

Issue #368, opened 2016-10-22
Loading indicator while bundle loads


outoftime:
Until the JavaScript bundle has loaded and React takes over, just show a spinner and the word “Loading”.

Also, if you’re signed in, we do another request to Firebase before displaying the editor panes—this should also have a spinner.

outoftime:
Fixed by #771

PR #771, opened 2017-04-14
Show a loading indicator while React is loading

inlinestyle:
Heyo I'm interested in helping out with this project and thought I'd get my feet wet with one of the `good-for-beginner` issues! I chose https://github.com/popcodeorg/popcode/issues/368.

Gif showing the loader when refreshing the page: http://g.recordit.co/oVu9hYkZxd.gif
Screenshot of the loader (on a 13" screen):
image

I've attempted to follow conventions but please let me know if I can make this better in any way. Thanks!

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -254,11 +254,8 @@ body {
 }
 
 .dashboard__menu-item_spinner {
+  background: url('../images/spinner.gif') center/contain no-repeat;
   color: transparent;
-  background-image: url('../images/spinner.gif');
-  background-repeat: no-repeat;
-  background-position: center;
-  background-size: contain;
   cursor: not-allowed;
 }
 
@@ -603,4 +600,11 @@ body {
 
 .u__hidden {
   display: none;
+}
+
+.loading {
+  background: url('../images/large-spinner.gif') center/20vh no-repeat;
+  cursor: not-allowed;
+  height: 100vh;
+  width: 100%;
 }
--- a/static/index.html
+++ b/static/index.html
@@ -8,7 +8,9 @@
     <link rel="icon" type="image/png" href="images/favicon.ico">
   </head>
   <body>
-    <div id="main"></div>
+    <div id="main">
+      <div class="loading"></div>
+    </div>
     <script src="compiled/application.js"></script>
   </body>
 </html>


src/css/application.css
```diff
@@ -604,3 +604,21 @@ body {
...
+
+.loading__spinner {
+  background-image: url('/images/spinner.gif');
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: contain;
+  cursor: not-allowed;
+  height: 100px;
+  width: 100px;
+}
```
outoftime: Looks like your text editor doesn’t automatically add a newline at the end of the file—mind adding one here? Or it might be worth changing your editor settings to always do it since I think that’s generally the preference : )

src/css/application.css
```diff
@@ -604,3 +604,21 @@ body {
...
+
+.loading {
+  align-items: center;
+  display: flex;
+  height: 100vh;
+  justify-content: center;
+  width: 100%;
+}
+
+.loading__spinner {
```
outoftime: Do we need a nested element here? Could we just have the `.loading` block fill the whole container (maybe just absolutely position it) and then put the spinner in the center with `background-position` and a `vh`/`vw` based `background-size`?

src/css/application.css
```diff
@@ -604,3 +604,21 @@ body {
...
+  background-image: url('/images/spinner.gif');
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: contain;
```
outoftime: Assuming there isn’t a disadvantage I’m not thinking of, my preference is to use shorthand properties like `background`

src/css/application.css
```diff
@@ -604,3 +604,21 @@ body {
...
+.loading {
+  align-items: center;
+  display: flex;
+  height: 100vh;
+  justify-content: center;
+  width: 100%;
+}
+
+.loading__spinner {
+  background-image: url('/images/spinner.gif');
```
outoftime: How do you feel about this particular spinner? I don’t hate it, but I feel like it’s a bit small and dark—might be worth trying to spinner style we use in the `output__delayed-error-overlay` element

src/css/application.css
inlinestyle: Done!

src/css/application.css
inlinestyle: Switched to `images/large-spinner.gif`

src/css/application.css
inlinestyle: Done! Also took the liberty of updating `.dashboard__menu-item_spinner`.

src/css/application.css
inlinestyle: Done! I don't usually set that up in favor of following local linter rules. Any feelings on adding something line `css-lint` (which has https://github.com/gajus/css-lint/blob/master/docs/rules/final-newline.md)?

inlinestyle:
@outoftime thanks for the quick turnaround! Agreed with and updated per each suggestion.

outoftime:
> Any feelings on adding something line `css-lint`

@inlinestyle

outoftime:
@inlinestyle this is great! one thing that might be worth thinking about as a followup—animating the spinner in, with a little bit of a delay before it starts appearing at all. someone once told me that showing a spinner for a very short amount of time actually increases perceived waiting time vs. doing nothing at all. still think this is a big improvement though, so going to merge it!

inlinestyle:
@outoftime RE: css linting - I didn't realize `stylelint` was already installed (for browser use). It has a rule for this (https://github.com/stylelint/stylelint/blob/master/lib/rules/no-missing-end-of-source-newline/README.md)!
I'm making a revision applying `stylelint` to the codebase in the `yarn pretest` phase.

Issue #375, opened 2016-10-22
“An <body> tag”


outoftime:
Do the correct “A” vs. “An” in error messages. I suspect `i18next` has an affordance for this.

schraj:
I agree that is seems like i18Next would have something out of the box for this : ) However, it is really more of an idiosyncracy of English so maybe not so much of an oversight on their part.

I don't love this answer, but it is the best I can think of using i18next.  i18next could handle it in the following way using a "format" function as part of the i18n interpolation process.

**Here is the key that you would use:**
"self-closing-non-void-element": "{{tag, en-handle-an-capitalized}} <{{tag}}> tag always needs a closing </{{tag}}> tag. Writing <{{tag}}/> is not allowed.",

**The "Format Keys": "en-handle-an"/"en-handle-an-capitalized" means the following:**
"en"-that this is only appropriate for "en" locale strings
"handle-an"-what it is doing
"capitalized"- that the A or An should/should not be capitalized.

**And here is the i18n init file:**

`import {init} from 'i18next';
import resources from '../../locales';

const getVariationOfAOrAn = function(value, capitalize) {
  const letters = ['a','e','i','o','u','h'];
  let firstLetter = value.substring(0,1);
  let correctWordForm = '';
  if (letters.find(function(l) {
    return firstLetter === l;
  })) {
    correctWordForm = capitalize ? 'An' : 'an';
  } else {
    correctWordForm =  capitalize ? 'A' : 'a';
  }

  return correctWordForm;
}

export default function() {
  init({
    fallbackLng: 'en',
    resources,
    interpolation: {
      format: function(value, format, lng) {
        if (format === 'en-handle-an') return (!lng || lng === 'en') ? getVariationOfAOrAn(value, false) : '';
        if (format === 'en-handle-an-capitalized') return (!lng || lng === 'en') ? getVariationAOrAn(value, true) : '';
        return value;
      }
    }
  });
}
`

outoftime:
@schraj this is a really cool proposal! My only (fairly superficial) feedback:

* I like the idea of including `en-` in the formatter name for self-documentation, but I don’t think we would need the check in the `format` function, since presumably the formatter would only be referenced in English translation files, right?
* Is it possible to compose two formatters? Seems like it would be nice to decouple the capitalization from the a/an of it all

That being said, I think this is a really clever approach and I’d love to go for it. Are you down to cook up a PR along those lines?

schraj:
[...]

PR #773, opened 2017-04-14
Localization formatting and formatters for capitalization and A/An switching

schraj:
This addresses issue #375.  Your idea of allowing format composition is great.  I added the ability to make a format string pipe-delimited, then parse it and apply the formatters sequentially.

The tests for the localization required a couple of new files so let me know if their placement is OK.  I also ran into an issue with loading a JSON file from within the unit tests.  I put the link to what looks to be the associated webpack issue next to that line in the code.  I found a workaround but had to "ignore" an eslint rule to do that.


spec/examples/util/i18n.spec.js
```diff
@@ -0,0 +1,51 @@
+/* eslint-env mocha */
```
outoftime: @schraj I hate to be a Major Payne here, but I’m actually in the process of transitioning everything from Mocha to Tape, and generally I’d like to avoid creating new Mocha specs. Would you mind porting this over to a Tape unit test in the `test/` directory?

spec/examples/util/initTestI18n.js
```diff
@@ -0,0 +1,34 @@
+import {init} from 'i18next';
+import formatFunc from '../../../src/util/i18nFormatting';
+
+// There is an issue when loading json directly that gives an error
+// saying it is not valid javascript.  I believe it may be due to
+// reusing the webpack config object, but I'm not sure.
+// https://github.com/webpack-contrib/json-loader/issues/13
+
+// eslint-disable-next-line import/no-webpack-loader-syntax
+import enTestResourceFile from '!json-loader!./test-translation.json';
```
outoftime: So, why use a JSON file at all? Seems like for the purposes of testing, we can just create the test data in code?

spec/examples/util/initTestI18n.js
```diff
@@ -0,0 +1,34 @@
...
+
+export default function() {
+  // put the locale into the correct namespace for the i18n cache.
+  // Because we are adding it to the cache manually,
+  // we have to give it the hierarchy manually
```
outoftime: I realize this is quite heterodox but I generally like to avoid to inline comments in code—my general philosophy is that good code should be sufficiently self-documenting to not require additional comments. I think in this case you picked a good variable name and it’s pretty clear from that name what the intent is here!

src/util/i18nFormatting.js
```diff
@@ -0,0 +1,30 @@
...
+Arguments:
+value:   The value of the variable that is being passed
+  in to be interpolated
+format:  the pipe-delimited list of formatting flags,
+  it will run the formatting functions sequentially
+  based on the sequence of the flags passed in.
+lng: the language used for the formatting
+NOTE: lng is available(from i18Next in this function) but is not
+currently used, so I've removed it from the arguments list.
+*/
```
outoftime: I’m not philosophically opposed to API doc style comments like this, but I think we should probably use the [JS Doc](http://usejsdoc.org/about-getting-started.html) format?

src/util/i18nFormatting.js
```diff
@@ -0,0 +1,30 @@
...
+export default function(value, format) {
+  let newVal = value;
+  const formatFlags = format.split('|');
+  formatFlags.forEach((flag) => {
+    // replace the value of the input value with our new value("a" or "an")
+    if (flag === 'en-handle-an') {
```
outoftime: Maybe overkill for just two flags, but would it be worth having an object that maps the flag name to a transform function? (Kind of like the [error maps](https://github.com/popcodeorg/popcode/blob/master/src/validations/javascript/jshint.js#L32) in the validation module)

spec/examples/util/i18n.spec.js
schraj: of course, no prob.

spec/examples/util/initTestI18n.js
schraj: LOL.  can't believe that didn't occur to me : )

spec/examples/util/initTestI18n.js
schraj: ok

src/util/i18nFormatting.js
schraj: ok. i'll take a look.

src/util/i18nFormatting.js
outoftime: BTW I don’t have an attachment to JS Doc specifically—there might be another documentation standard that is more widely used at this point—I just want to make sure we’re using whatever will be most familiar to folks (and ideally something that has an `eslint` plugin)

src/util/i18nFormatting.js
```diff
@@ -0,0 +1,24 @@
...
+}
+
+function getVariationOfAOrAn(value) {
+  const vowelishLetters = ['a', 'e', 'i', 'o', 'u', 'h'];
+  const isVowelish = vowelishLetters.find(l => value.substring(0, 1) === l);
+  return isVowelish ? 'an' : 'a';
+}
+
+// eslint-disable-next-line no-unused-vars
+export default function(value, format, lng) {
```
outoftime: Any reason not to just drop the `lng` parameter so we don’t need the eslint annotation?

src/util/i18nFormatting.js
```diff
@@ -0,0 +1,24 @@
...
+  let newVal = value;
+  const inputFlags = format.split('|');
+  inputFlags.forEach((flag) => {
+    if (flag === formatFlags.EN_HANDLE_AN) {
+      newVal = getVariationOfAOrAn(newVal, false);
+    } else if (flag === formatFlags.CAPITALIZE) {
+      newVal = newVal.charAt(0).toUpperCase() + newVal.slice(1);
```
outoftime: I think we can just use Lodash’s `capitalize` here?

src/util/i18nFormatting.js
```diff
@@ -0,0 +1,24 @@
...
+  const vowelishLetters = ['a', 'e', 'i', 'o', 'u', 'h'];
+  const isVowelish = vowelishLetters.find(l => value.substring(0, 1) === l);
+  return isVowelish ? 'an' : 'a';
+}
+
+// eslint-disable-next-line no-unused-vars
+export default function(value, format, lng) {
+  let newVal = value;
+  const inputFlags = format.split('|');
+  inputFlags.forEach((flag) => {
```
outoftime: Particularly combined with the `formatters` proposal below, a `reduce` would be very elegant here.

src/util/i18nFormatting.js
```diff
@@ -0,0 +1,24 @@
+const formatFlags = {
+  EN_HANDLE_AN: 'en-handle-an',
+  CAPITALIZE: 'capitalize'
+}
+
+function getVariationOfAOrAn(value) {
+  const vowelishLetters = ['a', 'e', 'i', 'o', 'u', 'h'];
```
outoftime: We can probably move this into the top level of the module scope to avoid reconstructing it each time the function is called?

Also, maybe use a `Set` so that we can do O(1) lookup? (Obviously this is probably neither here nor there for a six-element collection)

src/util/i18nFormatting.js
outoftime: Also, I’m a fan of still giving default exports a name (so `export default function i18nFormatting(…) {` although maybe I would change the module/function name to something like verb-y like `applyCustomI18nFormatters`)

test/unit/localization/initTestI18n.js
```diff
@@ -0,0 +1,28 @@
+import {init} from 'i18next';
```
outoftime: Couple thoughts here:

* Probably not discoverable enough, but generally I’ve been keeping `test/unit` to just the actual test files, and putting helpers like this in `test/helpers`
* It would be nice to avoid the side-effect of globally configuring `i18next` in the test environment—I think we can set up an isolated instance using [`i18next.createInstance()`](http://i18next.com/docs/api/#create-instance), and then make it the default export of this module?

test/unit/localization/localization.js
```diff
@@ -0,0 +1,58 @@
...
+test(`i18n-should do a/an replace and capitalize the passed variable 
+  when it is passed in with 2 flags`, (p) => {
+    const key = 'key-with-multiple-formats';
+    const options = {tag: 'section'};
+    const expected = 'string with A section';
+    const result = t(key, options);
+
+    p.equal(expected, result);
+    p.end();
+});
```
outoftime: Project style is to have the final newline in all files—I realized that the linter (which I believe catches this) is not running against the `test` directory, so I’ll open up a separate pull request to remedy that.

schraj:
All good suggestions!  I had considered the mapped functions briefly.  Will make the changes now.

schraj:
OK, I believe I've addressed the suggested items.  Lots of great refactors.  I learned a lot implementing them. thanks : )

I didn't exactly understand where this syntax was going(I did add the mapped functions, remove the if statement and add the reduce function), so have a look at what I've done and we can iterate if you like.
`const format = get(formatters, flag, identity)
return format(value);`

src/util/i18nFormatting.js
```diff
@@ -1,24 +1,16 @@
-const formatFlags = {
-  EN_HANDLE_AN: 'en-handle-an',
-  CAPITALIZE: 'capitalize'
-}
+import capitalize from 'lodash/capitalize';
+
+const vowelishLetters = new Set(['a', 'e', 'i', 'o', 'u', 'h']);
+
+const formatters = {
+  'en-handle-an': val => getVariationOfAOrAn(val, false),
```
outoftime: Looks like `getVariationOfAOrAn` just takes a single parameter at this point, so we don’t need the `false`?

src/util/i18nFormatting.js
```diff
@@ -1,24 +1,16 @@
...
-  inputFlags.forEach((flag) => {
-    if (flag === formatFlags.EN_HANDLE_AN) {
-      newVal = getVariationOfAOrAn(newVal, false);
-    } else if (flag === formatFlags.CAPITALIZE) {
-      newVal = newVal.charAt(0).toUpperCase() + newVal.slice(1);
-    }
-  });
-  return newVal;
+export default function applyCustomI18nFormatters(value, format) {
+  return format.split('|').reduce((acc, val) => formatters[val](acc), value);
```
outoftime: So what I was referring to before was to use the lodash function `get`, which is a concise way to retrieve a property from an object with a default fallback if the object doesn’t have that property:

```js
get(formatters, val, identity)(acc)
```

`identity` would just be the Lodash `identity` function, which returns whatever value it is passed.

As written this would throw an error if one tried to pass a formatter that we don‘t recognize—I think it’s probably fine to do that explicitly, or we may just want to use the approach above to fall back to a no-op if there’s no formatter defined.

test/helpers/localizationTest.js
```diff
@@ -0,0 +1,15 @@
+export default (instance, key, options, expected, description) =>
```
outoftime: I like this a lot!

outoftime: Should do this for validation tests too

outoftime:
@schraj do you have a code blog? i think this would make a great post—very elegant solution to what must be a ubiquitous problem.

src/util/i18nFormatting.js
schraj: Nice.  I wasn't aware of either of those lodash functions.  lodash has much more than I had used it for.  And yes, in my formatting changes, I had overlooked the obvious error case of being provided an invalid formatter : )

schraj:
Yeah, I've never done a blog but have thought that it would be a good idea.  That would be interesting to do on this.  Many(most : )) of the interesting ideas in this are yours so if I do write it up I'll definitely credit you!

outoftime:
AND WE ARE GOOD!

Just merged in master, gonna let the tests run, then will get this party on the highway.

In terms of the blog post—you’re too kind!—but what I think is primarily blog-post-worthy is your original idea of using `i18next` formatters in this way. As I recall I even googled around for a decent solution to the “a”/“an” problem in `i18next` and came up with nothing. This is a really elegant approach!

PR #776, opened 2017-04-16
Add stylelinting for internal css

inlinestyle:
This was proposed here: https://github.com/popcodeorg/popcode/pull/771

image

It looks like there are bunch of good plugin options that may be worth installing later:
- https://github.com/stylelint/stylelint-config-standard
- https://github.com/davidtheclark/stylelint-selector-bem-pattern


inlinestyle:
@outoftime I just ran `stylelint` with https://github.com/stylelint/stylelint-config-standard against `application.css`, here's what it came up with:

image

A few of these seem a *little* silly to me... I'll file a revision that adds the standard config that addresses all violations since it's still a very small list, and you can let me know if you want to exempt some or all of the violations.

inlinestyle:
closing & replacing with https://github.com/popcodeorg/popcode/pull/778.

PR #778, opened 2017-04-17
Apply stylelint to internal CSS and use stylelint standard config. Fixes up pre-existing violations.

inlinestyle:
A more ambitious version of https://github.com/popcodeorg/popcode/pull/776. Not quite mutually exclusive, but if this PR is approved https://github.com/popcodeorg/popcode/pull/776 should just be closed.

Installs https://github.com/stylelint/stylelint-config-standard and runs it on Popcode's internal CSS.

The rules are listed here: https://github.com/stylelint/stylelint-config-standard/blob/master/index.js

Here's the list of violations in `application.css` that I fixed:
image

Some of the default rules do seem a little arbitrary to me; it's easy to add and override rules though.


outoftime:
@inlinestyle I like it! Would it be possible to copy the standard config into the project’s `.stylelintrc` rather than extending it? I’m generally a fan of making linter rules explicit.

No objection to any of the actual rules in question / changes herein, though.

inlinestyle:
@outoftime updated with explicit JSON-ified copy of https://github.com/stylelint/stylelint-config-standard/blob/master/index.js

.stylelintrc.json
```diff
@@ -0,0 +1,129 @@
...
+    "shorthand-property-no-redundant-values": true,
+    "string-no-newline": true,
+    "unit-case": "lower",
+    "unit-no-unknown": true,
+    "value-list-comma-newline-after": "always-multi-line",
+    "value-list-comma-space-after": "always-single-line",
+    "value-list-comma-space-before": "never",
+    "value-list-max-empty-lines": 0
+  }
+}
```
outoftime: Is there a linter for the linter config? ; )

inlinestyle: I plead copypasta :(

Issue #313, opened 2016-10-22
Can’t use keyboard shortcuts like cmd-F while in editor window


outoftime:
It seems like ACE captures these—maybe we can turn that behavior off?

PR #783, opened 2017-04-18
Make search-in-code work

inlinestyle:
Fixes https://github.com/popcodeorg/popcode/issues/313

It turns out that ACE comes with a bunch of keyboard shortcuts out of the box, including search. It wasn't working because the necessary extension script wasn't being loaded:
image

Making the script available -> instant search capability!
image

I'd imagine that there are some other great shortcuts that we're erroring out on (here's the basic list of commands: https://github.com/ajaxorg/ace/blob/master/lib/ace/commands/default_commands.js)

### DIFF
--- a/src/components/Editor.jsx
+++ b/src/components/Editor.jsx
@@ -5,6 +5,7 @@ import get from 'lodash/get';
 import throttle from 'lodash/throttle';
 import noop from 'lodash/noop';
 
+import 'brace/ext/searchbox';
 import 'brace/mode/html';
 import 'brace/mode/css';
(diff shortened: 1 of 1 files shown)


outoftime:
> I'd imagine that there are some other great shortcuts that we're erroring out on

@inlinestyle I took a look through that file and I think we might be good? Only a handful of shortcuts require an extension; the ones we don’t already support are:

* Show settings menu
* Go to next/previous error

None of which seem particularly compelling for a classroom use case…

Anyway this is great!

Issue #781, opened 2017-04-17
Chill out spinner/error messages


outoftime:
Currently the spinner and error messages are quite visually intense, and can be very distracting for students who are trying to type code without constantly having their attention pulled to the right half of the screen. A few ideas for making this better:

* Change the spinner to something much less visually distracting. Perhaps just a semi-transparent dark overlay over the preview of the last good state of the page
* Don’t show all the error messages by default. Perhaps the above overlay could just have a chill **2 Errors** message in the middle, which you can then click to expand into full error details. You’d also of course always see the inline error indicators in the editors
* Something more clever like suppressing error messages on the currently focused line in the editor (at least for a longer period of time). This might be hard behavior to get exactly right, though

One thing to note is that the spinner behavior is primarily a UX trick—the idea is to not show error messages immediately as you’re typing. There is no substantial delay before the outcome of code validation is available. The idea of the spinner is to reduce the annoyance of immediately being hit with error messages as you’re typing, but it’s not terribly effective given that students tend to type slowly and pause a lot.

This is extracted from a discussion in https://github.com/popcodeorg/popcode/issues/295

inlinestyle:
@outoftime just a thought (also going back to https://github.com/popcodeorg/popcode/pull/771#issuecomment-294367256) - The loading message shown when loading from a Gist (looks like it's the `<PopThrobber>` component?) is a lot chiller than the spinner. Perhaps we could show that for loading stuff instead?

outoftime:
@inlinestyle definitely agree that the throbber is the chillest, although I think for the “you’re typing and the code doesn’t currently validate” state, it was a mistake to have _any_ sort of motion (my bad). I’m still pretty enamored with the “semi-transparent dark overlay over the preview of the last good state of the page” idea, although it would add a bit of complexity in maintaining the last good state of the page in application state.

As for the loading indicator you introduced in #771—realistically I would love to see the `<PopThrobber>` there but it would possibly be a bit of a pain to implement the throbber without React (and the nice SVG inlining we get for free etc.) so I didn’t want to let the perfect be the enemy of the good : )

inlinestyle:
@outoftime 
1) Makes sense.
2) You do lose the nice pre-made react component, but I don't *think* it's that much of a pain in absolute terms? I believe the pulsing image could be a single CSS class combining `background-image` and `animation`, which could be applied to any given element.

outoftime:
[...]

PR #789, opened 2017-04-20
Statically render "chill" loader while waiting for React

inlinestyle:
Addresses the startup portion of https://github.com/popcodeorg/popcode/issues/781

Here's a recording of the standard page load from popcode.org
chillloaderv1

And here's a recording of a page load of popcode.org with a gist querystring (also shows that the pre-existing chill loader is unchanged)
chillloaderv2

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -566,7 +566,7 @@ body {
   position: absolute;
   top: 50%;
   left: 50%;
-  height: 20%;
+  height: 50%;
   width: 20%;
   margin-left: -10%;
   margin-top: -10%;
@@ -580,6 +580,11 @@ body {
   font-size: 1.5vw;
   padding-left: 1em;
   color: var(--color-dark-gray);
+}
+
+.pop-throbber__image {
+  background: url('../images/pop/thinking.svg') center no-repeat;
+  height: 100%;
 }
 
 .pop-throbber path {
@@ -608,11 +613,4 @@ body {
 
 .u__hidden {
   display: none;
-}
-
-.loading {
-  background: url('../images/large-spinner.gif') center/20vh no-repeat;
-  cursor: not-allowed;
-  height: 100vh;
-  width: 100%;
 }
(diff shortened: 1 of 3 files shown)


static/index.html
```diff
@@ -9,7 +9,10 @@
   </head>
   <body>
     <div id="main">
-      <div class="loading"></div>
+      <div class="pop-throbber">
+        <div class="pop-throbber__message">Loading…</div>
+        <div class="pop-throbber__image--thinking"></div>
```
outoftime: This is annoying, but technically BEM requires that any modifier have a corresponding non-modified block/element. So for full compliance this should also have a class `pop-throbber__image`, even if that class isn’t referenced in the CSS. For simplicity, though, I’d suggest just getting rid of the modifier altogether, and making it implicit that the throbber’s image is a thinking-state Pop (which seems reasonable to assume).

src/css/application.css
```diff
@@ -566,7 +566,7 @@ body {
   position: absolute;
   top: 50%;
   left: 50%;
-  height: 20%;
+  height: 50%;
```
outoftime: Does this have any effect on the way it renders in the Gist import case? (Visual inspection would suggest no, but I can’t figure out why!)

src/css/application.css
```diff
@@ -582,6 +582,11 @@ body {
   color: var(--color-dark-gray);
 }
 
+.pop-throbber__image--thinking {
+  background: url('../images/pop/thinking.svg') center no-repeat;
```
outoftime: It would be crazy to inline this, right?

inlinestyle: Not that crazy? Especially if it were done cleanly, like if it was required as a template via ejs or something. I'd save it for a future PR though.

src/css/application.css
inlinestyle: Not that I could tell? I'd manually verified in dev tools that this property had no effect on display for Gist import. Are there other UI configurations that this is used for?

static/index.html
inlinestyle: Done!

PR #794, opened 2017-04-23
Change preview text

pwjablonski:
Hey Popcode Team! I'm trying to learn more about react by trying to navigate Popcode. This is my first time trying to contribute open source so pardon my mistakes in advance. 

I've noticed myself having difficulty keeping track of / naming / navigating old projects. One thought is using the previewHead text rather than previewBody in the project preview dashboard as way name/organize projects. This however may be confusing from a student learning perspective since this is not the intended function of the <title> tag. Thoughts?

**Suggested change looks like this**

screen shot 2017-04-23 at 2 29 52 pm

screen shot 2017-04-23 at 2 31 37 pm


outoftime:
@pwjablonski from a behavior standpoint this is definitely better than what Popcode currently does. And I actually do think, semantically, the `<title>` tag is a pretty good fit here.

It may look goofy for projects where the students have not changed the title, but hopefully any time we give them starter code the starter has a good title, and again, couldn’t possibly be worse than the way it works now.

One thing—the text content of the entire `<head>` tag is a bit over broad—this would contain, for instance, the contents of an internal stylesheet (if such a horror were perpetrated in Popcode). You should be able to access `previewDocument.title` directly to just pull the text contained in the `<title>` tag.

Issue #302, opened 2016-10-22
Resizable editor windows


outoftime:
Make it possible to resize editor windows by dragging them. Also make it possible to drag the center divider to change the width of the two main columns.

This seems like it should be a pretty straightforward feature, but I am honestly not sure of the right way to implement it given the current Flexbox-based layout!

inlinestyle:
This is an interesting one... I don't know of a way to do it other than reverting to fixed `%` heights which would be painstakingly controlled at the `<Workspace>` level.

outoftime:
@inlinestyle i was thinking the same thing but… [maybe not???](http://codepen.io/chriscoyier/pen/pJpbWK)

(i would prefer to avoid introducing jQuery but the conceptual approach would probably still work?)

outoftime:
`<DraggableCore>` from https://github.com/mzabriskie/react-draggable looks like it would be just the thing for abstracting the actual mouse events into a drag

inlinestyle:
@outoftime that jQuery UI solution is clever! And should port reasonably well to React. 

Do you have a timeframe for working on this? I'd take a stab at it tonight or tomorrow if you're not planning on doing it yourself in the near future.

outoftime:
@inlinestyle it’s all yours! my focus right now is on making some architectural changes that I’m hoping will make the codebase more approachable for new contributors, and also make the application faster.

(speaking of which, one thing I’m a little concerned about is whether having drag events go through the entire Redux lifecycle is going to feel sluggish… if so, I definitely have some optimizations in mind)

outoftime:
@inlinestyle btw, if you do start working on it, would you mind just assigning yourself the issue so no one else tries to grab it?

(Also let me know if that’s not possible—seems like as a popcode org member you should be able to, but you’re not coming up on the list when I click the assign button so maybe I need to do something else)

inlinestyle:
@outoftime Hmm I'm not able to assign myself to any issues in this repo. According to [this](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) I should see a checkbox next to issues, which I don't...

![image](https://cloud.githubusercontent.com/assets/1098749/25161110/1700482e-2488-11e7-860f-0757865bd729.png)

^my interpretation:
- *you* as a write-access member should be able to assign to *me* as a read-access member
- only write-access members can assign issues.

outoftime:
[...]

PR #802, opened 2017-04-28
Move <Workspace>._renderEditors into new <EditorsColumn> component

inlinestyle:
As suggested in https://github.com/popcodeorg/popcode/issues/302#issuecomment-297600122.

This is mostly cut/copypasta from `Workspace.jsx`, with some minor tweaks (using destructuring and `connect` a bit more aggressively).

@outoftime I'm going to file a PR (probably this weekend) against this branch (or `master` if this gets merged first) with an updated version of the demo also mentioned in https://github.com/popcodeorg/popcode/issues/302#issuecomment-297600122.

### DIFF
--- a/src/components/EditorsColumn.jsx
+++ b/src/components/EditorsColumn.jsx
@@ -1,0 +1,77 @@
+import React from 'react';
+import isEmpty from 'lodash/isEmpty';
+import includes from 'lodash/includes';
+import partial from 'lodash/partial';
+
+import EditorContainer from './EditorContainer';
+import Editor from './Editor';
+
+function allErrorsFor(language, errors, runtimeErrors) {
+  if (language === 'javascript') {
+    return errors.javascript.items.concat(runtimeErrors);
+  }
+  return errors[language].items;
+}
+
+export default function EditorsColumn({
+  currentProject,
+  errors,
+  onComponentMinimize,
+  onEditorInput,
+  onRequestedLineFocused,
+  runtimeErrors,
+  ui,
+}) {
+  const editors = [];
+  ['html', 'css', 'javascript'].forEach((language) => {
+    if (includes(ui.minimizedComponents, `editor.${language}`)) {
+      return;
+    }
+
+    editors.push(
+      <EditorContainer
+        key={language}
+        language={language}
+        source={currentProject.sources[language]}
+        onMinimize={
+          partial(onComponentMinimize, `editor.${language}`)
+        }
+      >
+        <Editor
+          errors={allErrorsFor(language, errors, runtimeErrors)}
+          key={language}
+          language={language}
+          percentageOfHeight={1 / editors.length}
+          projectKey={currentProject.projectKey}
+          requestedFocusedLine={ui.editors.requestedFocusedLine}
+          source={currentProject.sources[language]}
+          onInput={partial(onEditorInput, language)}
+          onRequestedLineFocused={onRequestedLineFocused}
+        />
+      </EditorContainer>,
+    );
+  });
+
+  if (isEmpty(editors)) {
+    return null;
+  }
+
+  return (
+    <div className="environment__column">
+      <div className="environment__columnContents editors">{editors}</div>
+    </div>
+  );
+}
+
+EditorsColumn.propTypes = {
+  currentProject: React.PropTypes.object.isRequired,
+  errors: React.PropTypes.object.isRequired,
+  runtimeErrors: React.PropTypes.array.isRequired,
+  ui: React.PropTypes.shape({
+    editors: React.PropTypes.object.isRequired,
+    minimizedComponents: React.PropTypes.array.isRequired,
+  }).isRequired,
+  onComponentMinimize: React.PropTypes.func.isRequired,
+  onEditorInput: React.PropTypes.func.isRequired,
+  onRequestedLineFocused: React.PropTypes.func.isRequired,
+};
--- a/src/components/Workspace.jsx
+++ b/src/components/Workspace.jsx
@@ -5,7 +5,6 @@ import {connect} from 'react-redux';
 import values from 'lodash/values';
 import bindAll from 'lodash/bindAll';
 import includes from 'lodash/includes';
-import isEmpty from 'lodash/isEmpty';
 import isNull from 'lodash/isNull';
 import partial from 'lodash/partial';
 import sortBy from 'lodash/sortBy';
@@ -50,8 +49,7 @@ import {
 
 import {getCurrentProject, isPristineProject} from '../util/projectUtils';
 
-import EditorContainer from './EditorContainer';
-import Editor from './Editor';
+import EditorsColumn from './EditorsColumn';
 import Output from './Output';
 import Sidebar from './Sidebar';
 import Dashboard from './Dashboard';
@@ -141,15 +139,6 @@ class Workspace extends React.Component {
     }
   }
 
-  _allErrorsFor(language) {
-    if (language === 'javascript') {
-      return this.props.errors.javascript.items.
-        concat(this.props.runtimeErrors);
-    }
-
-    return this.props.errors[language].items;
-  }
-
   _handleComponentMinimized(componentName) {
     this.props.dispatch(minimizeComponent(componentName));
   }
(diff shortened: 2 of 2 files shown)


src/components/EditorsColumn.jsx
```diff
@@ -0,0 +1,110 @@
...
+EditorsColumn.propTypes = {
+  actions: React.PropTypes.object.isRequired,
+  currentProject: React.PropTypes.object.isRequired,
+  errors: React.PropTypes.object.isRequired,
+  runtimeErrors: React.PropTypes.array.isRequired,
+  ui: React.PropTypes.object.isRequired,
+  onComponentMinimize: React.PropTypes.func.isRequired,
+};
+
+function mapDispatchToProps(dispatch) {
```
outoftime: Hey, so—using `mapDispatchToProps` generally is a definite :+1: that I’d like to do in `Workspace` as well, BUT the general structure of the components is currently that `Workspace` is the only component that interacts directly with the Redux store, and everything else pushes information up the component tree via callbacks.

I’m definitely open to arguments for an alternative structure in general, but absent a big rethink, I’d like to keep it consistent.

WDYT?

inlinestyle: @outoftime I hear you. I went this route for the sake of moving as much stuff out of `Workspace.jsx` as possible, but I can move the dispatches back and use callbacks. 

I could also put a `mapDispatchToProps` in `Workspace.jsx` while I'm at it...

As far as alternative structure... we could follow the "Containers" convention and move `Workspace` and `EditorsColumn` into a `containers/` directory reserved for components that are `connect`ed. This is a pretty minimal change (just moving two files and changing some imports) but I think people find it makes it easier to disambiguate top-level from display components.

outoftime:
> I hear you. I went this route for the sake of moving as much stuff out of `Workspace.jsx` as possible, but I can move the dispatches back and use callbacks.

Great—I think that would be the move for the purposes of this PR.

> I could also put a `mapDispatchToProps` in `Workspace.jsx` while I'm at it...

I’m not mad at that idea!

> As far as alternative structure... we could follow the "Containers" convention and move `Workspace` and `EditorsColumn` into a `containers/` directory reserved for components that are `connect`ed. This is a pretty minimal change (just moving two files and changing some imports) but I think people find it makes it easier to disambiguate top-level from display components.

Interesting! I’m not familiar with that convention—can you point me to a good writeup?

That said, I think if we did have multiple connected components, I wouldn’t anticipate them being `Workspace` and `EditorsColumn`—seems like it would make sense to have the formation of connected components existing at roughly the same level in the component hierarchy, and in particular to avoid having a connected component as the child of another connected component. So I would think maybe something along the lines of `Dashboard`, `EditorsColumn`, and `Output` as the containers? WDYT?

inlinestyle:
@outoftime 
> I think that would be the move for the purposes of this PR.

I'm going to move the `dispatch`es back to `Workspace.jsx` then.
> can you point me to a good writeup?

There are a bunch of hits for "redux containers" on google, but [this](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) seems to be the original proposal. There's a lot of good stuff, but if you ctrl-f for `"connect"` you'll find the relevant bit (6th bullet point in containers section).

> seems like it would make sense to have the formation of connected components existing at roughly the same level in the component hierarchy, and in particular to avoid having a connected component as the child of another connected component.

Dan Abramov actually says it's chill to nest containers (second bullet point in containers section). That said, `<Workspace>` (and this iteration of `<EditorsColumn>`) don't strictly adhere to the containers convention anyway, and I don't necessarily want to shove it into this project. `Dashboard`, `EditorsColumn`, and `Output` definitely make sense to me as containers, with or without `Workspace` being one as well.

inlinestyle:
@outoftime updated to remove `connect` from `<EditorsColumn>`!

src/components/EditorsColumn.jsx
```diff
@@ -0,0 +1,79 @@
...
+  _allErrorsFor(language) {
+    if (language === 'javascript') {
+      return this.props.errors.javascript.items.
+      concat(this.props.runtimeErrors);
+    }
+
+    return this.props.errors[language].items;
+  }
+
+  render() {
```
outoftime: So what are your thoughts on stateless components? Mine would be something along the lines of “use stateless components unless you need component lifecycle hooks”, but that’s a strong opinion weakly held.

src/components/EditorsColumn.jsx
```diff
@@ -0,0 +1,79 @@
...
+import partial from 'lodash/partial';
+
+import EditorContainer from './EditorContainer';
+import Editor from './Editor';
+
+class EditorsColumn extends React.Component {
+  _allErrorsFor(language) {
+    if (language === 'javascript') {
+      return this.props.errors.javascript.items.
+      concat(this.props.runtimeErrors);
```
outoftime: Nitpick but project style has this indented one level deeper than previous.

Also, probably a battle for another day, but this method has always felt bolted-on. Maybe something I can handle in the upcoming `reselect` refactor…

src/components/EditorsColumn.jsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  render() {
+    const {
+      currentProject,
+      onComponentMinimize,
+      onEditorInput,
+      onRequestedLineFocused,
+      ui,
+    } = this.props;
```
outoftime: Additional advantage of stateless: this destructuring happens in the params. Very satisfying!

src/components/EditorsColumn.jsx
```diff
@@ -0,0 +1,79 @@
...
+      </div>
+    );
+  }
+}
+
+EditorsColumn.propTypes = {
+  currentProject: React.PropTypes.object.isRequired,
+  errors: React.PropTypes.object.isRequired,
+  runtimeErrors: React.PropTypes.array.isRequired,
+  ui: React.PropTypes.object.isRequired,
```
outoftime: Would it make more sense to only pass the parts of the `ui` state that this component needs?

outoftime: (Maybe not)

src/components/EditorsColumn.jsx
inlinestyle: Will fix... with more destructuring...

src/components/EditorsColumn.jsx
inlinestyle: My opinions aren't super strong, but they boil down to:
- If I'm using Flow, use the `class` version because the speccing of `props` looks a little nicer than with stateless IMHO
- Otherwise, use stateless if there are no non-static methods (other than `render`). As an example, `_allErrorsFor` currently depends on `this.props`, so that would tilt me towards using `class`.

Of course `_allErrorsFor` could be refactored to be `this`-independent, which is what I'd do if you want to go stateless.

outoftime: Gotcha. Yeah, my inclination would be to go with an `allErrorsFor` function, which has the appeal of module encapsulation as well.

src/components/EditorsColumn.jsx
inlinestyle: I'd argue that in the absence of static typing getting more granular is a good thing; either splitting `ui` in the props here into `minimized` and `editors.requestedFocusLine`, _or_ switching `ui` from `PropTypes.object` into `PropTypes.shape`.

inlinestyle:
@outoftime made the following changes:
- switched to stateless component
- refactored `_allErrorsFor` method into `allErrorsFor` function
- switched `ui` prop from `PropTypes.object` to `PropTypes.shape`

src/components/EditorsColumn.jsx
```diff
@@ -0,0 +1,77 @@
...
+        source={currentProject.sources[language]}
+        onMinimize={
+          partial(onComponentMinimize, `editor.${language}`)
+        }
+      >
+        <Editor
+          errors={allErrorsFor(language, errors, runtimeErrors)}
+          key={language}
+          language={language}
+          percentageOfHeight={1 / editors.length}
```
outoftime: I think this is why editor resizing doesn’t work right. (No need to fix here)

src/components/EditorsColumn.jsx
```diff
@@ -0,0 +1,77 @@
...
+}
+
+EditorsColumn.propTypes = {
+  currentProject: React.PropTypes.object.isRequired,
+  errors: React.PropTypes.object.isRequired,
+  runtimeErrors: React.PropTypes.array.isRequired,
+  ui: React.PropTypes.shape({
+    editors: React.PropTypes.object.isRequired,
+    minimizedComponents: React.PropTypes.array.isRequired,
+  }).isRequired,
```
outoftime: This is cool—I’m hoping to move to passing `Immutable` props all around rather than converting to JS objects on every state transition. Happily, [it looks like `react-immutable-proptypes` does support an equivalent to `shape`](https://github.com/HurricaneJames/react-immutable-proptypes), and also I’m very interested in making use of [`Immutable.Record`](https://facebook.github.io/immutable-js/docs/#/Record).

outoftime:
BTW @inlinestyle I read the Medium article on containers—great food for thought. The current structure reflects a sort of degenerate form of that pattern by making the `Workspace` the only container (or “view controller” as the React docs put it), but it doesn’t follow the convention fully because it does mix in presentational components.

The container/component dichotomy reminds me of one of the (only) things I like about Angular, which is that components and their templates are (usually) separate files.

Anyway gonna give it more thought in terms of introducing intermediate containers—it seems to dovetail well with some other refactors I’m hoping to tackle soon, namely:

* Break up big ball of mud components like `Workspace` and `Dashboard` (metonym/real: no more `_renderX` methods in components)
* Pass Immutable props throughout the component tree rather than converting to JS
* Use `reselect` and selectors to retrieve data from the store (this I think would go *really* well with multiple containers since my read of `reselect` is that the question it answers really well is “here is the root state, get me a particular piece of information”)

src/components/EditorsColumn.jsx
inlinestyle: Interesting, I haven't seen that project before. +100

src/components/EditorsColumn.jsx
inlinestyle: Oh I meant to comment on the bug issue for that, my resize branch actually fixes that issue (by explicitly demanding resizing on min/maxing).

inlinestyle:
@outoftime Those first two bullets should make that part of the codebase much more legible. I'm not too familiar with `reselect` but it sounds good, curious to see what you'll do with it.

PR #805, opened 2017-04-30
Convert gist export process to sagas & reducers

outoftime:
Continues broad refactor of Redux system to have sagas and reducers do the work of handling complex state changes, rather than procedural thunk actions.

Of all the saga transitions I’ve done, this one (unexpectedly) was one of the most difficult to design. One reason is that the previous approach was particularly hacky—the React `Workspace` component actually initiated the gist export, and then passed the promise of the export’s outcome into the action creator, which then handled resolution and rejection using a thunk. The React code itself also listened for the outcome of the promise, and did its own logic in response.

The underlying complexity comes from these unusual UI specifications:

* When the export completes, the exported gist is loaded into a new tab.
* However, this new tab has to be *opened* when the export is initiated, because windows created outside a calling context initiated by a user action are blocked by modern browsers.
* Creating a new window / changing a window’s location are discrete procedural interactions with the DOM API that are not easily represented as React component state.
* Also, it’s possible that the user has closed the export window before the export completes, in which case we should fall back to displaying a notification containing a link to the exported gist.

So, this is how it ended up:

**`Workspace` component issues an `EXPORT_GIST` action**

* Also opens a window with a spinner and retains it in an instance variable

**`clients` saga consumes `EXPORT_GIST`**

* Selects current project state from store
* Calls `Gists.createFromProject` and waits for it to complete
* If successful, dispatches `GIST_EXPORTED`
* If failed, dispatches `GIST_EXPORT_ERROR`

**`clients` reducer consumes `EXPORT_GIST`**

* Sets `gists.currentExport` to `{status: "waiting"}`

**`clients` reducer consumes `GIST_EXPORTED`**

* Sets `gists.currentExport` to `{status: "ready", url}`

**`clients` reducer consumes `GIST_EXPORT_ERROR`**

* Sets `gists.currentExport` to `{status: "error", error}`

**`Workspace` watches changes to `gists.exportStatus`**

* If change is to `ready`, checks to see if export window is still open
  * If so, loads the gist into it and dispatches `GIST_EXPORT_DISPLAYED`
  * If not, dispatches `GIST_EXPORT_NOT_DISPLAYED`
* If change is to `error`, closes and releases the export window

**`ui` reducer consumes `GIST_EXPORT_NOT_DISPLAYED`**

* Adds notification about exported gist

**`ui` reducer consumes `GIST_EXPORT_ERROR`**

* Adds notification about export error
  * If it’s a `GistExportEmpty`, adds specific message for that


outoftime:
@inlinestyle would be interested in your thoughts on this!

inlinestyle:
@outoftime you're right, there are a surprising number of design constraints here! Taking a look now. Fair warning, I'm familiar but not an expert with `redux-saga` (use `redux-observable` at work). This is a great excuse for me to review the API docs :)

src/components/Workspace.jsx
```diff
@@ -394,7 +393,9 @@ class Workspace extends React.Component {
           allProjects={this.props.allProjects}
           currentProject={this.props.currentProject}
           currentUser={this.props.currentUser}
-          gistExportInProgress={this.props.clients.gists.exportInProgress}
+          gistExportInProgress={
+            get(this.props, 'clients.gists.lastExport') === 'waiting'
```
inlinestyle: Did you mean `'clients.gists.lastExport.status'` here?

src/components/Workspace.jsx
```diff
@@ -331,55 +352,33 @@ class Workspace extends React.Component {
...
-      return;
-    }
+  _handleExportGist() {
+    this._gistExportWindow = openWindowWithWorkaroundForChromeClosingBug(
+      `data:text/html;base64,${spinnerPage}`,
+    );
 
-    if (!this.props.currentUser.authenticated) {
-      // eslint-disable-next-line no-alert
-      if (!confirm(t('workspace.confirmations.anonymous-gist-export'))) {
```
inlinestyle: Was deleting this `confirm` was intentional? If so, should the copy be deleted from `translations.json`?

src/reducers/ui.js
```diff
@@ -24,8 +24,23 @@ function addNotification(state, type, severity, payload = {}) {
   );
 }
 
+export function reduceRoot(rootState, action) {
+  return rootState.update('ui', (state) => {
+    switch (action.type) {
+      case 'GIST_EXPORT_NOT_DISPLAYED':
```
inlinestyle: It looks like if there was an existing error message, a subsequent successful export will replace the error message with a success message if we're not able to open a new tab; however if we are able to open a new tab, the old error message is never cleared out. 

Do you think it's worth adding a `case 'GIST_EXPORT_DISPLAYED':` to clear out `state.ui.notifications`?

inlinestyle:
@outoftime looks good! I left a few questions/checks. I also built the branch and did a QA pass and everything works for me 👍

src/components/Workspace.jsx
outoftime: 😳

src/components/Workspace.jsx
outoftime: Deleting it was intentional, yep—figure it’s not really helping anything, and blocking alerts are generally a high UX cost. Good call on the translation!

src/reducers/ui.js
outoftime: That’s a good thought. Right now notifications are only ever removed by a click on the close button in the notification, but there are situations in which it makes sense to remove a notification about an unsuccessful action that was later done successfully, including the one you bring up here. Another example would be login errors. I don’t think we want to clear _all_ notifications in these scenarios, but we should clear out the relevant one(s).

That said, as written this has the same behavior as the status quo, and I’m inclined to make an overall auto-clear-error-notification change separately. In fact that seems like the kind of good-for-beginner change I’m anxious to accumulate in advance of Hack Day : )

outoftime:
This is great feedback @inlinestyle, thank you!

src/reducers/ui.js
inlinestyle: Sounds good ✏️ 🍎

inlinestyle:
No problem!

outoftime:
btw @inlinestyle definitely interested in your thoughts about `redux-saga` vs `redux-observable`; I don’t have any experience with the latter and have only dabbled in RxJS/reactive programming generally…

inlinestyle:
@outoftime treading carefully to avoid offending anyone, here we go:

`redux-saga` and `redux-observable` are different and excellent solutions to the same problem. Objectively, `redux-saga` has significantly greater community buy-in. I believe this is particularly evident when comparing the available testing ecosystems.

`redux-observable`'s [docs on testing](https://redux-observable.js.org/docs/recipes/WritingTests.html) contain this very memorable passage:
> Testing async code that creates side effects isn't easy. We're still learning the best way to test Epics. If you have found the perfect way, do share!

I really do want `redux-observable` to get more notice since I think ReactiveX is great and they're dominating in other platforms (Android), but I'd call it for `redux-saga` at this moment in time.

src/components/Workspace.jsx
```diff
@@ -128,6 +131,24 @@ class Workspace extends React.Component {
     addEventListener('beforeunload', this._confirmUnload);
   }
 
+  componentDidUpdate(prevProps) {
```
inlinestyle: @outoftime I was thinking about this during my commute this morning and it occurred to me that it might be cleaner to manage the entire lifecycle of the gist export window (currently `_handleExportGist`, `_loadExportedGist`, and `_cleanUpGistExport`) within the `exportGist` saga (or something like "displayGist" saga). 

`Workspace` and the gist export window have no inherent relationship (especially not with the refactoring you've already got here), and distancing them will 
1. make that clearer
2. remove the need for this lifecycle hook (comparing with prev props is a minor code smell IMHO).

outoftime: Hey @inlinestyle this is a good thought, but I feel pretty strongly that any interaction with the DOM should occur inside a React component. It would feel like a big muddling of concerns to me for a Saga to be calling `window.open()` etc.

I definitely agree with the sentiment that the window management has nothing to do with `Workspace` semantically. It would feel much better to have a separate `<GistExport>` component whose sole responsibility was to manage the lifecycle of the export window. I initially shied away from that solution because of the browser constraints around opening a `window` based on a user action, although I’m now thinking it might be possible, since there’s no reason to believe there is any asynchronicity in a Redux store update.

Which brings us to…

> comparing with prev props is a minor code smell IMHO

I *kind of* agree. I would say that, if you are comparing with previous props, it’s worth a good hard look at whether there’s a way to change that fact. BUT, there are situations in which it’s unavoidable, particularly for components that have to bridge the gap between React state and imperative APIs that are external to React (e.g. the `Editor` component; the `GistExport` component I propose above)

Generally though I think a component that is watching prop changes probably shouldn’t _also_ going through normal `render` cycles.

outoftime: [Oh hey](https://github.com/JakeGinnivan/react-popout)

inlinestyle: Switching to declarative (by way of a component wrapper) is way better than what I was proposing 👍 💯

Just for the sake of discussion, I'll say I don't have an issue with something like a saga being devoted to UI provided that's the *only* thing it does; my thinking is that interacting with a window actually has a lot in common with interacting with a server, and that if sagas are appropriate for the latter then they might be appropriate for the former.

**Anyway** I think the encapsulated component approach is best, *but* `react-popout` doesn't look like it's getting much in the way of support these days. [This PR](https://github.com/JakeGinnivan/react-popout/pull/32) calls out the fact that it's pinned to React 14 😞 You could roll your own?

outoftime: Yikes, good catch on the issues with `react-popout`. Well, I think it should be pretty straightforward to hand-roll something that does what we need!

outoftime: @inlinestyle buckle up this is going to get crazy

inlinestyle: jstew

outoftime:
@inlinestyle gonna dump some thoughts in this thread soon—definitely feeling a lot of ambivalence between different approaches we’ve discussed and need your input! Stay tuned

inlinestyle:
image

outoftime:
@inlinestyle OK, so here are the options as I see them:

### Status quo ante code review

`Workspace` handles the window lifecycle directly. You’ve seen it.

### Clients saga manages window lifecycle

I think this would have the advantage of having the overall least code. It keeps things very simple, since we can create a `Window` object at the beginning of the saga, keep a reference to it in a local variable, and do exactly what we need to do when we need to do it. It also looks the most like the current logic in master, just moved from `Workspace` to the saga.

### A different saga manages window lifecycle

I think this would still be pretty easy to implement by just using the right sequence of `take`s and basically structuring the saga itself similarly to the previous proposal. It has the advantage of separating concerns, and it makes it easy to change our mind later. I’m receptive to your earlier observation that interacting with `window` is not _so_ different from interacting with an external API. In particular, creating a new window is different from, say, interacting with the DOM, because the latter requires knowledge of the UI structure, while the former does not.

### Declarative solution

In this case, we’d have a `<PopoutWindow>` component that is given a `url` prop, and maintains a window that it navigates to the given `url` whenever it changes. When the component is unmounted, the window is closed.

I’m having trouble thinking of a satisfying way to set this up, though, because of the interplay of a couple of concerns.

First, each export should get its own window. That is to say, if I export a gist, keep the window open, and then export a second gist, the second gist should open in its own window, without interfering with the first gist.

It follows, then, that our state needs to contain a collection of _all_ exported gists, at least those that may still need an open window.

Second, I don’t believe there is a reliable way to get a DOM event when the window is closed. `react-popout` [uses `onbeforeunload`](https://github.com/JakeGinnivan/react-popout/blob/master/lib/react-popout.jsx#L98), but this will fire any time the window navigates (including because our application has told it to), and anyway we don’t have access to it once we’ve navigated to a cross-origin URL. I think the least bad approach would be to add a handler to [`window.onclose`](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onclose) and fall back to a `setInterval` to poll the `window.closed` property for browsers that don’t support `onclose`.

So, we end up with our state containing a collection of all the exports that we want to be displayed in a pop-out window (e.g. this would not include failed exports). Optionally, we could try to clean up the collection when the user closes the window themselves.

## In summary

I’m leaning toward the single-purpose saga at this point, mostly because it is very simple, and because the declarative solution yields a fairly awkward model of the world, which is kind of missing the point of being declarative. I’d love to hear your thoughts on this, though!

inlinestyle:
@outoftime single-purpose saga sounds best to me too. It will probably be easier to unit test than the component anyway!

I knew the window management api was finicky, but I admit I hadn't really considered how messy managing multiple windows declaratively would get. Given your experience so far, rolling your own doesn't seem worth it anymore.

outoftime:
@inlinestyle OK, I’m feeling good about this—mind giving the saga bit a look-over? Otherwise not much has changed other than fixing that goof you pointed out in previous review.

.eslintrc
```diff
@@ -49,7 +49,8 @@
             {"before": false, "after": true}
         ],
         "complexity": [
-          2
+          2,
+          {"max": 21}
```
inlinestyle: Just wondering, did this come up somewhere in this PR?

outoftime: Yeah, in one of the reducers. My next big goal is to unwind the ball-of-mud `case` statements into the better structure `redux-actions` makes available.

PR #806, opened 2017-04-30
Sticky editor open/close state

### LINKED ISSUE #799: Store visible/hidden tile state, perhaps in URL?
Apologies if this is the wrong place to put enhancement ideas, or if this enhancement has been previously suggested (I searched but couldn't find it.) The idea is to store in the URL which of the Popcode tiles (HTML/CSS/JavaScript/Output) are currently visible. This way, when sharing a link to a Gist, unimportant tiles will remain minimized. CodePen does this using a bitmask (`?editors=1011`); we could steal their implementation?

ajgreenb:
This patch updates Popcode to persist information about which of the editors (incl. "Output") are currently minimized. The information is stored in an array under the key `minimizedComponents` in a project's `popcode.json` file. My implementation moves the `minimizedComponents` state from `ui` to `projects`, since the `popcode.json` file gets *created from* the `project` state. Is that cool? I could modify the implementation so that the `minimizedComponents` state stays under the `ui` subtree, but I think it would require some extra logic during the dehydrate/rehydrate `popcode.json` steps.

Also: on this branch, if you visit Popcode, minimize and maximize an editor, and then close the tab, you get a dialog box warning that you might lose unsaved changes. It's because in `Workspace#_confirmUnload`, `isPristineProject` checks the current project's `updatedAt` time (which is changed when you minimize/maximize an editor. This also happens when you toggle a library, though. Is this behavior desired? And if not, should I address it as part of this PR or should it be separate?

The code for this project is pretty well organized—good job on that! Wasn't too hard for me to jump in and figure out how to implement this feature.

Closes #799

### DIFF
--- a/src/reducers/projects.js
+++ b/src/reducers/projects.js
@@ -141,7 +141,7 @@ export default function reduceProjects(stateIn, action) {
         action.meta.timestamp,
       );
 
-    case 'MINIMIZE_COMPONENT':
+    case 'HIDE_COMPONENT':
       return state.updateIn(
         [action.payload.projectKey, 'hiddenUIComponents'],
         hiddenUIComponents =>
@@ -151,7 +151,7 @@ export default function reduceProjects(stateIn, action) {
         action.meta.timestamp,
       );
 
-    case 'MAXIMIZE_COMPONENT':
+    case 'UNHIDE_COMPONENT':
       return state.updateIn(
         [action.payload.projectKey, 'hiddenUIComponents'],
         hiddenUIComponents =>
--- a/src/reducers/projects.js
+++ b/src/reducers/projects.js
@@ -144,9 +144,8 @@ export default function reduceProjects(stateIn, action) {
     case 'MINIMIZE_COMPONENT':
       return state.updateIn(
         [action.payload.projectKey, 'minimizedComponents'],
-        (minimizedComponents) => {
-          return minimizedComponents.add(action.payload.componentName);
-        },
+        minimizedComponents =>
+          minimizedComponents.add(action.payload.componentName),
       ).setIn(
         [action.payload.projectKey, 'updatedAt'],
         action.meta.timestamp,
@@ -155,9 +154,8 @@ export default function reduceProjects(stateIn, action) {
     case 'MAXIMIZE_COMPONENT':
       return state.updateIn(
         [action.payload.projectKey, 'minimizedComponents'],
-        (minimizedComponents) => {
-          return minimizedComponents.delete(action.payload.componentName);
-        },
+        minimizedComponents =>
+          minimizedComponents.delete(action.payload.componentName),
       ).setIn(
         [action.payload.projectKey, 'updatedAt'],
         action.meta.timestamp,
--- a/src/reducers/projects.js
+++ b/src/reducers/projects.js
@@ -25,13 +25,14 @@ const newProject = Immutable.fromJS({
     javascript: '',
   },
   enabledLibraries: new Immutable.Set(),
+  minimizedComponents: new Immutable.Set(),
 });
 
 function projectToImmutable(project) {
-  return Immutable.fromJS(project).set(
-    'enabledLibraries',
-    new Immutable.Set(project.enabledLibraries),
-  );
+  return Immutable.fromJS(project).merge({
+    enabledLibraries: new Immutable.Set(project.enabledLibraries),
+    minimizedComponents: new Immutable.Set(project.minimizedComponents),
+  });
 }
 
 function addProject(state, project) {
@@ -60,6 +61,7 @@ function importGist(state, projectKey, gistData) {
         join('\n\n'),
       },
       enabledLibraries: popcodeJson.enabledLibraries || [],
+      minimizedComponents: popcodeJson.minimizedComponents || [],
     },
   );
 }
@@ -133,6 +135,28 @@ export default function reduceProjects(stateIn, action) {
             return enabledLibraries.delete(libraryKey);
           }
           return enabledLibraries.add(libraryKey);
+        },
+      ).setIn(
+        [action.payload.projectKey, 'updatedAt'],
+        action.meta.timestamp,
+      );
+
+    case 'MINIMIZE_COMPONENT':
+      return state.updateIn(
+        [action.payload.projectKey, 'minimizedComponents'],
+        (minimizedComponents) => {
+          return minimizedComponents.add(action.payload.componentName);
+        },
+      ).setIn(
+        [action.payload.projectKey, 'updatedAt'],
+        action.meta.timestamp,
+      );
+
+    case 'MAXIMIZE_COMPONENT':
+      return state.updateIn(
+        [action.payload.projectKey, 'minimizedComponents'],
+        (minimizedComponents) => {
+          return minimizedComponents.delete(action.payload.componentName);
         },
       ).setIn(
         [action.payload.projectKey, 'updatedAt'],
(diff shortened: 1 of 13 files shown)


src/reducers/projects.js
```diff
@@ -25,13 +25,14 @@ const newProject = Immutable.fromJS({
     javascript: '',
   },
   enabledLibraries: new Immutable.Set(),
+  minimizedComponents: new Immutable.Set(),
```
outoftime: Does the name `minimizedComponents` still make sense in the context of a project? Honestly I don’t have a better suggestion, just throwing the question out there…

outoftime:
> The code for this project is pretty well organized—good job on that! Wasn't too hard for me to jump in and figure out how to implement this feature.

Thanks and that’s great to hear! If you came across anything where the structure could be improved for clarity/legibility/approachability though, please don’t be shy about letting me know : )

src/reducers/projects.js
ajgreenb: That's a good question. I don't think it fits as well as `enabledLibraries`, but I haven't come up with anything better either. It's probably because `minimizedComponents` really *is* a UI concern that's being crammed into the project state. Renaming `minimizedComponents` as `ui`<sup>1</sup> clarifies that the state is UI-related, but clashes with the actual `ui` state subtree and could lead to confusion.

Since I'm new to the project I'm less qualified to make this call, but I'd say given that `minimizedComponents` isn't a major part of the code base it's probably okay to leave as such, despite the cognitive dissonance.

<sup>1</sup>Idea dump: `projectUI`, `hiddenUI`, `hiddenUIComponents`, `uiState`, `editorState`, `hiddenEditors`, `editors` (this would be a whitelist instead of a blacklist)...none jump out at me as the Right answer.

outoftime: Good brain dump! How about `minimizedUIComponents` or `hiddenUIComponents`?

outoftime: Wait that second one is on your list! Let’s do it!

ajgreenb: Hah, nice! &#8275;synergy&#8275;

Do you think I should change the vocabulary for the actions, too? I.e.,

* `MINIMIZE_COMPONENT` &rarr; `HIDE_COMPONENT`
* `minimizeComponent()` &rarr; `hideComponent()`

outoftime: Yep, I think that’d be great @ajgreenb

ajgreenb:
@outoftime I made all the vocabulary changes. I decided to use "hide"/"unhide" as opposed to "hide"/"show" since I think "unhide" more obviously describes the action. Let me know what you think!

test/unit/reducers/projects.js
```diff
@@ -45,7 +47,7 @@ test('projectCreated', (t) => {
 
 test('updateProjectSource', reducerTest(
   reducer,
-  initProjects({[projectKey]: false}),
+  initProjects({[projectKey]: true}),
```
outoftime: Why this change?

test/unit/reducers/projects.js
```diff
@@ -164,20 +172,44 @@ tap(initProjects({1: false}), projects =>
...
   return reduce(map, (projectsIn, modified, key) => {
-    const projects = reducer(projectsIn, projectCreated(key));
+    let projects = reducer(projectsIn, projectCreated(key));
     if (modified) {
-      return reducer(
-        projects,
-        updateProjectSource(key, 'css', '', now),
-      );
+      projects = reducer(projects, updateProjectSource(key, 'css', '', now));
+      projects = reducer(projects, hideComponent(key, 'output', now));
```
outoftime: I’m not sure that this change is needed either—the idea here is to just modify the project in _some_ way, so that the `updatedAt` flag is set

test/unit/reducers/projects.js
ajgreenb: The change you commented on below broke this particular test; changing `false` to `true` fixes it without affecting the purpose of the test itself.

test/unit/reducers/projects.js
ajgreenb: I wanted an easy way to test `unhideComponent()`, so I needed the stub project's "before" state to have a currently hidden component. I just co-opted the `modified` flag to add a hidden component when the stub project is being created. Is there a better stub to use?

outoftime: I would suggest just using what you have in the previous test, but reversing the order of the input/output. If you keep the initial stub modified (but without changing what that means) I think that should work perfectly.

May seem like a small point but the intent with passing a modified flag to the factory method here is *just* to mark it as having been modified; I’d prefer that the tests not make any other assumptions about what the actual modification is.

ajgreenb: Ah, I love that solution&mdash;great call, thanks! When I wrote the test, I hadn't realized that the `modified` flag was *only* meant to signify that the `updatedAt` flag is set. Completely agree that it's best not to have tests rely on the actual modifications if that's not the intention of the stub. I'll make your change when I get back home.

test/unit/reducers/projects.js
```diff
@@ -164,6 +172,30 @@ tap(initProjects({1: false}), projects =>
...
+    ),
+  )),
+);
+
+tap(initProjects({1: true}), projects =>
+  test('unhideComponent', reducerTest(
+    reducer,
+    projects.update('1', projectIn =>
+      projectIn.set('hiddenUIComponents', new Immutable.Set(['output'])).
+        set('updatedAt', now),
```
ajgreenb: @outoftime I reverted the changes to `initProjects` and updated this test. One thing I noticed is that this second `set` operation is redundant since passing the `modified` flag sets the `updatedAt` property. So maybe it's fine to omit it. And for the new test above this one, I could use `initProjects({1: true})` and do the same. Thoughts?

outoftime: Yep, agreed that omitting the `updatedAt` is probably best since it removes a peripheral concern from the test, although I also don’t feel terribly strong about it.

test/helpers/factory.js
```diff
@@ -18,11 +19,11 @@ export function gistData({html, css, javascript, enabledLibraries = []} = {}) {
       content: javascript,
     });
   }
-  if (!isEmpty(enabledLibraries)) {
```
ajgreenb: I can totally make that change, but it might require something like

```javascript
const popcodeJson = {};
if (!isEmpty(enabledLibraries)) {
  popcodeJSON.enabledLibraries = enabledLibraries;
}
if (!isEmpty(hiddenUIComponents)) {
  popcodeJSON.hiddenUIComponents = hiddenUIComponents;
}
if (!isEmpty(popcodeJSON)) {
  files.push({
    language: 'JSON',
    filename: 'popcode.json',
    content: JSON.stringify(popcodeJSON),
  });
}
```

That way we'd avoid accidentally adding an empty `Array` for `enabledLibraries` or `hiddenUIComponents`. I'm sure there's a nicer way to implement that logic—just sketching it out to express my thoughts.

I modified `gistData` so that it no longer includes default empty `Array` parameters for `enabledLibraries` or `hiddenUIComponents`, so they will be falsey if not passed and `JSON.stringify` won't include them. This would break if anyone ever *purposefully* passed empty `Array`s to `gistData` (because then `popcode.json` would have empty `Array` values), but that's currently not done anywhere `gistData` is used. But it makes sense to me if it should be able to properly handle purposeful empty `Array` arguments. What are your thoughts on that?

ajgreenb: Also please let me know if my comment doesn't make sense, I'm running on fumes at this point and wouldn't be surprised if I'm incomprehensible.

outoftime: You know what—I’m the one who’s not making sense. Somehow I failed to notice this was in the test helper code. What you’ve got here is entirely reasonable!

ajgreenb:
Awesome, thanks @outoftime!! 🎉

Issue #782, opened 2017-04-18
Closing CSS/Javascript doesn't expand view for html box


sarahzinger:
Ran into a minor styling bug: 
![wz2ifqzyoc](https://cloud.githubusercontent.com/assets/6620164/25112741/3032d8a6-23c1-11e7-9fa3-5f9a8e45d954.gif)

outoftime:
@sarahzinger you are absolutely right which is weird because [I totally fixed this issue once already](https://github.com/popcodeorg/popcode/pull/128)! But it does not appear to work now.

Took a quick glance and the [`_resize()` call](https://github.com/popcodeorg/popcode/blob/master/src/components/Editor.jsx#L85) is happening, so maybe it’s being called too soon for the editor to properly infer its size?

ianmclaury:
In case it helps diagnose the problem, I noticed if you click on the unpainted / blank area the missing html will suddenly appear.

outoftime:
thanks @ianmclaury! that is expected since we do explicitly listen for the focus event on the editor and tell it to resize at that time, but that should really be a fallback for resizing when the actual container resizes.

PR #808, opened 2017-05-02
Make editor windows vertically resizable

inlinestyle:
Addresses part of https://github.com/popcodeorg/popcode/issues/302 (horizontal resizing should be comparatively easy to add later)

@outoftime I feel a bit silly turning `EditorsColumn` back into a stateful component here, but I believe it's necessary as long as the implementation relies on refs. Speaking of refs, I looked up a few declarative-style solutions for managing element height, and didn't find anything that solid (nothing with >100 stars && actively maintained). Still, the ref management here is the part I'm least satisfied with. Please let me know if you have any better ideas/libraries for this!

OTOH moving from `state` to `redux` went smoothly and the dragging might feel just a little jittery but I don't think it's too bad.

This also happens to fix https://github.com/popcodeorg/popcode/issues/782.

[Demo!](http://recordit.co/iCnHgBmmOe)

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -11,6 +11,10 @@ export const editorFocusedRequestedLine = createAction(
   'EDITOR_FOCUSED_REQUESTED_LINE',
 );
 
+export const editorsUpdateVerticalFlex = createAction(
+  'EDITORS_UPDATE_VERTICAL_FLEX',
+);
+
 export const notificationTriggered = createAction(
   'NOTIFICATION_TRIGGERED',
   (type, severity = 'error', payload = {}) => ({type, severity, payload}),
--- a/src/components/EditorsColumn.jsx
+++ b/src/components/EditorsColumn.jsx
@@ -1,4 +1,6 @@
 import React from 'react';
+import {DraggableCore} from 'react-draggable';
+import bindAll from 'lodash/bindAll';
 import isEmpty from 'lodash/isEmpty';
 import includes from 'lodash/includes';
 import partial from 'lodash/partial';
@@ -13,58 +15,116 @@ function allErrorsFor(language, errors, runtimeErrors) {
   return errors[language].items;
 }
 
-export default function EditorsColumn({
-  currentProject,
-  errors,
-  onComponentHide,
-  onEditorInput,
-  onRequestedLineFocused,
-  runtimeErrors,
-  ui,
-}) {
-  const editors = [];
-  ['html', 'css', 'javascript'].forEach((language) => {
-    if (includes(currentProject.hiddenUIComponents, `editor.${language}`)) {
-      return;
+function getNodeHeights(refs) {
+  return Array.from(refs).sort().map(([, node]) => {
+    if (node) {
+      const minHeight = window.getComputedStyle(node).minHeight;
+      return {
+        height: node.offsetHeight,
+        minHeight: parseInt(minHeight.replace('px', ''), 10),
+      };
     }
-
-    editors.push(
-      <EditorContainer
-        key={language}
-        language={language}
-        source={currentProject.sources[language]}
-        onHide={
-          partial(onComponentHide, `editor.${language}`)
-        }
-      >
-        <Editor
-          errors={allErrorsFor(language, errors, runtimeErrors)}
-          key={language}
-          language={language}
-          percentageOfHeight={1 / editors.length}
-          projectKey={currentProject.projectKey}
-          requestedFocusedLine={ui.editors.requestedFocusedLine}
-          source={currentProject.sources[language]}
-          onInput={partial(onEditorInput, language)}
-          onRequestedLineFocused={onRequestedLineFocused}
-        />
-      </EditorContainer>,
-    );
+    return {};
   });
+}
 
-  if (isEmpty(editors)) {
-    return null;
+export default class EditorsColumn extends React.Component {
+  constructor(props) {
+    super(props);
+    this.dividerRefs = new Map([[1, null], [2, null], [3, null]]);
+    this.editorRefs = new Map([[1, null], [2, null], [3, null]]);
+    bindAll(
+      this,
+      '_storeDividerRef',
+      '_storeEditorRef',
+      '_handleEditorDividerDrag',
+    );
   }
 
-  return (
-    <div className="environment__column">
-      <div className="environment__columnContents editors">{editors}</div>
-    </div>
-  );
+  _storeEditorRef(index, editor) {
+    this.editorRefs.set(index, editor);
+  }
+
+  _storeDividerRef(index, divider) {
+    this.dividerRefs.set(index, divider);
+  }
+
+  _handleEditorDividerDrag(index, _, {deltaY, lastY, y}) {
+    const {onUpdateFlex} = this.props;
+    const editorHeights = getNodeHeights(this.editorRefs);
+    const dividerHeights = getNodeHeights(this.dividerRefs);
+    onUpdateFlex({deltaY, dividerHeights, editorHeights, index, lastY, y});
+  }
+
+  render() {
+    const {
+      currentProject,
+      editorsFlex,
+      errors,
+      onComponentHide,
+      onEditorInput,
+      onRequestedLineFocused,
+      runtimeErrors,
+      ui,
+    } = this.props;
+
+    const children = [];
+    const languages = ['html', 'css', 'javascript'].filter(language =>
+      !includes(currentProject.hiddenUIComponents, `editor.${language}`),
+    );
+    languages.forEach((language, index) => {
+      children.push(
+        <EditorContainer
+          key={language}
+          language={language}
+          source={currentProject.sources[language]}
+          style={{flex: editorsFlex[index]}}
+          onHide={partial(onComponentHide, `editor.${language}`)}
+          onRef={partial(this._storeEditorRef, index)}
+        >
+          <Editor
+            errors={allErrorsFor(language, errors, runtimeErrors)}
+            key={language}
+            language={language}
+            percentageOfHeight={1 / languages.length}
+            projectKey={currentProject.projectKey}
+            requestedFocusedLine={ui.editors.requestedFocusedLine}
+            source={currentProject.sources[language]}
+            onInput={partial(onEditorInput, language)}
+            onRequestedLineFocused={onRequestedLineFocused}
+          />
+        </EditorContainer>,
+      );
+      if (index < languages.length - 1) {
+        children.push(
+          <DraggableCore
+            key={`divider:${language}`}
+            onDrag={partial(this._handleEditorDividerDrag, index)}
+          >
+            <div
+              className="editors__divider"
+              ref={partial(this._storeDividerRef, index)}
+            />
+          </DraggableCore>,
+        );
+      }
+    });
+
+    if (isEmpty(children)) {
+      return null;
+    }
+
+    return (
+      <div className="environment__column">
+        <div className="environment__columnContents editors">{children}</div>
+      </div>
+    );
+  }
 }
 
 EditorsColumn.propTypes = {
   currentProject: React.PropTypes.object.isRequired,
+  editorsFlex: React.PropTypes.array.isRequired,
   errors: React.PropTypes.object.isRequired,
   runtimeErrors: React.PropTypes.array.isRequired,
   ui: React.PropTypes.shape({
@@ -73,4 +133,5 @@ EditorsColumn.propTypes = {
   onComponentHide: React.PropTypes.func.isRequired,
   onEditorInput: React.PropTypes.func.isRequired,
   onRequestedLineFocused: React.PropTypes.func.isRequired,
+  onUpdateFlex: React.PropTypes.func.isRequired,
 };
--- a/src/components/Workspace.jsx
+++ b/src/components/Workspace.jsx
@@ -41,6 +41,7 @@ import {
   toggleDashboardSubmenu,
   userRequestedFocusedLine,
   editorFocusedRequestedLine,
+  editorsUpdateVerticalFlex,
   notificationTriggered,
   userDismissedNotification,
   exportingGist,
@@ -79,6 +80,7 @@ function mapStateToProps(state) {
     errors: state.get('errors').toJS(),
     runtimeErrors: state.get('runtimeErrors').toJS(),
     isUserTyping: state.getIn(['ui', 'editors', 'typing']),
+    editorsFlex: state.getIn(['ui', 'editors', 'verticalFlex']).toJS(),
     currentUser: state.get('user').toJS(),
     ui: state.get('ui').toJS(),
     clients: state.get('clients').toJS(),
@@ -96,6 +98,7 @@ class Workspace extends React.Component {
       '_handleComponentHide',
       '_handleDashboardSubmenuToggled',
       '_handleEditorInput',
+      '_handleEditorsUpdateVerticalFlex',
       '_handleErrorClick',
       '_handleLibraryToggled',
       '_handleLogOut',
@@ -240,6 +243,10 @@ class Workspace extends React.Component {
 
   _handleToggleDashboard() {
     this.props.dispatch(toggleDashboard());
+  }
+
+  _handleEditorsUpdateVerticalFlex(data) {
+    this.props.dispatch(editorsUpdateVerticalFlex(data));
   }
 
   _listenForAuthChange() {
@@ -394,12 +401,14 @@ class Workspace extends React.Component {
       <div className="environment">
         <EditorsColumn
           currentProject={currentProject}
+          editorsFlex={this.props.editorsFlex}
           errors={this.props.errors}
           runtimeErrors={this.props.runtimeErrors}
           ui={this.props.ui}
           onComponentHide={this._handleComponentHide}
           onEditorInput={this._handleEditorInput}
           onRequestedLineFocused={this._handleRequestedLineFocused}
+          onUpdateFlex={this._handleEditorsUpdateVerticalFlex}
         />
         {this._renderOutput()}
       </div>
@@ -431,6 +440,7 @@ Workspace.propTypes = {
   currentProject: React.PropTypes.object,
   currentUser: React.PropTypes.object.isRequired,
   dispatch: React.PropTypes.func.isRequired,
+  editorsFlex: React.PropTypes.array.isRequired,
   errors: React.PropTypes.object.isRequired,
   isUserTyping: React.PropTypes.bool,
   runtimeErrors: React.PropTypes.array.isRequired,
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -396,21 +396,23 @@ body {
 
 .editors__editor-container {
   box-sizing: border-box;
-  flex: 0 1 100vh;
   border-right: 4px solid var(--color-dark-gray);
-  border-bottom: 4px solid var(--color-dark-gray);
   position: relative;
   z-index: 0;
   display: flex;
-}
-
-.editors__editor-container:last-child {
-  border-bottom: 0;
+  flex-direction: column;
+  min-height: 85px;
 }
 
 .editors__editor {
   z-index: 0;
-  flex: 1 0 100%;
+  flex: 1;
+}
+
+.editors__divider {
+  background-color: var(--color-dark-gray);
+  cursor: ns-resize;
+  min-height: 4px;
 }
 
 .editors__help-text {
--- a/src/reducers/ui.js
+++ b/src/reducers/ui.js
@@ -1,8 +1,14 @@
 import Immutable from 'immutable';
 import pick from 'lodash/pick';
+import {updateVerticalFlex} from '../util/resize';
+
+export const DEFAULT_VERTICAL_FLEX = new Immutable.List(['1', '1', '1']);
 
 const defaultState = new Immutable.Map().
-  set('editors', new Immutable.Map({typing: false})).
+  set('editors', new Immutable.Map({
+    typing: false,
+    verticalFlex: DEFAULT_VERTICAL_FLEX,
+  })).
   set('requestedLine', null).
   set('notifications', new Immutable.Set()).
   set(
@@ -23,7 +29,6 @@ function addNotification(state, type, severity, payload = {}) {
   );
 }
 
-
 function ui(stateIn, action) {
   let state = stateIn;
   if (state === undefined) {
@@ -31,6 +36,15 @@ function ui(stateIn, action) {
   }
 
   switch (action.type) {
+    case 'CHANGE_CURRENT_PROJECT':
+    case 'HIDE_COMPONENT':
+    case 'PROJECT_CREATED':
+    case 'UNHIDE_COMPONENT':
+      return state.setIn(
+        ['editors', 'verticalFlex'],
+        DEFAULT_VERTICAL_FLEX,
+      );
+
     case 'UPDATE_PROJECT_SOURCE':
       return state.setIn(['editors', 'typing'], true);
 
@@ -62,6 +76,12 @@ function ui(stateIn, action) {
 
     case 'EDITOR_FOCUSED_REQUESTED_LINE':
       return state.setIn(['editors', 'requestedFocusedLine'], null);
+
+    case 'EDITORS_UPDATE_VERTICAL_FLEX':
+      return state.updateIn(['editors', 'verticalFlex'], (prevFlex) => {
+        const newFlex = updateVerticalFlex(action.payload);
+        return newFlex ? Immutable.fromJS(newFlex) : prevFlex;
+      });
 
     case 'GIST_NOT_FOUND':
       return addNotification(
--- a/src/util/resize.js
+++ b/src/util/resize.js
@@ -1,0 +1,64 @@
+function isSecondDividerPushingUpFirstDivider(
+  [
+    {height: height0, minHeight: editorMinHeight0},
+    {height: height1, minHeight: editorMinHeight1},
+  ],
+  deltaY,
+) {
+  return height0 > editorMinHeight0 &&
+    height1 === editorMinHeight1 &&
+    deltaY < 0;
+}
+
+function constrainY([,, {height, minHeight}], lastY, deltaY, offset) {
+  const maxDeltaY = height - minHeight;
+  return lastY + Math.min(deltaY, maxDeltaY) - offset;
+}
+
+export function updateVerticalFlex({
+  deltaY,
+  dividerHeights: [{minHeight: dividerMinHeight0}],
+  editorHeights,
+  index,
+  lastY,
+  y,
+}) {
+  const [
+    {height: editorHeight0},
+    {height: editorHeight1},
+    {height: editorHeight2, minHeight: editorMinHeight2},
+  ] = editorHeights;
+  if (index === 0) {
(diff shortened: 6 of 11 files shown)


src/actions/ui.js
```diff
@@ -11,6 +11,13 @@ export const editorFocusedRequestedLine = createAction(
   'EDITOR_FOCUSED_REQUESTED_LINE',
 );
 
+export const editorsUpdatedFlex = createAction(
+  'EDITORS_UPDATED_FLEX',
+  flex => ({flex}),
+);
```
outoftime: Generally my naming convention for actions is:

* If an action is directly triggered by a user interaction, use an imperative (e.g. `UPDATE_EDITOR_FLEX`)
* If an action is triggered by the outcome of an asynchronous operation, use the past tense passive voice (e.g. `GIST_IMPORTED`)

src/components/EditorsColumn.jsx
```diff
@@ -1,70 +1,164 @@
...
 import partial from 'lodash/partial';
 
 import EditorContainer from './EditorContainer';
 import Editor from './Editor';
 
+
+const DIVIDER_HEIGHT = 4;
+const EDITOR_MIN_HEIGHT = 85;
+
+
```
outoftime: Extra whitespace ; )

outoftime:
> OTOH moving from `state` to `redux` went smoothly and the dragging might feel just a little jittery but I don't think it's too bad.

This is mildly astonishing!

src/components/EditorsColumn.jsx
```diff
@@ -1,70 +1,164 @@
...
+  const nodes = editorRefs.filter(Boolean).map(ReactDOM.findDOMNode);
+  const [
+    {offsetHeight: height0},
+    {offsetHeight: height1},
+  ] = nodes;
+  return [
+    height0,
+    height1,
+    (nodes.length === 3) ? nodes[2].offsetHeight : null,
+  ];
```
outoftime: I think it would feel better to just `map` over the nodes? Do we need there to always be three elements?

outoftime:
Couple more thoughts:

* The math here is complicated enough that I think it’s worth encapsulating it in its own module (probably just something in `util/`)
* Also, call me crazy, but should we push the calculation down into the reducer? Seems like the closest thing to the standard pattern would be for the component to just dispatch an action saying “there has been a drag event, here is the relevant information, figure it out” and the reducer does the heavy lifting (or delegates it in the case of the preceding bullet point)

outoftime:
BTW I would say if we do those two things there is less of an argument for the single-purpose resizable component idea

inlinestyle:
@outoftime great feedback! Hope you're feeling at least a little better... Gonna respond to a few of the smaller items then move on to higher-level direction.

**Generalizing to N sections**
I did initially consider going for this! I chose to drop it because it makes the math more complex and subjective; it requires making decisions like "if I move down the 2nd divider in a 10 section column, and sections 7 & 8 are both enlarged while the other sections are at minimum height, how should sections 7 & 8 be collapsed? One at a time? In which order?" (this decision making doesn't exist when we have only three sections, since one is always allowed to flex) 

That, and I figured that getting so abstract would make it tougher to mentally connect this PR with the existing codebase and make the code harder to review than it already is. My thinking is that handling an arbitrary number of sections would be a good followup, but I'll expand on that a bit lower down.

**Abstracting into a `<ResizeableColumn>` component (`<Resizable axis='y'>`, etc)**
I'm for it in principle, but it isn't without cost. The internal component management will get more complex, particularly in regards to managing the dividers - Currently the dividers are *inside* the `EditorContainer`s, which means that their offset values (that we get from `DraggableCore`) are purely relative to their editor; `<ResizableColumn>` would presumably be inserting the dividers *between* `EditorContainer`s which means that for e.g. the divider between sections 2 & 3, all of your calculation needs to take section 1 into account. Further, I'm not sure if there's a non-sucky way to retrieve the heights of the editors from `props.children`. Going to expand on whether I think this is worth doing lower down as well.

**Increasing the draggable surface area of the dividers (without increasing the height of the dividers themselves)**
I'm open to this but I have concerns. Take this example:

image

I'm worried that if someone is actually trying to select the editor row immediately above or below the divider, the drag behavior will get in the way. It's possible there are heuristics we could use to get around that, but that's another level of complexity I'd think we should hold off on (unless you have a really simple solution in mind).

**Higher level direction - two roads diverged in a yellow wood**
- Moving flex calculation into redux (or at least a separate `util/`-style file)
    - This would be low-effort and might make the code cleaner. It does preclude the end-stage of the next option though.
- Moving flex management into something like `<ResizableColumn>`, `<Resizable axis='y'>`, etc
    - Here's one way I can see this going:
        1. Merge initial version `<ResizeableColumn>` that is specific to 0-3 sections
        2. Enhance to handle N sections
        3. Spin off as separate project (as `popcode/resizable-flex` or `inlinestyle/resizable-flex`), since at this point it will be pretty abstracted from Popcode proper
    - I think that paying down the effort of abstracting this into a wrapper component (which is non-trivial, see explanation above) is worth it if we have the end goal of making it available to other projects.

Apologies for the wall of text... to summarize, I think abstracting into a component is worth doing if we want to spin this off into a separate package, otherwise just modularizing the code a bit more (but keeping the component structure the same) would be my vote.

outoftime:
@inlinestyle amazing writeup! And thanks for the well wishes—I am indeed feeling better today.

Section-by-section thoughts:

### Generalizing to N sections

Your reasoning here is solid. I had an intuition that perhaps a general implementation would be *easier* to comprehend than one that’s specific to three sections, but I now understand how the general solution has additional algorithmic complexity, so I think you’re right to at least *start* with a N=3 solution.

### Abstracting into a `<ResizeableColumn>` component

Given all the considerations, I don’t think this is worth the costs at this point. I also think that what a lot of my initial feedback is gesturing at is the pain point of the `EditorsColumn` module being split between two responsibilities—defining the structure of the editors column UI, and dealing with the intricacies of flex resizing. That pain point would be solved by a purpose-built resizing component, but it would also be solved by encapsulating the flex calculations. Overall, I think we should be thinking in terms of how to minimize the surface area between the application’s interface logic and the concern of flex resizing.

### Increasing the draggable surface area of the dividers

The example you bring up is definitely real, but I’d say it’s enough of an edge case that I’m not terribly concerned. I am concerned that right now one has to put a bit of effort into landing the mouse in the right place to resize. Even just adding a couple of pixels north and south of the divider would go a long way I think. That said, dealing with the click target is a great candidate for a followup PR.

(Also: it seems in principle like it should be possible for _clicks_ in the transparent extra-drag-target area to still propagate down to the editor, while _drags_ are captured by the divider, but I couldn’t tell you off the top of my head how I’d actually pull that off)

### Moving flex calculation into redux (or at least a separate `util/`-style file)

Still very bullish on both of these (extract into a utility module, and also move the responsibility for delegating that module into the reducer). I think it’d do very well on the surface area test I propose above. A good (if probably obvious) question to ask is: how much work will it be to reuse this code to deal with the vertical divider? I think reducer-delegating-to-utility-module would do well on that score.

### If we have the end goal of making it available to other projects

I’m definitely into the idea of extracting parts of Popcode into standalone reusable libraries, and this would be a great one in theory. Reading your analysis, though, I’m nervous that actually making something generally reusable but still meaningfully useful would be very very difficult.

src/components/Workspace.jsx
```diff
@@ -140,10 +144,12 @@ class Workspace extends React.Component {
   }
 
   _handleComponentMinimized(componentName) {
+    this.props.dispatch(editorsResetFlex());
```
outoftime: This is a pretty random/specific thing to throw in when we’re having a high-level discussion, BUT since we’re talking about pushing things down into the reducers, I had this on the brain. One of the design principles underlying my in-progress refactor of actions/reducers/sagas is that actions should communicate _what happened_, and the question of _what to do_ should be left to reducers and sagas. So, in this particular case, I don’t think we want an `editorsResetFlex` action—instead, the UI reducer should consume `MINIMIZE_COMPONENT`/`MAXIMIZE_COMPONENT` and reset the flex in response to those actions.

inlinestyle:
@outoftime to the reducer it is! I will attempt to transfer over your comments regarding various redux/testing conventions as well, but apologies in advance if I forget any...

inlinestyle:
@outoftime updated to address all of the comments I could find!
- Moves resize logic to new `util/resize.js`, which is called from the reducer instead of the event handler.
- Replaces `ui.spec.js` tests with `tape` tests.
- Moves flex reset on min/max to reducer for those actions.
- Changes tense to `UPDATE` from `UPDATED`.
- Cleans up the mapping of refs to heights.

Additional change:
- Added the word `VERTICAL` everywhere mentioning flex, except for `EditorsColumn` where I felt that it was implied.

Followup question:
- The constants I put at the top of `util/resize.js` - they're not horribly out of place, but it might be cleaner to put them somewhere like a hypothetical `src/constants/ui`. Thoughts?

src/components/EditorsColumn.jsx
```diff
@@ -13,58 +17,99 @@ function allErrorsFor(language, errors, runtimeErrors) {
...
+          style={{flex: editorsFlex[index], minHeight: EDITOR_MIN_HEIGHT}}
+          onMinimize={
+            partial(onComponentMinimize, `editor.${language}`)
+          }
+        >
+          <Editor
+            errors={allErrorsFor(language, errors, runtimeErrors)}
+            key={language}
+            language={language}
+            percentageOfHeight={1 / languages.length}
```
outoftime: yup

src/components/EditorsColumn.jsx
```diff
@@ -13,58 +17,99 @@ function allErrorsFor(language, errors, runtimeErrors) {
...
+            projectKey={currentProject.projectKey}
+            ref={partial(this._storeEditorRef, index)}
+            requestedFocusedLine={ui.editors.requestedFocusedLine}
+            source={currentProject.sources[language]}
+            onInput={partial(onEditorInput, language)}
+            onRequestedLineFocused={onRequestedLineFocused}
+          />
+          {(index < languages.length - 1) &&
```
outoftime: I’m not a huge fan of using boolean logic operators for flow control—could we just extract this out into a `divider` variable that may or may not be null?

src/components/EditorsColumn.jsx
```diff
@@ -13,58 +17,99 @@ function allErrorsFor(language, errors, runtimeErrors) {
...
+            <DraggableCore
+              axis="y"
+              bounds="parent"
+              key={`divider:${language}`}
+              onDrag={partial(this._handleEditorDividerDrag, index)}
+            >
+              <div
+                className="editors__divider"
+                style={{height: DIVIDER_HEIGHT}}
```
outoftime: Could we instead:

* Give the divider a height in CSS
* Interrogate the DOM to get the divider height, and pass it to the resizer util as one of the inputs to its calculation?

src/components/EditorsColumn.jsx
```diff
@@ -13,58 +17,99 @@ function allErrorsFor(language, errors, runtimeErrors) {
...
+            requestedFocusedLine={ui.editors.requestedFocusedLine}
+            source={currentProject.sources[language]}
+            onInput={partial(onEditorInput, language)}
+            onRequestedLineFocused={onRequestedLineFocused}
+          />
+          {(index < languages.length - 1) &&
+            <DraggableCore
+              axis="y"
+              bounds="parent"
+              key={`divider:${language}`}
```
outoftime: I was recently reading over the React docs on keys, and was reminded that keys need only be unique within the context of their own siblings. So I think `language` alone should be sufficient for a key here.

src/components/EditorsColumn.jsx
outoftime: Seems like we could maybe even use `getComputedStyle` to do the same thing for the `EDITOR_MIN_HEIGHT`? It would be really slick if those values were totally controlled by the stylesheet : )

src/util/resize.js
```diff
@@ -0,0 +1,50 @@
+export const DIVIDER_HEIGHT = 4;
+export const DEFAULT_VERTICAL_FLEX = [
+  `1 1 ${DIVIDER_HEIGHT}px`,
+  `1 1 ${DIVIDER_HEIGHT}px`,
```
outoftime: Confused why we would use the divider height as the `flex-basis` for the editors themselves?

outoftime: I would think the default vertical flex for all the editor containers would still just be `0 1 100vh`—this seems like another thing we could keep in CSS, though? I.e. use the stylesheet flex unless there has been a resize, and only apply a `style` attribute if we need to override the default

outoftime: There may be some subtlety I’m missing here with the relationship of `<Editor>`, `<EditorContainer>`, and the divider though…

inlinestyle: Yup your third comment has it. I think a heavily simplified version should make what's happening clearer:
```jsx
<EditorsColumn>
  <EditorContainer flex='?'>
    <Editor />
    <Divider height='4px'/>
  </EditorContainer>
  <EditorContainer flex='?'>
    <Editor />
    <Divider height='4px'/>
  </EditorContainer>
  <EditorContainer flex='?'>
    <Editor />
  </EditorContainer>
</EditorsColumn>
```
Our `flex` is applied to `EditorContainer`; if we use a default of `1`, (or the equivalent `1 1 0` or the `0 1 100vh` you mention) then each `EditorContainer` gets the same height.

The trouble is that that's not our goal; our goal is to give each *`Editor`* the same height. 

If we give each `EditorContainer` the same height, then the 3rd one will give its `Editor` the full height. However, the 1st `EditorContainer` has to fit both an `Editor` *and* a `Divider` in its allotted height. The `Divider` height is inflexible, so its the height of `Editor` that gets cut by the height of `Divider`. The same goes for the 2nd container.

As a concrete example, if we have an `EditorsColumn` of height `900px`, then each `EditorContainer` will get height `300px`, but the top two `Editor`s will only get height `296px` while the 3rd will be `300px`.

The solution to this lies in altering the `flex-basis` of the three `EditorContainer`s. By giving the 1st and 2nd containers a boosted starting point (boosted by the height of the `Divider`), they'll take up a correspondingly larger allottment of the available height. This way, given an available height of `908px`, the first two `EditorContainer`s will get height `304px` while the 3rd gets height `300px`, and all `Editor`s get height `300px`.

I hope that helps, and that I'm not missing something blindingly obvious...

src/components/EditorsColumn.jsx
inlinestyle: Ah, I will check that out.

src/components/EditorsColumn.jsx
inlinestyle: Sure! I don't mind it personally, but this block *is* getting pretty long.

src/components/EditorsColumn.jsx
inlinestyle: @outoftime that *might* work, and if it does then I'll be happy to use it.

I'm slightly worried that this could turn into one of those times where one needs a property of a rendered thing before it's rendered; in this case, per the discussion about needing the divider's height in order to set the default flex properties of the editor containers, it's possible we need the height of the divider before we've rendered the divider's container.

I can think of a few hacky workarounds for that 
- Say "screw it" to equal default editor heights and allow the top two editors to be slightly smaller (again see discussion about default flex)
- The above hack but then adjusting as soon as we get the divider height

I'll note that I believe this issue only applies to the divider height; I can't think of any trouble using `getComputedStyle` for the editor min height.

Thoughts?

inlinestyle:
@outoftime I decided to bite the bullet and refactor the divider component placement to be *between* editor containers. I'm actually rather pleased with the outcome; having to calculate an offset isn't so bad in comparison to the other height math it ended up simplifying (as an example, the default flex goes back to being the same for all containers).

This also unblocks using `getComputedStyle` for everything, in principle. I did come up against one rather silly issue though - `EditorContainer` is currently a stateless component, so I can't put a `ref` on it for the purposes of getting its `min-height`. If I *do* turn it into a stateful component, then eslint complains because [react/prefer-stateless-function](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-stateless-function.md) isn't nuanced enough to tell that what looks like a stateless component is essentially being used statefully by a parent. 

Two options come to mind for getting around this for the sake of `getComputedStyle`:
- Just `eslint-ignore` the warning
- A child of `EditorContainer`, `Editor`, is already stateful. We could put the `min-height` on `Editor` and retrieve it from there instead; but this isn't great for two reasons-
    1. We still need to put the `min-height` on `EditorContainer`, because that's what's technically being `flex`ed. We could share the value with a `var` in css but...
    2. The "container"-ness of `EditorContainer` implies to me, the reader, that it should be responsible for the height, not the child. Getting it from the `Editor` just feels weird.

outoftime:
> This also unblocks using `getComputedStyle` for everything, in principle. I did come up against one rather silly issue though - `EditorContainer` is currently a stateless component, so I can't put a `ref` on it for the purposes of getting its `min-height`.

Interesting! I wasn’t aware of this restriction but it makes perfect sense.

Here’s an imperfect alternative proposal:

Change `<EditorContainer>` to a class component, but *also* have it capture a `ref` of its container `<div>`. It would then “yield” an object to its parent via a callback; this object would provide an interface for the parent component to directly interrogate the relevant information about the DOM node. The `<EditorsColumn>` would capture that object as an instance variable, just as it would be for something declared directly as a ref.

Slight remix: just delegate the `ref` handler in `<EditorContainer>` to a callback prop, allowing the parent component complete access to the encapsulated DOM element.

One way to summarize this proposal is “a more complicated way to make `<EditorContainer>` a class component”. The main advantage would be avoiding `ReactDOM.findDOMNode` altogether; I think it would mainly be worth doing if we did something analogous with `<Editor>` in service of `extractNodeHeights`.

I’m also very much open to just making `<EditorContainer>` a class component and adding an `eslint-ignore`. Let me know what appeals!

outoftime:
> Slight remix: just delegate the `ref` handler in `<EditorContainer>` to a callback prop, allowing the parent component complete access to the encapsulated DOM element.

Having meditated on it a bit further, I think this has a certain elegance to it and would like to upgrade its status to that of primary imperfect proposal.

src/components/EditorsColumn.jsx
```diff
@@ -13,58 +17,99 @@ function allErrorsFor(language, errors, runtimeErrors) {
...
+            partial(onComponentMinimize, `editor.${language}`)
+          }
+        >
+          <Editor
+            errors={allErrorsFor(language, errors, runtimeErrors)}
+            key={language}
+            language={language}
+            percentageOfHeight={1 / languages.length}
+            projectKey={currentProject.projectKey}
+            ref={partial(this._storeEditorRef, index)}
```
outoftime: (I feel like maybe you addressed this in your initial writeup but now I can’t find it, apologies if I’m retreading old ground)

src/components/EditorsColumn.jsx
```diff
@@ -13,58 +17,99 @@ function allErrorsFor(language, errors, runtimeErrors) {
...
+      editorsFlex,
+      errors,
+      onComponentMinimize,
+      onEditorInput,
+      onRequestedLineFocused,
+      runtimeErrors,
+      ui,
+    } = this.props;
+
+    const editors = [];
```
outoftime: Very nitpicky but maybe we should call this variable something like `children` since it now contains things besides editors : )

src/components/EditorsColumn.jsx
```diff
@@ -13,58 +17,99 @@ function allErrorsFor(language, errors, runtimeErrors) {
...
+            source={currentProject.sources[language]}
+            onInput={partial(onEditorInput, language)}
+            onRequestedLineFocused={onRequestedLineFocused}
+          />
+        </EditorContainer>,
+      );
+      if (index < languages.length - 1) {
+        editors.push(
+          <DraggableCore
+            key={`divider:${language}`}
```
outoftime: My previous note about this got diff-orphaned, so just wanted to make sure it doesn’t get lost

src/components/Workspace.jsx
```diff
@@ -231,6 +234,10 @@ class Workspace extends React.Component {
     this.props.dispatch(toggleDashboard());
   }
 
+  _handleEditorsUpdateVerticalFlex(data) {
+    this.props.dispatch(editorsUpdateVerticalFlex(data));
+  }
```
outoftime: This kind of thing (while 100% appropriate for the way Popcode’s components are currently structured) definitely feels like a strong argument for sprinkling more connected containers throughout the hierarchy.

src/css/application.css
```diff
@@ -396,21 +396,21 @@ body {
...
   border-right: 4px solid var(--color-dark-gray);
-  border-bottom: 4px solid var(--color-dark-gray);
   position: relative;
   z-index: 0;
   display: flex;
-}
-
-.editors__editor-container:last-child {
-  border-bottom: 0;
+  flex-direction: column;
```
outoftime: Wait how did this ever work without this line…

src/reducers/ui.js
```diff
@@ -53,13 +65,13 @@ function ui(stateIn, action) {
       });
 
     case 'COMPONENT_MINIMIZED':
-      return state.update(
+      return state.update(resetEditorsVerticalFlex).update(
```
outoftime: This is a nice way to write this code!

src/util/resize.js
```diff
@@ -0,0 +1,49 @@
...
+export function updateVerticalFlex({deltaY, heights, index, lastY, y}) {
+  const [height0, height1, height2] = heights;
+  if (index === 0) {
+    return [
+      `0 1 ${y}px`,
+      '1',
+      height2 ? `0 1 ${height2}px` : '1',
+    ];
+  }
+  const offset = DIVIDER_HEIGHT + height0;
```
outoftime: For clarity, can we call this something really verbose like `editor1Offset`?

src/util/resize.js
```diff
@@ -0,0 +1,49 @@
...
+  const offsetY = y - offset;
```
outoftime: Similarly—`dragTargetDistanceFromTopOfEditor1` (this is not a serious pitch, but something a little more verbose would help a lot)

src/util/resize.js
```diff
@@ -0,0 +1,49 @@
...
+  ) {
+    return [
+      `0 1 ${height0 + deltaY}px`,
+      `0 1 ${offsetY}px`,
+      '1',
+    ];
+  } else if (height2 > EDITOR_MIN_HEIGHT) {
+    return [
+      `0 1 ${height0}px`,
+      `0 1 ${adjustY(height2, lastY, deltaY) - offset}px`,
```
outoftime: I feel like it might be a little clearer to place the constraint on the final outcome rather than an intermediate step, e.g. (bad pitch):

```js
const maximumHeight1 = height1 + height2 - EDITOR_MIN_HEIGHT;
const constrainedHeight1 = Math.min(height1 + deltaY, maximumHeight1);
```

Then this element becomes `0 1 ${constrainedHeight1}`

src/util/resize.js
```diff
@@ -0,0 +1,49 @@
...
+      '1',
+    ];
+  } else if (deltaY < 0 && offsetY <= height1) {
```
outoftime: I’m probably missing something, but aren’t these tests redundant? In other words, it seems like we’re asking, “is the direction of the drag upwards, AND is the target of the drag above the current bottom of the second editor?” which seems like the same question?

inlinestyle: You're correct about what each condition is testing, but there's an extra bit to consider: what happens when you drag the divider down as far as it will go, and you keep going with the mouse? 

Without some form of correction for this, what happens is the height of the middle container gets inflated, and the top container will suddenly shrink. You can see it for yourself in the original js codepen that you sent me http://codepen.io/chriscoyier/pen/pJpbWK (drag the second divider past the min height of the right section, then drag it back to the middle without letting go).

I can leave comments on these conditions if you want? I know the general project style is to favor code clarity over comments though, which I'm a big fan of.

src/components/EditorsColumn.jsx
inlinestyle: That was good advice! But... when I moved the divider components from inside the containers to between, that put them at the same level as the containers, so they're now sharing a keyspace. I can change what the key looks like of course, but it does have to take the editor container keys into account.

src/components/EditorsColumn.jsx
inlinestyle: That would be fine. My initial reasoning for using an `Array` was that if it were an `Object` I would still probably have the keys be indices, making it equivalent to an `Array` anyway. 

I could use the `language` for the key but then we'd have to move the knowledge of language display order outside of this method.

Did you have something else in mind?

src/css/application.css
inlinestyle: I take that back. It's not technically necessary, but without it the `Editor` is properly sized because it's flexing horizontally, which feels a little weird.

inlinestyle:
@outoftime updated with an attempt to hit all the key notes:
- All `min-heights` are now sourced from CSS.
   - This includes addition of an `onRef` handler, which motivates the elevation of `EditorContainer` to a class component.
- Moved setup of `editorRefs` and `dividerRefs` from `render` to the `constructor`, which I initially felt slightly uneasy about but React seems to be good enough at nulling out dead refs.
- Converted ref storage to use `Immutable.Map`. I'm thinking that a vanilla `Object` might look better though, considering that ref management is an inherently stateful endeavor.
- Made variables in `updateVerticalFlex` more verbose.
- Rebased (I chose to keep editor show/hide handling inside the `ui` reducer even though the action lives in `projects` but that can be changed).

outoftime:
@inlinestyle awesome! I might not have a chance to go through the code again until tomorrow evening unfortunately, but quick responses to your writeup:

> Converted ref storage to use `Immutable.Map`. I'm thinking that a vanilla `Object` might look better though, considering that ref management is an inherently stateful endeavor.

Oh, yep, I wasn’t suggesting an `Immutable.Map`—just an ES6 `Map`. Agreed that for this purpose a mutable data structure is more appropriate.

> I chose to keep editor show/hide handling inside the `ui` reducer even though the action lives in `projects` but that can be changed

This definitely sounds right, but one thing to keep in mind is that editor minimization state is now a property of a project, meaning swiching projects or creating a new one can change which editors are minimized. So, I think we should consume those actions and reset the flex in those cases too.

> React seems to be good enough at nulling out dead refs

This has been my experience as well

inlinestyle:
@outoftime understood! I've updated per the two remarks you did make (good catch with project switching/creation).

src/components/EditorsColumn.jsx
```diff
@@ -13,58 +16,115 @@ function allErrorsFor(language, errors, runtimeErrors) {
...
+        </EditorContainer>,
+      );
+      if (index < languages.length - 1) {
+        children.push(
+          <DraggableCore
+            key={`divider:${language}`}
+            ref={partial(this._storeDividerRef, index)}
+            onDrag={partial(this._handleEditorDividerDrag, index)}
+          >
+            <div className="editors__divider" />
```
outoftime: I think if we attach `_storeDividerRef` to this `<div>` rather than to the `<DraggableCore>`, we should be able to avoid using `ReactDOM.findDOMNode()` altogether?

inlinestyle: Yup, works perfectly.

inlinestyle:
@outoftime updated!

PR #834, opened 2017-05-25
Reducer unhides relevant editor when line focused

### LINKED ISSUE #812: TypeError in /
## Error in Popcode

**TypeError** in **/**
Uncaught TypeError: Cannot read property 'delete' of undefined

[View on Bugsnag](https://app.bugsnag.com/popcode/popcode/errors/5911b7ef263d9b00180e7829?event_id=5911b7ef263d9b00180e7828&i=gh&m=ci)

## Stacktrace

    webpack:///src/reducers/projects.js:118 - 

[View full stacktrace](https://app.bugsnag.com/popcode/popcode/errors/5911b7ef263d9b00180e7829?event_id=5911b7ef263d9b00180e7828&i=gh&m=ci)

### LINKED ISSUE #827: TypeError in /
## Error in Popcode

**TypeError** in **/**
Uncaught TypeError: Cannot read property 'delete' of undefined

[View on Bugsnag](https://app.bugsnag.com/popcode/popcode/errors/59246995a695330018855e4b?event_id=59246995a695330018855e4a&i=gh&m=ci)

## Stacktrace

    webpack:///src/reducers/projects.js:118 - 

[View full stacktrace](https://app.bugsnag.com/popcode/popcode/errors/59246995a695330018855e4b?event_id=59246995a695330018855e4a&i=gh&m=ci)

### LINKED ISSUE #829: TypeError in /
## Error in Popcode

**TypeError** in **/**
Uncaught TypeError: Cannot read property 'delete' of undefined

[View on Bugsnag](https://app.bugsnag.com/popcode/popcode/errors/5925860ddfa8e500187a8f05?event_id=5925860ddfa8e500187a8f04&i=gh&m=ci)

## Stacktrace

    webpack:///src/reducers/projects.js:17 - 

[View full stacktrace](https://app.bugsnag.com/popcode/popcode/errors/5925860ddfa8e500187a8f05?event_id=5925860ddfa8e500187a8f04&i=gh&m=ci)

outoftime:
Previously the Workspace component’s handler for line focus would dispatch an `unhideComponent` action from the handler for a focused line request.  This had trivially buggy behavior in that it omitted the required first argument to `unhideComponent`, which should have been the current project ID.

However, it also doesn’t follow the recently adopted pattern of reducers knowing what to do in response to actions, rather than procedural code dispatching multiple actions to deal with different aspects of responding to a particular user action or external event.

So, remove the call to `unhideComponent` and instead have the `ui` reducer handle the focus line action.

Also while we’re in there, rename the rather awkward `userRequestedFocusedLine` to `focusLine`, following the convention of using the active voice for user-initiated actions.

Fixes #812 
Fixes #827 
Fixes #829

### DIFF
--- a/src/reducers/projects.js
+++ b/src/reducers/projects.js
@@ -45,6 +45,13 @@ function removePristineExcept(state, keepProjectKey) {
   ));
 }
 
+function unhideComponent(state, projectKey, component, timestamp) {
+  return state.updateIn(
+    [projectKey, 'hiddenUIComponents'],
+    hiddenUIComponents => hiddenUIComponents.delete(component),
+  ).setIn([projectKey, 'updatedAt'], timestamp);
+}
+
 function importGist(state, projectKey, gistData) {
   const files = values(gistData.files);
   const popcodeJsonFile = find(files, {filename: 'popcode.json'});
@@ -83,6 +90,13 @@ export function reduceRoot(stateIn, action) {
             projects.get(currentProjectKey),
           );
         }
+      case 'USER_REQUESTED_FOCUSED_LINE':
+        return unhideComponent(
+          projects,
+          stateIn.getIn(['currentProject', 'projectKey']),
+          `editor.${action.payload.language}`,
+          action.meta.timestamp,
+        );
     }
     return projects;
   });
@@ -152,12 +166,10 @@ export default function reduceProjects(stateIn, action) {
       );
 
     case 'UNHIDE_COMPONENT':
-      return state.updateIn(
-        [action.payload.projectKey, 'hiddenUIComponents'],
-        hiddenUIComponents =>
-          hiddenUIComponents.delete(action.payload.componentName),
-      ).setIn(
-        [action.payload.projectKey, 'updatedAt'],
+      return unhideComponent(
+        state,
+        action.payload.projectKey,
+        action.payload.componentName,
         action.meta.timestamp,
       );
 
--- a/src/reducers/projects.js
+++ b/src/reducers/projects.js
@@ -90,7 +90,7 @@ export function reduceRoot(stateIn, action) {
             projects.get(currentProjectKey),
           );
         }
-      case 'USER_REQUESTED_FOCUSED_LINE':
+      case 'FOCUS_LINE':
         return unhideComponent(
           projects,
           stateIn.getIn(['currentProject', 'projectKey']),
--- a/spec/examples/actions/ui.spec.js
+++ b/spec/examples/actions/ui.spec.js
@@ -7,7 +7,7 @@ import createApplicationStore from '../../../src/createApplicationStore';
 
 import {
   editorFocusedRequestedLine,
-  userRequestedFocusedLine,
+  focusLine,
   notificationTriggered,
   userDismissedNotification,
 } from '../../../src/actions';
@@ -27,9 +27,9 @@ describe('interfaceStateActions', () => {
     clock.restore();
   });
 
-  describe('userRequestedFocusedLine', () => {
+  describe('focusLine', () => {
     beforeEach(
-      () => store.dispatch(userRequestedFocusedLine('javascript', 4, 2)),
+      () => store.dispatch(focusLine('javascript', 4, 2)),
     );
 
     it('sets requestedFocusedLine to given value', () => {
@@ -44,7 +44,7 @@ describe('interfaceStateActions', () => {
 
   describe('editorFocusedRequestedLine', () => {
     beforeEach(() => {
-      store.dispatch(userRequestedFocusedLine('javascript', 4, 2));
+      store.dispatch(focusLine('javascript', 4, 2));
       store.dispatch(editorFocusedRequestedLine());
     });
 
--- a/spec/examples/actions/ui.spec.js
+++ b/spec/examples/actions/ui.spec.js
@@ -6,8 +6,6 @@ import {assert} from 'chai';
 import createApplicationStore from '../../../src/createApplicationStore';
 
 import {
-  editorFocusedRequestedLine,
-  focusLine,
   notificationTriggered,
   userDismissedNotification,
 } from '../../../src/actions';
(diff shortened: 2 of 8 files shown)


outoftime:
@ajgreenb thoughts on this?

outoftime:
N.B. I continue to not love the `reduceRoot` pattern, and note with dismay that this PR expands its usage to some extent.

That this isn’t really a new use case—there is still basically a single purpose for `reduceRoot` in the code, namely getting the current project’s key in order to perform a manipulation on it in the `projects` subtree. This is just another case of that.

There are many other actions that amount to “do something to the current project”, but all of those take the current project’s `projectKey` in the action payload. I don’t think there’s any category difference between the latter actions and the couple that use `reduceRoot`, although it would feel particularly goofy to, for instance, pass a `projectKey` to the `logOut` action. I do recall reading someone smart-sounding recommending doing just this sort of thing in a thread somewhere, though.

Anyway, my preference would either be to get rid of `reduceRoot` or to commit fully to its use (possibly getting rid of `combineReducers` entirely in favor of having all subtree reducers operating on the full tree, with the understanding that each will only modify the part of the tree it has responsibility for). I lean toward the former; it would be very awkward, for instance, that the order in which reducers process the state tree would become significant.

@inlinestyle @ajgreenb @alexpelan @jwang1919 definitely interested in your thoughts on this!

jwang1919:
@outoftime I think I've been too long away from Popcode (or my inexperience with JS 😄). I reread your comments several times and still can't make heads or tails onto what exactly this PR is addressing.

outoftime:
@jwang1919 I feel you. I don’t think this is an area of the code you’ve spent a lot of time in—I’d be happy to give you a higher-level overview of the issue at hand IRL tomorrow if we have a few minutes to chat.

src/reducers/projects.js
```diff
@@ -83,6 +83,21 @@ export function reduceRoot(stateIn, action) {
...
           );
         }
+      case 'FOCUS_LINE':
+        {
+          const currentProjectKey =
+            stateIn.getIn(['currentProject', 'projectKey']);
+
+          if (isNil(currentProjectKey)) {
+            return stateIn;
+          }
```
ajgreenb: When would a `FOCUS_LINE` action occur with a nil `currentProjectKey`? My impression is that the only time there's not a current project key is when the app is loading.

src/reducers/projects.js
```diff
@@ -83,6 +83,21 @@ export function reduceRoot(stateIn, action) {
...
+
+          return projects.updateIn(
+            [currentProjectKey, 'hiddenUIComponents'],
+            hiddenUIComponents =>
+              hiddenUIComponents.remove(`editor.${action.payload.language}`),
+          );
+        }
```
ajgreenb: Not sure it matters too much, but would it be worth abstracting this and the logic in the `UNHIDE_COMPONENT` reducer into a single function? Not sure if the DRYness is worth the abstraction overhead in this case, though...

ajgreenb:
So the general idea is that instead of having `_handleErrorClick` dispatch two actions, it just dispatches `FOCUS_LINE`, which gets consumed by two reducers (the `ui` reducer and `reduceRoot`)?

Seems reasonable to me! Especially if you're trying to have components just dispatch single actions that can be handled by multiple reducers.

Is your bone with `reduceRoot` that it allows more-than-required access to the state tree? Does seem like a good pattern to restrict reducers to just the subtrees for which they're responsible.

src/reducers/projects.js
outoftime: Good call—this is overly defensive. I’ll remove it.

src/reducers/projects.js
outoftime: Definitely a good thought—I’m kind of on the fence and am planning a bigger refactor of the reducers very soon, so I’m comfortable with a bit of duplication for now.

outoftime:
@ajgreenb I actually *don’t* have a problem with `reduceRoot` having access to the entire tree—that is the explicit intent, after all. The reasons I’ve grown to dislike the pattern are:

* It’s inconsistently applied (most reducer code is subtree-only)
* If it _were_ consistently applied, the behavior of the system would depend on what order the reducers were run in (reducer B might look at a value in subtree A, which may have been modified by Reducer A in response to the same action, but only if Reducer A ran first)

I think the second issue points at a pretty fundamental problem with the pattern, and is probably the best argument I’ve come across for total isolation.

Another option would be to organize reducers around actions, rather than around subtrees—e.g., one reducer (in its own module) that takes the root tree and does everything it needs to do to handle e.g. `focusLine`. As long as you have only one reducer per action, the order issue goes away. But that seems to be a less common pattern and I like having the subtree reducers be the “authority” on the shape of their subtree (setting up initial state etc.)

outoftime:
@ajgreenb in response to your first question, yes—in general I strive to have actions simply report discrete events (user clicks something, API response returned, etc.) without the dispatching code having an opinion on what should happen. The reducers and sagas should then define how the system reacts to the action. This is a fairly new approach—I’ve just completed overhauling the action creators and introducing sagas to effect it—but I continue to like it much better than the previous pattern of procedurally dispatching multiple actions in response to one event.

src/reducers/projects.js
inlinestyle: Just to check my understanding: if you *were* keeping this line, wouldn't you want to return `projects`, not `stateIn`?

src/reducers/projects.js
inlinestyle: Alternatively - is there a reason this has to be on its own and not be tacked on top of the existing `UNHIDE_COMPONENT` reducer like so?
```
    case 'FOCUS_LINE':
    case 'UNHIDE_COMPONENT':
      return state.updateIn(
        [action.payload.projectKey, 'hiddenUIComponents'],
        hiddenUIComponents =>
          hiddenUIComponents.delete(action.payload.componentName),
      ).setIn(
        [action.payload.projectKey, 'updatedAt'],
        action.meta.timestamp,
      );
```
You'd just have to change `action.payload.language` to `action.payload.componentName` (or vice versa) and consolidate how you're handling `projectKey`. You'd also get the benefit of the `updatedAt` getting properly updated (not sure if that was intentionally left out of the new proposed reducer case or not).

src/reducers/projects.js
outoftime: @inlinestyle yup

src/reducers/projects.js
outoftime: @inlinestyle I actually think changing the action payload structure is a pretty strong argument against this approach—from the perspective of the action, you’ve just asked to focus a particular line of code. It shouldn’t need to know anything about “component names” or even that one of the responses to that action is to maximize the editor component in question.

outoftime: However that is a good point about the `updatedAt`, so I will go ahead and extract a function that does the right thing.

ajgreenb:
Looks like there are two failing tests, but other than that, LGTM!

ajgreenb:
LGTM

outoftime:
Thanks y’all!

PR #836, opened 2017-05-28
Horizontally resizeable workspace columns

inlinestyle:
The second half of https://github.com/popcodeorg/popcode/issues/302!
All of the changes in `resize.js` are one-to-one vocabulary changes:
- `height` -> `size`
- `y` -> `position`
- `editors.verticalFlex` -> `workspace.columnFlex`

I also modified the editor hide/unhide behavior to allow for e.g. horizontal resize proportions to remain constant when an editor component is hidden.

The only complication is in capturing the movement event when the mouse moves over the preview iframe. The most graceful hacky solution I could come up with was enabling same-origin rights in the iframe and manually dispatching a rough clone of the internally captured event, with adjustments for the iframe's position.

If same-origin is a deal breaker I can restructure in terms of `postMessage`.

Recording of resizing, also showing off component hide/unhide behavior: http://recordit.co/IzcC8B7VYk

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -50,7 +50,7 @@
         ],
         "complexity": [
           2,
-          {"max": 22}
+          {"max": 26}
         ],
         "computed-property-spacing": [
             2,
--- a/src/components/EditorsColumn.jsx
+++ b/src/components/EditorsColumn.jsx
@@ -5,6 +5,7 @@ import bindAll from 'lodash/bindAll';
 import isEmpty from 'lodash/isEmpty';
 import includes from 'lodash/includes';
 import partial from 'lodash/partial';
+import {getNodeHeights} from '../util/resize';
 
 import EditorContainer from './EditorContainer';
 import Editor from './Editor';
@@ -16,24 +17,11 @@ function allErrorsFor(language, errors, runtimeErrors) {
   return errors[language].items;
 }
 
-function getNodeHeights(refs) {
-  return Array.from(refs).sort().map(([, node]) => {
-    if (node) {
-      const minHeight = window.getComputedStyle(node).minHeight;
-      return {
-        height: node.offsetHeight,
-        minHeight: parseInt(minHeight.replace('px', ''), 10),
-      };
-    }
-    return {};
-  });
-}
-
 export default class EditorsColumn extends React.Component {
   constructor(props) {
     super(props);
-    this.dividerRefs = new Map([[1, null], [2, null], [3, null]]);
-    this.editorRefs = new Map([[1, null], [2, null], [3, null]]);
+    this.dividerRefs = [null, null];
+    this.editorRefs = [null, null, null];
     bindAll(
       this,
       '_storeDividerRef',
@@ -43,18 +31,22 @@ export default class EditorsColumn extends React.Component {
   }
 
   _storeEditorRef(index, editor) {
-    this.editorRefs.set(index, editor);
+    this.editorRefs[index] = editor;
   }
 
   _storeDividerRef(index, divider) {
-    this.dividerRefs.set(index, divider);
+    this.dividerRefs[index] = divider;
   }
 
   _handleEditorDividerDrag(index, _, {deltaY, lastY, y}) {
-    const {onUpdateFlex} = this.props;
-    const editorHeights = getNodeHeights(this.editorRefs);
-    const dividerHeights = getNodeHeights(this.dividerRefs);
-    onUpdateFlex({deltaY, dividerHeights, editorHeights, index, lastY, y});
+    this.props.onDividerDrag({
+      index,
+      dividerHeights: getNodeHeights(this.dividerRefs),
+      editorHeights: getNodeHeights(this.editorRefs),
+      deltaY,
+      lastY,
+      y,
+    });
   }
 
   render() {
@@ -64,8 +56,10 @@ export default class EditorsColumn extends React.Component {
       errors,
       onComponentHide,
       onEditorInput,
+      onRef,
       onRequestedLineFocused,
       runtimeErrors,
+      style,
       ui,
     } = this.props;
 
@@ -103,7 +97,7 @@ export default class EditorsColumn extends React.Component {
             onDrag={partial(this._handleEditorDividerDrag, index)}
           >
             <div
-              className="editors__divider"
+              className="editors__row-divider"
               ref={partial(this._storeDividerRef, index)}
             />
           </DraggableCore>,
@@ -116,7 +110,7 @@ export default class EditorsColumn extends React.Component {
     }
 
     return (
-      <div className="environment__column">
+      <div className="environment__column" ref={onRef} style={style}>
         <div className="environment__columnContents editors">{children}</div>
       </div>
     );
@@ -128,11 +122,13 @@ EditorsColumn.propTypes = {
   editorsFlex: PropTypes.array.isRequired,
   errors: PropTypes.object.isRequired,
   runtimeErrors: PropTypes.array.isRequired,
+  style: PropTypes.object.isRequired,
   ui: PropTypes.shape({
     editors: PropTypes.object.isRequired,
   }).isRequired,
   onComponentHide: PropTypes.func.isRequired,
+  onDividerDrag: PropTypes.func.isRequired,
   onEditorInput: PropTypes.func.isRequired,
+  onRef: PropTypes.func.isRequired,
   onRequestedLineFocused: PropTypes.func.isRequired,
-  onUpdateFlex: PropTypes.func.isRequired,
 };
--- a/src/components/Workspace.jsx
+++ b/src/components/Workspace.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import {DraggableCore} from 'react-draggable';
 import {connect} from 'react-redux';
 import get from 'lodash/get';
 import values from 'lodash/values';
@@ -13,6 +14,7 @@ import isError from 'lodash/isError';
 import isString from 'lodash/isString';
 import {t} from 'i18next';
 import qs from 'qs';
+import {getNodeWidth, getNodeWidths} from '../util/resize';
 import Bugsnag from '../util/Bugsnag';
 import {
   onSignedIn,
@@ -37,7 +39,10 @@ import {
   toggleDashboardSubmenu,
   focusLine,
   editorFocusedRequestedLine,
-  editorsUpdateVerticalFlex,
+  dragRowDivider,
+  dragColumnDivider,
+  startDragColumnDivider,
+  stopDragColumnDivider,
   notificationTriggered,
   userDismissedNotification,
   exportGist,
@@ -64,8 +69,12 @@ function mapStateToProps(state) {
     currentProject: getCurrentProject(state),
     errors: state.get('errors').toJS(),
     runtimeErrors: state.get('runtimeErrors').toJS(),
+    isDraggingColumnDivider: state.getIn(
+      ['ui', 'workspace', 'isDraggingColumnDivider'],
+    ),
     isUserTyping: state.getIn(['ui', 'editors', 'typing']),
-    editorsFlex: state.getIn(['ui', 'editors', 'verticalFlex']).toJS(),
+    editorsFlex: state.getIn(['ui', 'workspace', 'columnFlex']).toJS(),
+    rowsFlex: state.getIn(['ui', 'workspace', 'rowFlex']).toJS(),
     currentUser: state.get('user').toJS(),
     ui: state.get('ui').toJS(),
     clients: state.get('clients').toJS(),
@@ -82,8 +91,11 @@ class Workspace extends React.Component {
       '_handleComponentUnhide',
       '_handleComponentHide',
       '_handleDashboardSubmenuToggled',
+      '_handleDividerDrag',
+      '_handleDividerStart',
+      '_handleDividerStop',
       '_handleEditorInput',
-      '_handleEditorsUpdateVerticalFlex',
+      '_handleEditorsDividerDrag',
       '_handleErrorClick',
       '_handleLibraryToggled',
       '_handleLogOut',
@@ -95,7 +107,10 @@ class Workspace extends React.Component {
       '_handleRequestedLineFocused',
       '_handleNotificationDismissed',
       '_handleExportGist',
+      '_storeDividerRef',
+      '_storeColumnRef',
     );
+    this.columnRefs = [null, null];
   }
 
   componentWillMount() {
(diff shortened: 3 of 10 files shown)


src/components/Workspace.jsx
```diff
@@ -65,13 +67,27 @@ function mapStateToProps(state) {
...
 
+function getNodeWidth(node) {
+  const minWidth = window.getComputedStyle(node).minWidth;
+  return {
+    width: node.offsetWidth,
+    minWidth: parseInt(minWidth.replace('px', ''), 10),
+  };
+}
+
+function getNodeWidths(refs) {
```
outoftime: Should we extract this and friends into their own module?

src/components/Workspace.jsx
```diff
@@ -65,13 +67,27 @@ function mapStateToProps(state) {
...
+  return Array.from(refs).sort().map(([, node]) =>
```
outoftime: I think the fact that the component stores the `refs` in a map is outside the concern of this function—I’d `map` them before passing them into this function.

src/components/PreviewFrame.jsx
```diff
@@ -9,6 +9,7 @@ import {sourceDelimiter} from '../util/generatePreview';
 const sandboxOptions = [
   'allow-forms',
   'allow-popups',
+  'allow-same-origin',
```
outoftime: So yes, this is a problem. In particular, it presents a security vulnerability—a user could click a malicious gist import link, which contains JavaScript that breaks out of the iframe and attempts to hijack, for instance, the logged in GitHub credentials. The stakes probably aren’t _that_ high given the details of what we actually auth with GitHub etc., but I am a proponent of bright lines where this stuff is concerned.

What about this: when dragging is started, pop a transparent div into existence that overlays the preview frame. That should naturally catch the subsequent `mouseover` events. When dragging ends, of course, we’d make the frame disappear. Thoughts?

outoftime: (Incidentally, the inability to do a sandboxed client-side-only iframe is why Popcode doesn’t support IE/Edge at all.)

src/components/Workspace.jsx
```diff
@@ -95,7 +112,10 @@ class Workspace extends React.Component {
       '_handleRequestedLineFocused',
       '_handleNotificationDismissed',
       '_handleExportGist',
+      '_storeDividerRef',
+      '_storeColumnRef',
     );
+    this.columnRefs = new Map([[1, null], [2, null], [3, null]]);
```
outoftime: It feels weird that we have to initialize a map with three elements, even though it will only ever hold two. How difficult would it be to be more flexible about that, or at least push that concern into the resizer module?

outoftime: Also, I think I had proposed in the previous PR that we store refs in a `Map`—however I think I was thinking about having that map be keyed by the language (of the editors) rather than by index. At this point I definitely agree that it makes more sense to key by index, but with that in mind it’s not clear that a `Map` is better than just using an `Array`…

src/components/Workspace.jsx
```diff
@@ -330,8 +358,36 @@ class Workspace extends React.Component {
...
+  _handleDividerDrag(_, {deltaX, lastX, x}) {
+    const contentSizes = getNodeWidths(this.columnRefs);
+    const dividerSize = getNodeWidth(this.dividerRef);
+    this.props.dispatch(columnsUpdateHorizontalFlex({
+      contentSizes,
+      deltaPosition: deltaX,
+      dividerSizes: [dividerSize],
+      index: 0,
+      lastPosition: lastX,
+      position: x,
```
outoftime: Same general thought as above re ordering of the properties here

src/reducers/ui.js
```diff
@@ -1,15 +1,19 @@
 import Immutable from 'immutable';
 import pick from 'lodash/pick';
-import {updateVerticalFlex} from '../util/resize';
+import {updateFlex} from '../util/resize';
 
-export const DEFAULT_VERTICAL_FLEX = new Immutable.List(['1', '1', '1']);
+const DEFAULT_FLEX = new Immutable.List(['1', '1', '1']);
+export const DEFAULT_WORKSPACE_FLEX = new Immutable.Map({
+  columnFlex: DEFAULT_FLEX,
```
outoftime: Similar to above, seems like default `columnFlex` should ideally be `['1', '1']`

outoftime: (And with that in mind, I’d probably rename `DEFAULT_WORKSPACE_FLEX` to `DEFAULT_FLEX` and declare `DEFAULT_ROW_FLEX` and `DEFAULT_COLUMN_FLEX`)

src/reducers/ui.js
```diff
@@ -77,9 +83,15 @@ export default function ui(stateIn, action) {
     case 'EDITOR_FOCUSED_REQUESTED_LINE':
       return state.setIn(['editors', 'requestedFocusedLine'], null);
 
-    case 'EDITORS_UPDATE_VERTICAL_FLEX':
-      return state.updateIn(['editors', 'verticalFlex'], (prevFlex) => {
-        const newFlex = updateVerticalFlex(action.payload);
+    case 'WORKSPACE_UPDATE_COLUMN_FLEX':
```
outoftime: Riffing on the naming question above, I’d think about naming these actions something that is more focused on user intent rather than what specifically we’re going to do—e.g. maybe `MOVE_COLUMN_DIVIDER`/`MOVE_ROW_DIVIDER`??

src/util/resize.js
```diff
@@ -1,62 +1,71 @@
...
+      `0 1 ${position}px`,
       '1',
-      editorHeight2 ? `0 1 ${editorHeight2}px` : '1',
+      contentSize2 ? `0 1 ${contentSize2}px` : '1',
     ];
   }
-  const distanceToTopOfEditor1 = dividerMinHeight0 + editorHeight0;
-  const projectedHeightOfEditor1 = y - distanceToTopOfEditor1;
+  const distanceToTopOfEditor1 = dividerMinSize0 + contentSize0;
+  const projectedSizeOfEditor1 = position - distanceToTopOfEditor1;
```
outoftime: Couple places here where you’re still using “Editor” rather than “Content”

inlinestyle: 🐑ish

src/components/PreviewFrame.jsx
inlinestyle: @outoftime that makes sense, I didn't consider the case of opening someone else's project. Your proposal to attack the issue from the other side of the iframe has actually given me what I think is an even smoother solution - just add/remove `pointer-events: none` from the iframe's `style` on the divider's `onMouseUp`/`onMouseDown`! I'm going to roll with that.

src/components/EditorsColumn.jsx
```diff
@@ -52,9 +52,16 @@ export default class EditorsColumn extends React.Component {
...
-    onUpdateFlex({deltaY, dividerHeights, editorHeights, index, lastY, y});
+    const contentSizes = getNodeHeights(this.editorRefs);
+    const dividerSizes = getNodeHeights(this.dividerRefs);
+    onUpdateFlex({
+      contentSizes,
+      deltaPosition: deltaY,
+      dividerSizes,
+      index,
+      lastPosition: lastY,
+      position: y,
```
inlinestyle: I can do that; I do alphabetical order by habit just because it's generally uncontroversial :)

inlinestyle:
@outoftime updates! Lots of updates.
- Removed `allow-same-origin` and `mousemove` raising from preview iframe, replaced with toggling of `pointer-events: none;` on `Output` (adds `(START|STOP)_DRAG_COLUMN_DIVIDER` actions). 
- Renamed divider actions to reflect user actions (`DRAG_ROW_DIVIDER`, `DRAG_COLUMN_DIVIDER`).
- Replaced use of `Map` with `Array` for storing refs.
- Updated drag actions to be direction-specific, to be transformed into a direction-agnostic format by helper wrapper methods in resize.js
- Updated drag size data structure to be more flexible with regard to length
- Updated leftover uses of "editor" -> "content" in the resize utility method.
- Some reordering of drag action fields for readability

RE: extracting `getNodeWidths`, `getNodeHeights` - now that we're back to using "height" and "width" outside of the reducer, these methods are kind of specific. I could move them into `resize.js` or somewhere, but I doubt they're doing something that's going to be generalized any time soon.

.eslintrc
```diff
@@ -50,7 +50,7 @@
         ],
         "complexity": [
           2,
-          {"max": 22}
+          {"max": 26}
```
outoftime: 😿

src/components/EditorsColumn.jsx
```diff
@@ -17,23 +17,23 @@ function allErrorsFor(language, errors, runtimeErrors) {
...
+    return null;
   });
 }
 
 export default class EditorsColumn extends React.Component {
   constructor(props) {
     super(props);
-    this.dividerRefs = new Map([[1, null], [2, null], [3, null]]);
-    this.editorRefs = new Map([[1, null], [2, null], [3, null]]);
+    this.dividerRefs = [null, null, null];
```
outoftime: Aren’t there only two dividers?

src/components/EditorsColumn.jsx
```diff
@@ -43,18 +43,22 @@ export default class EditorsColumn extends React.Component {
...
-    this.dividerRefs.set(index, divider);
+    this.dividerRefs[index] = divider;
   }
 
   _handleEditorDividerDrag(index, _, {deltaY, lastY, y}) {
-    const {onUpdateFlex} = this.props;
-    const editorHeights = getNodeHeights(this.editorRefs);
-    const dividerHeights = getNodeHeights(this.dividerRefs);
-    onUpdateFlex({deltaY, dividerHeights, editorHeights, index, lastY, y});
+    this.props.onUpdateFlex({
```
outoftime: How about pushing the `dividerDrag` abstraction down and make the prop name `on-divider-drag`?

src/components/Workspace.jsx
```diff
@@ -64,14 +68,28 @@ function mapStateToProps(state) {
     currentProject: getCurrentProject(state),
     errors: state.get('errors').toJS(),
     runtimeErrors: state.get('runtimeErrors').toJS(),
+    isMaskingOutput: state.getIn(['ui', 'workspace', 'isMaskingOutput']),
```
outoftime: `isMaskingOutput` at first blush made me think that the output is visually occluded—maybe something like `isCapturingOutputPointerEvents`?

Or, alternatively, `isDraggingColumnDivider`? (I.e., let the individual components decide what to do about that fact)

src/components/Workspace.jsx
outoftime: I would still lean toward making these named exports of the `resizer` module, even if we won’t be reusing it, just to reduce the amount of drag-to-resize support code we need in the components. Also has the advantage of fully abstracting the question of “what does the resizer need to know about node dimensions?” from the components.

src/components/Workspace.jsx
```diff
@@ -206,20 +230,32 @@ class Workspace extends React.Component {
...
-        errors={this.props.errors}
+        errors={errors}
         isHidden={includes(hiddenUIComponents, 'output')}
-        project={this.props.currentProject}
-        runtimeErrors={this.props.runtimeErrors}
+        project={currentProject}
+        runtimeErrors={runtimeErrors}
+        style={{
+          flex: rowsFlex[1],
+          pointerEvents: isMaskingOutput ? 'none' : 'all',
```
outoftime: It seems like this decision should be made by the `<Output>` component? Arguably.

src/components/Workspace.jsx
```diff
@@ -330,8 +366,41 @@ class Workspace extends React.Component {
...
+  }
+
+  _handleDividerStop() {
+    this.props.dispatch(stopDragColumnDivider());
+  }
+
+  _handleDividerDrag(_, {deltaX, lastX, x}) {
+    this.props.dispatch(dragColumnDivider({
+      columnWidths: getNodeWidths(this.columnRefs),
+      dividerWidth: getNodeWidth(this.dividerRef),
```
outoftime: I could see an argument for making `dragColumnDivider` still take a one-element array of `dividerWidths` just for consistency / since the fact that there is only one divider isn’t fundamentally related to the fact that we’re working with columns here.

src/reducers/ui.js
```diff
@@ -1,14 +1,23 @@
...
+import {
+  updateEditorColumnFlex,
+  updateWorkspaceRowFlex,
+} from '../util/resize';
 
-export const DEFAULT_VERTICAL_FLEX = new Immutable.List(['1', '1', '1']);
+
+const DEFAULT_FLEX = new Immutable.List(['1', '1', '1']);
+export const DEFAULT_WORKSPACE_FLEX = new Immutable.Map({
+  columnFlex: DEFAULT_FLEX,
```
outoftime: Still seems like the default `columnFlex` should be `['1', '1']`?

src/reducers/ui.js
```diff
@@ -1,14 +1,23 @@
...
+  rowFlex: DEFAULT_FLEX,
+  isMaskingOutput: false,
```
outoftime: A bit odd to have `isMaskingOutput` in a data structure that, going by its name, is concerned with flex dimensions (maybe it just needs a new name)

src/components/Workspace.jsx
inlinestyle: I like `isDraggingColumnDivider`!

src/components/Workspace.jsx
inlinestyle: I have a mild preference for sticking with the non-plural here; I think it supports the goal of `dragColumnDivider`'s signature clearly defining what situation it's meant for.

src/reducers/ui.js
inlinestyle: Hmm yeah. I think you actually suggested this in a comment I somehow dropped; this should be `DEFAULT_WORKSPACE`.

inlinestyle:
@outoftime Friday night updates! The best kind.
- Moved management of `pointer-events` on the preview to `<Output>`
- Renamed `isMaskingOutput` -> `isDraggingColumnDivider`
- Renamed `onUpdateFlex` -> `onDividerDrag`
- Removed the extra slot in `dividerRefs`
- Split out `DEFAULT_FLEX` -> `DEFAULT_COLUMN_FLEX` & `DEFAULT_ROW_FLEX`
- Renamed `DEFAULT_WORKSPACE_FLEX` -> `DEFAULT_WORKSPACE`
- Moved node dimension getters (`getNodeWidths`, etc) to `resize.js`

src/components/Workspace.jsx
outoftime: fair enough!

PR #848, opened 2017-06-02
Add stylelint enforcement of BEM pattern

outoftime:
Adds the `stylelint-selector-bem-pattern` linter and configures it. Uses the BEM preset with global exceptions for selecting `<svg>` and `<path>` elements since we can’t give them class names.

Use the `@define` comment method to mark off components within the single big stylesheet. I think it‘s about time to break each component into its own file, but getting the postcss `@import` of it all to work properly seems non-trivial, so I figured this would be a good first step.

Also:

* Move all the `layout` component selectors into one place
* Make `fontawesome` a utility element rather than its own component

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,4 +1,6 @@
 {
+  "plugins": ["stylelint-selector-bem-pattern"],
+
   "rules": {
     "at-rule-empty-line-before": ["always", {
       "except": [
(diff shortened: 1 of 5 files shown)


inlinestyle:
Yup! 👾 👍

outoftime:
Thanks @inlinestyle !

PR #852, opened 2017-06-04
Debug mode

outoftime:
It can be quite difficult to step through compiled ES2017 code in a debugger, particularly if the code started out with generators or async/await involved. So, introduce “debug mode”, which changes Babel’s behavior such that its output targets latest Chrome, meaning the ES2017 features remain largely (perhaps entirely) intact.

This means we need to configure Babel dynamically, using the `babel-preset-env` plugin in favor of the deprecated `babel-preset-latest`. If the `DEBUG` environment variable is `true`, the `env` preset targets latest Chrome; otherwise, it loads the `config/browsers.json` file and passes the configured browser versions to `env`. This is an aesthetic improvement over using `latest` although I’m fairly sure the outcome is the same (the oldest browser versions we support don’t support anything north of ES5 I don’t think).

Getting rid of `.babelrc` also breaks the `brfs-babel` Browserify transform, which we had been using to be able to do `fs.readFileSync` in our code. Instead, just use `import`s with the `raw` loader to accomplish this.

Finally, I discovered that we were actually generating some invalid JS that gets executed in the preview frame, specifically using a lambda function for an IIFE which is not syntactically valid. This just happens to have worked because the function is compiled to a normal ES5 `function` before it’s stringified. But if we are not compiling to ES5, we get a syntax error. So, I fixed that.


inlinestyle:
This should be really helpful! I'd recommend updating the `README` both to surface the debug option, and also to caution that YMMV on non-Chrome browsers.

PR #855, opened 2017-06-04
Throttle Firebase saves to 2 per second

### LINKED ISSUE #854: Chill out saving the project
Like #781 but regarding saving the project itself.

As I spoke with @outoftime, saving every time the user edits the project probably induces a heavier network load than necessary, something to be wary of given we target underresourced schools of varying Wi-Fi performance. He suggested throttling a save by 10 seconds.

tenyoung795:
This simply debounces for half a second. This could be more sophisticated but
seems like a good start given the many save requests this would cancel.

Fixes #854


tenyoung795:
I added `console.log('Saving')` to `FirebasePersistory.save` to demonstrate the effect to @inlinestyle and he LGTMed the behaivor.

inlinestyle:
LGTM! 👾

ajgreenb:
Does Firebase not do any of its own throttling, or is it actually firing XHRs each time you save? I think Firebase might do its own local saving to allow for offline use.

outoftime:
@ajgreenb hard to say, since Firebase client isn’t open source. i definitely remember reading in the documentation that it will hold on to changes for offline use, but no idea whether it also batches updates in normal network conditions.

It definitely doesn’t do a bunch of XHRs, though. Pretty sure it uses websockets (at least on browsers that support it).

A bit speculative but I don’t see a lot of downside here, you?

tenyoung795:
Firebase is indeed open source ([repo](https://github.com/firebase/firebase-js-sdk)), although it's pretty difficult to follow when actual network activity occurs. That said, up to [here](https://github.com/firebase/firebase-js-sdk/blob/c054dab68ce4d30af7fcb0bb850f1a064ba023c5/src/database/js-client/core/Repo.js#L296) I see no indication of throttling.

Strangely, no network activity from Firebase showed up on Chrome developer tools with or without this commit. I found [this Stack Overflow question](https://stackoverflow.com/q/33997443) which seems to imply that React's polyfill of `XMLHttpRequest` renders it unrecognizable by Chrome developer tools, so I can try [using the native `XMLHttpRequest`](https://stackoverflow.com/a/38063958) to assess how this commit affects Firebase's network activity.

tenyoung795:
Actually, I found the Firebase requests; they're in the "Frames" tab of the Firebase websocket:

image

The green frames (I believe they're requests) appear much slower with this commit.

ajgreenb:
@tenyoung795 nice sleuthing, thanks for the explanation!

outoftime:
whoa! open source firebase is a new development! (i am pretty sure)

PR #858, opened 2017-06-04
Add a grace period during which a closed workspace can be reloaded

ajgreenb:
Currently works if you close a "dirty" tab and then open a new tab.
If you don't close a dirty tab and open a new tab, it will be
pristine. Still need to

* Add a time limit.
* Add test cases.

Is this behavior correct?


ajgreenb:
Just realized...the list of projects when you're logged in refers to the list of Firebase-persisted projects, rather than the list of exported Gists, right?

On this branch, when Popcode creates a new project (rather than loading from a Gist), it will attempt to rehydrate its Workspace state from `localStorage`. This could result in a dupe in the list of projects if a student is authenticated. They might have the FB-persisted project in their list *and* the new one that was created by rehydrating from `localStorage`. Does that make sense?

Perhaps the approach should be,

* If the student is authenticated, get the `localStorage`-serialized project's `projectKey` and use it to look up the project in the list of projects.
* If the student is *not* authenticated, create a new project by rehydrating the whole thing from `localStorage`.

Thoughts?

outoftime:
@ajgreenb you’re right that as written this will lead to duplicates, but I think the solution can be as simple as just saving the project key in local storage when we dump the project, and then holding into it when we pull it back out. Because projects are stored in a `Map` that is keyed by `projectKey`, we’ll naturally avoid any duplication.

ajgreenb:
@outoftime yup, that should work for an authenticated student. But if you're not auth'd, the project isn't saved, right? In that case I imagine we'd want to load the whole project state from localStorage.

In general, what are your thoughts and feelings about this approach? Are there any arguments for saving more of the application state than just the current project?

outoftime:
@ajgreenb sorry, I don’t think I was very clear—I was proposing loading the entire project in all circumstances, but holding on to the original `projectKey` in all cases in order to ensure that we don’t end up with two copies of the same project. It might actually be as simple as just *removing* the line from the current PR that sets a fresh project key? This is the same mechanism by which logging out and back in doesn‘t duplicate the current project.

ajgreenb:
Just a heads up that I'm still planning on continuing this work; I've been slammed this week but hope to find some time next week.

ajgreenb:
Whoof! Been gone a minute, but I'm at it again. Seeing some buggy behavior when saving the editor state while logged in. I added a check to see if the dehydrated project's `projectKey` exists in the list of Firebase-persisted projects, but the check is always returning `false`. This is causing project dupes to be created in the list of projects.

Pretty sure what's happening is that the project is being rehydrated *before* the persisted projects are loaded from Firebase. Does this issue make sense?

Is it possible to defer an action from being dispatched until another occurs? When unauthenticated, it's safe to rehydrated the project immediately, but seems like if you *are* authenticated, you want to wait until you have the list of saved projects to see if you should load one of those rather than creating a new one.

This might be another edge case—you're signed in, so your current project is persisted. Then you sign out. Then you close your tab. Then you open a new tab. Your previous editor state is rehydrated into a new project, since you are logged out. Then you log in. Your current state will be saved as a new project, even though you have a saved version of this project from your last logged-in session.

This really is a tricky feature to implement!

outoftime:
@ajgreenb so happy to see your comment! I was just thinking about this PR : )

So I’m still a little mystified as to the duplicate project issue—since projects are stored in a `Map` keyed by `projectKey`, it seems that as long as the `projectKey` coming from local storage is the same as the `projectKey` coming from Firebase, it shouldn’t be an issue? One will “clobber” the other but in principle they should be identical (except in the edge case where the user manages to make changes to the project before Firebase comes back, but we can probably elide that for now).

I guess it would probably make sense for me to just check out the project and see for myself… maybe i can provide some insight. I think I should probably have time to do that this evening!

To answer your question…

> Is it possible to defer an action from being dispatched until another occurs?

Nope—if this ends up being necessary we would just need to construct a saga that coordinated the various concerns here (loading from Firebase, rehydrating from local storage, etc.). Alternatively we could explicitly dispatch an action indicating that all projects were loaded from Firebase, and then consume that action in another saga that loads the state from local storage, but I think that would end up more complicated than just doing it all in one coordinator saga (this is in fact where sagas really shine in my experience).

> This might be another edge case—you're signed in, so your current project is persisted. Then you sign out. Then you close your tab. Then you open a new tab. Your previous editor state is rehydrated into a new project, since you are logged out. Then you log in. Your current state will be saved as a new project, even though you have a saved version of this project from your last logged-in session.

Tricky! However I think this would still be fine as long as the project keeps the same `projectKey` throughout? (Similarly to how if you sign out and sign back in, all in the same session, you don’t end up with a duplicate project)

src/sagas/projects.js
```diff
@@ -47,6 +50,19 @@ export function* changeCurrentProject() {
...
 }
 
+export function* rehydrateProject({payload: {rehydratedProject}}) {
+  const state = yield select();
+  const allProjectKeys = getProjectKeys(state);
+  if (allProjectKeys.includes(rehydratedProject.projectKey)) {
+    yield put(__changeCurrentProject(rehydratedProject.projectKey));
+  } else {
+    rehydratedProject.projectKey = generateProjectKey();
+    rehydratedProject.updatedAt = null;
```
outoftime: @ajgreenb OK so I just checked this out and poked around with it. A couple of thoughts:

1. This is really cool and a total game changer.
2. If you simply remove the above two lines, as well as the entire `if` branch, you should be good. By simply keeping the `projectKey` stable throughout the process, key collisions ensure that we never get a duplicate when we don’t want one.

I have some more nuts-and-bolts feedback on the code as well but I’ll hold off until you’ve got it where you want it. Let me know!

ajgreenb:
I can no longer remember why I decided to regenerate the `projectKey` and reset `updatedAt` upon rehydration&mdash;I think I was paranoid about accidentally clobbering a Firebase-persisted project, but that seems like a vanishingly rare edge case after thinking about it some more. Happy to hear any kind of feedback you may have!

src/sagas/projects.js
```diff
@@ -47,6 +50,17 @@ export function* changeCurrentProject() {
   yield call(saveCurrentProject, state);
 }
 
+export function* rehydrateProject({payload: {rehydratedProject}}) {
+  const state = yield select();
+  const allProjectKeys = getProjectKeys(state);
+  if (allProjectKeys.includes(rehydratedProject.projectKey)) {
+    yield put(__changeCurrentProject(rehydratedProject.projectKey));
```
outoftime: This isn’t hurting anything but I don’t think it’s really necessary—if anything the copy of the project store in local storage is likely to be _more_ up-to-date than the one stored in Firebase, so we might as well let the local storage version clobber it?

src/sagas/projects.js
```diff
@@ -47,6 +50,17 @@ export function* changeCurrentProject() {
   yield call(saveCurrentProject, state);
 }
 
+export function* rehydrateProject({payload: {rehydratedProject}}) {
```
outoftime: Are you down to add some tests for this as well as a test for `applicationLoaded` that asserts that it’s called if there is a rehydrated project?

src/sagas/projects.js
```diff
@@ -47,6 +50,17 @@ export function* changeCurrentProject() {
...
+  const state = yield select();
+  const allProjectKeys = getProjectKeys(state);
+  if (allProjectKeys.includes(rehydratedProject.projectKey)) {
+    yield put(__changeCurrentProject(rehydratedProject.projectKey));
+  } else {
+    yield put(projectLoaded(rehydratedProject));
+    yield put(__changeCurrentProject(rehydratedProject.projectKey));
```
outoftime: Generally I consider it a red flag any time more than one action is called from the same place in the code. I try to follow a pattern where actions simply encode _what happened_ (generally either the user took some action, or an asynchronous operation has completed); then reducers and sagas decide what to do about it. In this case I would suggest a `REHYDRATED_PROJECT_IMPORTED` action, equivalent to `SNAPSHOT_IMPORTED` and `GIST_IMPORTED`, and then update the appropriate reducers (and their tests!)

outoftime: Generally I consider it a red flag any time more than one action is called from the same place in the code. I try to follow a pattern where actions simply encode _what happened_ (typically either the user took some action, or an asynchronous operation has completed); then reducers and sagas decide what to do about it. In this case I would suggest an e.g. `PROJECT_RESTORED_FROM_LAST_SESSION` action, equivalent to `SNAPSHOT_IMPORTED` and `GIST_IMPORTED`, and then update the appropriate reducers (and their tests!)

src/sagas/projects.js
```diff
@@ -16,13 +16,14 @@ import {
   gistNotFound,
   projectCreated,
   projectLoaded,
+  changeCurrentProject as __changeCurrentProject,
```
outoftime: I don’t think we’ll need this import if we just create a new `projectRestoredFromLastSession` action, but FWIW I think it’s better to disambiguate more explicitly, e.g. `changeCurrentProject as createChangeCurrentProjectAction`

src/sagas/projects.js
ajgreenb: For sure!

src/sagas/projects.js
ajgreenb: Here's my understanding of your suggestion:

`rehydrateProject()` dispatches `REHYDRATED_PROJECT_IMPORTED` by calling `projectRestoredFromLastSession()`. If you're *not* authenticated, the reducer creates a new project from the rehydrated version. If you *are* authenticated, the reducer loads the saved project from Firebase and overwrites it with the rehydrated version.

Is that what you had in mind? Sounds reasonable to me! Let me know if I've misunderstood.

src/components/Workspace.jsx
```diff
@@ -96,30 +96,41 @@ class Workspace extends React.Component {
...
   componentWillUnmount() {
-    removeEventListener('beforeunload', this._confirmUnload);
+    removeEventListener('beforeunload', this._handleUnload);
   }
 
-  _confirmUnload(event) {
-    const {currentUser, currentProject} = this.props;
-    if (!currentUser.authenticated) {
-      if (!isNull(currentProject) && !isPristineProject(currentProject)) {
-        event.returnValue = t('workspace.confirmations.unload-unsaved');
```
ajgreenb: ✅

ajgreenb:
Just pushed a commit that abstracts the interaction with `localStorage` into a lightweight client&mdash;let me know if you have any suggestions for improvement.

src/sagas/projects.js
outoftime: @ajgreenb nope, what I’m proposing is a bit different / simpler I think:

The `rehydrateProject()` saga dispatches `PROJECT_RESTORED_FROM_LAST_SESSION` by `put`ting `projectRestoredFromLastSession(rehydratedProject)`. Regardless of login state, the reducers and sagas handle this action appropriately, to wit:

* The `currentProject` reducer updates the `currentProject.projectKey` to be that of the rehydrated project
* The `projects` reducer adds the project to the collection of projects
* The `errors` saga initiates validation

Effectively the system would react to `PROJECT_RESTORED_FROM_LAST_SESSION` in the same way it reacts to `SNAPSHOT_IMPORTED`. However, there’s nothing fundamental that _requires_ that the system react in the same way to those two actions, so it’s best to keep them as separate actions to allow more differences to emerge in the future.

src/clients/projectStorage.js
```diff
@@ -0,0 +1,20 @@
...
+    project,
+  }));
+}
+
+export function rehydrateProject() {
+  const dehydrated = localStorage.getItem(PROJECT_STORAGE_KEY);
+  localStorage.removeItem(PROJECT_STORAGE_KEY);
+  if (dehydrated) {
+    const rehydrated = JSON.parse(dehydrated);
+    if (Date.now() - rehydrated.dehydratedAt <= 5 * 60 * 1000) {
```
outoftime: I’d also pull the grace period into a constant

src/clients/projectStorage.js
```diff
@@ -0,0 +1,20 @@
+const PROJECT_STORAGE_KEY = 'last-closed-session-project-state';
```
outoftime: This looks good structurally, but I’d prefer to follow the pattern of naming the client module after the service it wraps rather than the purpose it’s currently serving (e.g. other clients are `firebase`, `gitHub`, etc.). So probably just `src/clients/localStorage.js`

ajgreenb:
That was an awesome simplification! Thanks! I like that the logic can be identical regardless of authentication state. I *believe* I've addressed all your suggestions so far, with the exception of adding tests. I'll start working on those now.

src/sagas/projects.js
ajgreenb: Now that this function only contains one line, do you think its body should be inlined into `applicationLoaded`?

outoftime: @ajgreenb we have the facts and we’re voting yes!

outoftime:
@ajgreenb awesome!! just looked it over and agreed, this looks great and ready to roll modulo adding tests!

ajgreenb:
Just added some tests—let me know if there are any more you'd like to see.

outoftime:
@ajgreenb want to rebase off master and i’ll release!!??

ajgreenb:
Let's do it!!! Thanks for all your help on this @outoftime!

outoftime:
@ajgreenb thank YOU! this is a kickass PR!

PR #859, opened 2017-06-04
Insert popcode JS and CSS libraries on the top of the head tag

### LINKED ISSUE #815: External starter code scripts can’t use libraries
If you, for instance, write an external starter code script and host it on RawGit, the script can’t access jQuery, because the jQuery tag is put at the very bottom of the parsed DOM body.

Instead, we should put library scripts at the top of the `<head>`, so that other scripts can use them.

jwang1919:
Also loaded css/js as an internal source instead of a data64 src text. Discovered that loading via data64 does not happen synchronously as expected.

Fixes #815

### DIFF
--- a/src/util/generatePreview.js
+++ b/src/util/generatePreview.js
@@ -181,7 +181,8 @@ class PreviewGenerator {
 
   _attachJavascriptLibrary(javascript) {
     const scriptTag = this.previewDocument.createElement('script');
-    scriptTag.innerHTML = javascript;
+    const javascriptText = String(javascript);
+    scriptTag.innerHTML = javascriptText.replace(/<\/script>/g, '<\\/script>');
     this._previewHead.insertBefore(scriptTag, this._previewHead.firstChild);
   }
 }
--- a/src/util/generatePreview.js
+++ b/src/util/generatePreview.js
@@ -1,11 +1,9 @@
 import castArray from 'lodash/castArray';
 import pick from 'lodash/pick';
-import base64 from 'base64-js';
 import loopBreaker from 'loop-breaker';
 import libraries from '../config/libraries';
 import previewFrameLibraries from '../config/previewFrameLibraries';
 
-const textEncoder = new TextEncoder('utf-8');
 const parser = new DOMParser();
 
 const sourceDelimiter = '/*__POPCODESTART__*/';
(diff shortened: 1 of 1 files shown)


src/util/generatePreview.js
```diff
@@ -176,18 +175,14 @@ class PreviewGenerator {
...
+    this._previewHead.insertBefore(linkTag, this._previewHead.firstChild);
   }
 
   _attachJavascriptLibrary(javascript) {
     const scriptTag = this.previewDocument.createElement('script');
-    const base64encoded = base64.fromByteArray(javascript);
-    scriptTag.src =
-      `data:text/javascript;charset=utf-8;base64,${base64encoded}`;
-    this.previewBody.appendChild(scriptTag);
+    scriptTag.innerHTML = javascript;
```
outoftime: wowowow if you do the same thing with a `<div>` it escapes the text!!

src/util/generatePreview.js
```diff
@@ -176,18 +175,14 @@ class PreviewGenerator {
   _attachCssLibrary(css) {
     const linkTag = this.previewDocument.createElement('link');
     linkTag.rel = 'stylesheet';
-
-    const base64encoded = base64.fromByteArray(css);
-    linkTag.href = `data:text/css;charset=utf-8;base64,${base64encoded}`;
-    this._previewHead.appendChild(linkTag);
+    linkTag.innerHTML = css;
+    this._previewHead.insertBefore(linkTag, this._previewHead.firstChild);
```
outoftime: Is there a specific motivation for this change? Just for consistency?

jwang1919: Yep

jwang1919:
I escaped the closing `</script>` tag which seems to have done it.
To test this functionality, I re-enabled the ember library and it works fine.

Comments are welcome.

jwang1919:
Rebased to latest master commit

src/util/generatePreview.js
```diff
@@ -177,18 +175,15 @@ class PreviewGenerator {
...
   }
 
   _attachJavascriptLibrary(javascript) {
     const scriptTag = this.previewDocument.createElement('script');
-    const base64encoded = base64.fromByteArray(textEncoder.encode(javascript));
-    scriptTag.src =
-      `data:text/javascript;charset=utf-8;base64,${base64encoded}`;
-    this.previewBody.appendChild(scriptTag);
+    const javascriptText = String(javascript);
+    scriptTag.innerHTML = javascriptText.replace(/<\/script>/g, '<\\/script>');
```
outoftime: Will this kind of escaping maintain existing functionality, though? That’s a JS escape rather than an HTML escape so I fear not.

E.g., as a reduction, would this “library” continue to function?

```js
console.log('</script>');
```

jwang1919: Surprisingly yes!

I tested with this example:
`const javascriptText = String("console.log('</script>')");`

outoftime: Wow that’s kind of insane… but… okay!!!

PR #860, opened 2017-06-04
WIP OfflinePlugin

quixotically:
No errors in online mode, but does not seem to work offline on localhost

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,6 +98,7 @@
     "lodash": "^4.15.0",
     "loop-breaker": "^0.1.0-alpha.7",
     "moment": "^2.14.1",
+    "offline-plugin": "^4.8.1",
     "pify": "^2.3.0",
     "promise-retry": "^1.1.1",
     "prop-types": "^15.5.10",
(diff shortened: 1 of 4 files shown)


outoftime:
@quixotically YO GUESS WHAT

I just checked this out and ran it locally through an nginx server that does some kind of DNS and…

it totally works!!

I am still curious as to why it doesn’t work on `localhost:3000` (it doesn’t for me either) but this is a very exciting development…!

Seems like at least for the purposes of dev you could probably just throw something in `/etc/hosts`?

outoftime:
@quixotically so one thing I noticed playing around with this more is that the service worker is `scope`d to `/compiled`, because that’s where `sw.js` ends up. This is an artifact of the fact that `static/` contains some files that are checked in to version control (e.g. `static/index.html`), while compiled files go into `static/compiled`. This isn’t a great structure and I’m considering just porting the build chain over to [Blendid](https://github.com/vigetlabs/blendid) which is a good framework and also doesn’t have this problem.

PR #861, opened 2017-06-04
Rich text error messages

### LINKED ISSUE #374: Rich text formatting for error messages
Make it possible to format error messages (probably using Markdown). Take advantage of this to, e.g., put code in monospace fonts, etc.

aduros:
This is a WIP for rich text error messages.

It was a little trickier than I thought because the plain text error is used inside Ace when you hover over the red X. We use a markdown remover to remove markdown syntax from those errors. There are still certain messages that double-escape the text, especially when there are special characters inside of backticks, which ends up with ugly  #`"&lt;"` strings in the plain text error. "mismatched-close-tag" is one example of the messages affected.

My CSS changes failed linting, feel free to tweak that.

Fixes #374.


outoftime:
> My CSS changes failed linting, feel free to tweak that.

@aduros doesn’t look like I have write access to your branch, but you should be able to fix the tests by just adding:

```css
/* postcss-bem-linter: ignore */
```

right above the descendant-selectors you added to the stylesheet (descendant selectors are indeed fine in this particular case, since we’re dealing with automatically generated markup).

> There are still certain messages that double-escape the text, especially when there are special characters inside of backticks, which ends up with ugly  #`"&lt;"` strings in the plain text error. 

Do you have a good sense of why this is? I can dig into it if you’re stuck…

> It was a little trickier than I thought because the plain text error is used inside Ace when you hover over the red X. We use a markdown remover to remove markdown syntax from those errors.

Ugh, this is like the ur-annoying problem with translations! Your solution here seems as good as anything I can think of.

aduros:
On 06/06/2017 08:21 PM, Mat Brown wrote:
> Do you have a good sense of why this is? I can dig into it if you’re 
> stuck…

I think it has to do with double escaping between both markdown and the 
i18n library. I'm pretty stuck, feel free to take it from here!

By the way, I want to call out the potential security implications of 
this change. The way that the markdown is rendered through React could 
be open to XSS.

outoftime:
> I'm pretty stuck, feel free to take it from here!

OK, sure, I’ll take a look.

> By the way, I want to call out the potential security implications of this change. The way that the markdown is rendered through React could be open to XSS.

@aduros thanks for flagging this—can you elaborate a bit? I’m no fan of security holes (that’s why Popcode doesn’t support IE at all). However in this case I can’t see a vector for attack—since we are rendering Markdown out of our translation files, and those files are taking input from discrete tokens  output by various parsers/linters, I’m not sure how one could get a wayward `<script>` tag (etc.) into Popcode’s DOM.

aduros:
On 06/11/2017 12:54 PM, Mat Brown wrote:
> @aduros <https://github.com/aduros> thanks for flagging this—can you 
> elaborate a bit? I’m no fan of security holes (that’s why Popcode 
> doesn’t support IE at all). However in this case I can’t see a vector 
> for attack—since we are rendering Markdown out of our translation 
> files, and those files are taking input from discrete tokens output by 
> various parsers/linters, I’m not sure how one could get a wayward 
> |<script>| tag (etc.) into Popcode’s DOM.

I should have mentioned in that comment, but for some translations I 
used an unescaping syntax in an attempt to get rid of the double 
escaping: https://www.i18next.com/interpolation.html#unescape. I haven't 
tested this, but I think you might be able to craft an input token that 
contains a backtick character followed by raw HTML, and that'll be put 
into the DOM unescaped?

Come to think of it, even without the unescaping syntax, you could still 
have an input token that is interpreted as markdown when it shouldn't 
be. I guess the worst case there is just that people could mess with the 
formatting of messages, which probably isn't as big a deal.

Issue #785, opened 2017-04-19
Reset button


outoftime:
Often when building interactive pages, you may want to test an interaction, and then reset the page back to its initial state to test it again. As a workaround, coders can pop the page out and refresh it; or make a trivial modification to one of the sources to trigger a refresh.

Better to have first-class support for this operation. From a UI standpoint, probably a little reload icon sitting next to the pop-out icon in the top right corner of the output would be just the thing.

outoftime:
Closed by #862

PR #862, opened 2017-06-04
reset button

kaylee42:
NOT FINISHED WORK IN PROGRESS ON #785 reset button

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -37,3 +37,8 @@ export const userDismissedNotification = createAction(
   'USER_DISMISSED_NOTIFICATION',
   type => ({type}),
 );
+
+export const setRefreshTimestamp = createAction(
+  'SET_REFRESH_TIMESTAMP',
+  (timestamp = Date.now()) => ({timestamp})
+);
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -40,5 +40,5 @@ export const userDismissedNotification = createAction(
 
 export const refreshPreview = createAction(
   'REFRESH_PREVIEW',
-  (timestamp = Date.now()) => ({timestamp})
+  timestamp => ({timestamp}),
 );
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -38,7 +38,7 @@ export const userDismissedNotification = createAction(
   type => ({type}),
 );
 
-export const setRefreshTimestamp = createAction(
-  'SET_REFRESH_TIMESTAMP',
+export const refreshPreview = createAction(
+  'REFRESH_PREVIEW',
   (timestamp = Date.now()) => ({timestamp})
 );
--- a/src/components/Preview.jsx
+++ b/src/components/Preview.jsx
@@ -7,22 +7,17 @@ import classnames from 'classnames';
 import generatePreview from '../util/generatePreview';
 import {openWindowWithWorkaroundForChromeClosingBug} from '../util';
 import PreviewFrame from './PreviewFrame';
-import {
-  refreshPreview
-} from '../actions';
 
 class Preview extends React.Component {
   constructor() {
     super();
     bindAll(this, '_handlePopOutClick');
-    bindAll(this, '_handleRefreshClick');
   }
 
-  _generateDocument(isLivePreview = false, timestamp = this.props.lastRefreshTimestamp) {
+  _generateDocument(isLivePreview = false) {
     if (!this.props.isValid) {
       return '';
     }
-
     const project = this.props.project;
 
     if (project === undefined) {
@@ -36,7 +31,7 @@ class Preview extends React.Component {
         propagateErrorsToParent: isLivePreview,
         breakLoops: isLivePreview,
         nonBlockingAlertsAndPrompts: isLivePreview,
-        refreshTimestamp: timestamp,
+        lastRefreshTimestamp: this.props.lastRefreshTimestamp,
       },
     );
   }
@@ -54,15 +49,6 @@ class Preview extends React.Component {
     openWindowWithWorkaroundForChromeClosingBug(url);
   }
 
-  _handleRefreshClick() {
-    this.props.dispatch(refreshPreview(Date.now()));
-  }
-
-  // _refresh() {
-  //   const currentTimestamp = Date.now();
-  //   this._generateDocument(true, currentTimestamp);
-  // }
-
   render() {
     return (
       <div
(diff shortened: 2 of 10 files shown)


outoftime:
@ibrand @kaylee42 i know this is still a work in progress but would you be open to a bit of feedback on what you’ve got so far? high level this looks great, just a couple of minor course-corrections that might be easier at this stage rather than when you’ve got it more cooked. i’m happy to hold off too, though, just let me know!

ibrand:
@outoftime That sounds great! Definitely wanna make sure we're headed in the right direction. We also know that some of that code is probably a little off track, so feedback is welcome.

src/actions/ui.js
```diff
@@ -37,3 +37,8 @@ export const userDismissedNotification = createAction(
   'USER_DISMISSED_NOTIFICATION',
   type => ({type}),
 );
+
+export const setRefreshTimestamp = createAction(
+  'SET_REFRESH_TIMESTAMP',
```
outoftime: Very low-level comment but as a naming convention I generally like to try to describe user-initiated actions from the standpoint of user intent, so this one something like `REFRESH_PREVIEW`.

src/components/Preview.jsx
```diff
@@ -49,6 +54,15 @@ class Preview extends React.Component {
     openWindowWithWorkaroundForChromeClosingBug(url);
   }
 
+  _handleRefreshClick() {
+    this.props.dispatch(setRefreshTimestamp(Date.now()));
```
outoftime: So dispatching the `setRefreshTimestamp` action is definitely the right move, but the way things are currently arranged, the `Workspace` component is the only one that is ”connected“ to the Redux store. So, you’ll want to pass a callback down the component tree from `Workspace` and then have this component just call the callback it’s given.

(Full disclosure: I’m pretty strongly considering changing the convention above and using many more connected components, probably in line with the [Presentational and Container Component pattern](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0), but I think it’s a good idea to stick with the existing convention until it’s officially changed).

src/css/application.css
```diff
@@ -517,6 +517,22 @@ body {
   opacity: 0.5;
 }
 
+.preview__reset-button {
+  background: url(../images/popout.png);
```
outoftime: Looks like FontAwesome [does have a refresh icon](http://fontawesome.io/icon/refresh/) and we could also switch the existing [pop-out icon to FontAwesome](http://fontawesome.io/icon/external-link/) while we’re in there

src/components/Preview.jsx
```diff
@@ -7,14 +7,18 @@ import classnames from 'classnames';
...
 
 class Preview extends React.Component {
   constructor() {
     super();
     bindAll(this, '_handlePopOutClick');
+    bindAll(this, '_handleRefreshClick');
   }
 
-  _generateDocument(isLivePreview = false) {
+  _generateDocument(isLivePreview = false, timestamp = this.props.lastRefreshTimestamp) {
```
outoftime: I don’t think there’s any need for `timestamp` to be a parameter here—seems like it will always come from the `props`?

src/components/Preview.jsx
```diff
@@ -32,6 +36,7 @@ class Preview extends React.Component {
         propagateErrorsToParent: isLivePreview,
         breakLoops: isLivePreview,
         nonBlockingAlertsAndPrompts: isLivePreview,
+        refreshTimestamp: timestamp,
```
outoftime: I think we’ll want to only pass this in if it’s `isLivePreview` (the pop-out version should generally be as “pure” as possible).

src/util/generatePreview.js
```diff
@@ -122,6 +124,12 @@ class PreviewGenerator {
     this._previewHead.appendChild(styleTag);
   }
 
+  _addRefreshTimestamp(timestamp) {
+    const comment = this.previewDocument.createComment(`Last refresh on: ${String(new Date(timestamp))}`);
+    console.log(comment);
+    this.previewBody.prepend(comment);
```
outoftime: All else being equal I’d probably put this at the end although obviously doesn’t matter too much either way…

kaylee42:
@outoftime hi Mat! This is ready for review again. We are getting one linting error we've never seen before and don't understand so don't know how to resolve: 

```
/home/travis/build/popcodeorg/popcode/src/reducers/ui.js
 42:16  error  Function 'ui' has a complexity of 27  complexity
```
Would appreciate help on this one!

kaylee42:
Okay just read a very helpful article and understand why the case statement has a complexity of 27 - but isn't clear to us what best course of action is. Seems like with that many case statements they would have to be split into another reducer but that feels over the top... What do you think?

alexpelan:
@kaylee42 that rule has to do with cyclomatic complexity: http://eslint.org/docs/rules/complexity - the usual fix is to factor some of the logic out into subroutine, or in this case a sub reducer probably - @outoftime might just be okay with relaxing the requirement in this case though

outoftime:
> Okay just read a very helpful article and understand why the case statement has a complexity of 27 - but isn't clear to us what best course of action is. Seems like with that many case statements they would have to be split into another reducer but that feels over the top... What do you think?

@kaylee42 @ibrand so yeah, as @alexpelan guessed, I’m fine for the time being with just bumping the maximum complexity in `.eslintrc`. There is a [better way to structure reducers](https://github.com/acdlite/redux-actions#handleactionsreducermap-defaultstate), which handles this problem altogether; I’m hoping to do that refactor in the fairly near term.

kaylee42:
@outoftime this is totally finished now (minus merging master in + rebasing), but wanted you to take a look before we did that!

outoftime:
@kaylee42 @ibrand awesome! i don't think I'll have time to look today but definitely tomorrow. can't wait to release this!

src/css/application.css
```diff
@@ -517,6 +517,22 @@ body {
   opacity: 0.5;
 }
 
+.preview__reset-button {
```
outoftime: It seems like this shares enough with `.preview__pop-out-button` to justify a shared component (e.g. `preview__button` with modifiers `preview__button_pop-out` and `preview__button_reset`)

outoftime:
@ibrand @kaylee42 oh also would y’all mind adding a finalized PR description as well? thanks!

outoftime:
@ibrand @kaylee42 hiiiii not to be annoying but this is really close, any sense of when we might be able to get it merged??

Issue #386, opened 2016-10-22
Preview document title should be window title


outoftime:
The title of the Popcode tab should be whatever the student has entered into the `<title>` tag in the HTML, ideally _as long as they have changed it from the default_. No particularly elegant way to satisfy that second requirement comes to mind.

Seems like we could use the [`react-element-portal`](https://github.com/zapier/react-element-portal) library for the time being, and switch to React 16’s built in implementation of portals once that’s out.

ianmclaury:
Giving this a shot

maxjacobson:
@outoftime mind unassigning me?

outoftime:
@maxjacobson done!

outoftime:
I think we’re good here now that the page title is reflected in the preview chrome.

PR #864, opened 2017-06-04
Alexpelan/GitHub repo export

alexpelan:
If the user is logged in and the `experimental` url parameter is set, then a button will be displayed that will export the current state of the editor to a Github repo. The title of the repo is the current `<title>` tag value and a timestamp for uniqueness. The exact source files are pushed to the `master` branch while the preview document that displays in the popout/preview panel is pushed to the `gh-pages` branch as `index.html` so this creates an auto-deployable link.


outoftime:
@pwjablonski Alex is doing some really cool stuff here, but (as he and I discussed IRL today) it’s not obvious what we should be driving to from a product standpoint. I know the idea of pushing to GitHub repos has been floating around for almost as long as Popcode has existed, but do you have any specific thoughts on how a full GitHub repo export might fit into classroom workflows? Or is this a good idea whose time has not quite yet come…?

pwjablonski:
@outoftime This looks really cool! Yeah - I don't think I have fully wrestled with usecases / how it would fit into classroom workflows. Trying to synthesize my thoughts but let me know if it's not clear.

A couple of questions: 
1) What is the intent of this feature. (To give students a working live link with gh-pages, to serve as a place to store students code, or both)
2) Would/could this be a replacement for gists entirely?

Having a both an export repo button and export to gist may seem confusing/redundant to students. 

However - I do think this is really great from a building student portfolio perspective and I like the idea of introducing repositories earlier on.

alexpelan:
@outoftime not quite ready for merge, but this is ready for feedback on both code and a 'how should this work' perspective.

Basically this creates a repo called 'popcode-repo-{timestamp}' for uniqueness and writes any non empty html/css/javascript files as well as the preview file (Preview is always `index.html`). Then it creates a gh-pages branch so the preview file is automatically deployed.

A few questions/comments for you / anyone else who cares to opine:
- Can always tweak wording of repo name / commit messages / descriptions, I just chose things that made sense to me and the timestamp for uniqueness
- In the GenericNotificationWithURL component, since I am passing a translated string I am now getting an error if two of the same notification (gist ready or whatever) appear on screen at the same time - I assume I am violating some i18n best practice, let me know if there's a better way to share a component but pass in the display text.
- Not passing CI due to the cyclomatic complexity of the ui reducer
- Testing could be improved, as we were discussing in parallel on another issue there isn't a lot of testing of the Github integration right now, so the meaty code that creates the repo and files is untested

src/components/notifications/RepoExportNotification.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {t} from 'i18next';
+import GenericNotificationWithURL from './GenericNotificationWithURL';
+
+export default function RepoExportNotification(props) {
```
outoftime: Ibid. `props` destructuring

src/sagas/ui.js
```diff
@@ -29,9 +33,30 @@ export function* exportGist() {
...
+    if (exportWindow.closed) {
+      yield put(repoExportNotDisplayed(url));
+    } else {
+      exportWindow.location.href = url;
+      yield put(repoExportDisplayed());
+    }
+  } else {
+    yield call([exportWindow, 'close']);
+  }
+}
```
outoftime: This looks very similar to `exportGist`—can we extract a generic function that both sagas can call?

outoftime: (Is this more annoying because these are generators?)

src/util/performWithRetries.js
```diff
@@ -0,0 +1,23 @@
+import assign from 'lodash/assign';
+import promiseRetry from 'promise-retry';
+
+export default function performWithRetries(
+  perform,
+  errorsToRetry = [],
```
outoftime: I think this should probably be a function that is passed an error and returns `true` if the error merits a retry

src/sagas/ui.js
outoftime: I think probably not that annoying with `yield*`

alexpelan:
@outoftime thanks, lots of great feedback here - wanted to acknowledge I've seen it but can't respond right away. I think my next step is gonna be to create a separate PR to set up "experimental" mode and get that merged ahead of this - I think I can get to that this weekend.

src/components/Dashboard.jsx
```diff
@@ -75,6 +75,15 @@ class Dashboard extends React.Component {
...
       );
 
+      exportRepoButton = (
+        <div
+          className="dashboard__menu-item dashboard__menu-item_grid"
+          onClick={this.props.onExportRepo}
+        >
+          {t('dashboard.menu.export-repo')}
+        </div>
+      );
```
alexpelan: Done in af6f43f06276456375cf27dda05cf10d8bb642e8

src/clients/gists.js
```diff
@@ -1,25 +1,10 @@
-import assign from 'lodash/assign';
```
alexpelan: Combined `clients`, `repos`, and `services/github` into a single `clients/github` module

src/clients/gists.js
```diff
@@ -1,25 +1,10 @@
...
-      return Promise.reject(error);
-    }),
-    assign({
-      retries: 5,
-      factor: 2,
-      minTimeout: 1000,
-      maxTimeout: 10000,
-    }, options),
-  );
+function gistPerformWithRetries(perform, options = {}) {
```
alexpelan: 👍

src/clients/gists.js
```diff
@@ -103,26 +72,27 @@ function createPopcodeJson(project) {
 }
 
 export async function createFromProject(project, user) {
-  const github = clientForUser(user);
+  const github = gitHub.clientForUser(user);
```
alexpelan: Like with the comment about repos/gists, combined `clients`, `repos`, and `services/github` into a single `clients/github` module

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+import gitHub from '../services/gitHub';
+import generatePreview from '../util/generatePreview';
+import performWithRetries from '../util/performWithRetries';
+
+const COMMIT_MESSAGE = 'Created using Popcode: https://popcode.org';
+const MASTER_BRANCH = 'master';
+
+// Sometimes there are race conditions with creating multiple files in separate
+// API requests
+function repoPerformWithRetries(perform) {
```
alexpelan: just went with `performWithRetryNotFoundOrNetworkErrors`, combined with the feedback below about also retrying on Network.

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+  return performWithRetries(perform, ['Not Found']);
+}
+
+async function createHtmlFile(github, project, userName, repoName) {
```
alexpelan: 👍

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+    return github.getRepo(userName, repoName).writeFile(
+      MASTER_BRANCH, 'script.js', project.sources.js, COMMIT_MESSAGE, {});
+  }
+  return Promise.resolve();
+}
+
+async function createPreviewFile(github, project, userName, repoName) {
+  if (trim(project.sources.html) ||
+    trim(project.sources.css) ||
+    trim(project.sources.javascript)) {
```
alexpelan: Removed check

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+    const preview = generatePreview(
+      project,
+      {
+        targetBaseTop: false,
+        propagateErrorsToParent: false,
+        breakLoops: false,
+        nonBlockingAlertsAndPrompts: false,
+      },
```
alexpelan: Ok

outoftime:
@alexpelan heyo, noticed a bunch of recent work here—this ready for another review pass?

alexpelan:
@outoftime hopefully today!

outoftime:
@alexpelan ok cool! no rush, just wanted to make sure I wasn’t holding it up

src/sagas/ui.js
alexpelan: Wasn't bad

src/util/performWithRetries.js
alexpelan: Resolved (sorry, just need to type these in to manage the to-do list of various comments - I don't think there's a discrete way to do this in github code review?)

src/components/notifications/RepoExportNotification.jsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <GenericNotificationWithURL
+      payload={props.payload}
+      text={t('notifications.repo-export-complete')}
+      type={props.type}
+      urlText={t('notifications.gist-export-link')}
```
alexpelan: github-export-link for now

src/components/notifications/RepoExportNotification.jsx
alexpelan: Done

src/components/notifications/GenericNotificationWithURL.jsx
```diff
@@ -0,0 +1,20 @@
...
+      </a>
+    </span>
+  );
+}
+
+GenericNotificationWithURL.propTypes = {
+  payload: PropTypes.object.isRequired,
+  text: PropTypes.string.isRequired,
+  type: PropTypes.string.isRequired,
+  urlText: PropTypes.string.isRequired,
```
alexpelan: Done

src/components/notifications/GenericNotificationWithURL.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function GenericNotificationWithURL(props) {
```
alexpelan: Very nice

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+    );
+
+  const userName = response.data.owner.login;
+
+  await createFiles(github, project, userName, REPO_NAME);
+  await repoPerformWithRetries(
+    () => createBranch(github, userName, REPO_NAME, MASTER_BRANCH, 'gh-pages'),
+  );
+
+  // we don't know the github username at time of repo creation so update here
```
alexpelan: Removed comment

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+      description: url,
+      homepage: url,
+    });
+}
+
+export async function createRepoFromProject(project, user) {
+  const github = gitHub.clientForUser(user);
+  const REPO_NAME = `popcode-repo-${Date.now()}`;
+
+  const response =
```
alexpelan: Done

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+  return github.getRepo(userName, repoName).
+    createBranch(oldBranch, newBranch);
+}
+
+async function updateRepoDescription(github, userName, repoName) {
+  const url = `https://${userName}.github.io/${repoName}`;
+  return github.getRepo(userName, repoName).
+    updateRepository({
+      name: repoName,
+      description: url,
```
alexpelan: As far as I can tell, you cannot use markdown in the Github repository description

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+  );
+  await repoPerformWithRetries(
+    () => createJsFile(github, project, userName, repoName),
+  );
+  await repoPerformWithRetries(
+    () => createPreviewFile(github, project, userName, repoName),
+  );
+}
+
+async function createBranch(github, userName, repoName, oldBranch, newBranch) {
```
alexpelan: inlined

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+  );
+  await repoPerformWithRetries(
+    () => createCssFile(github, project, userName, repoName),
+  );
+  await repoPerformWithRetries(
+    () => createJsFile(github, project, userName, repoName),
+  );
+  await repoPerformWithRetries(
+    () => createPreviewFile(github, project, userName, repoName),
+  );
```
alexpelan: I tried to do this, and hit the same thing that I hit on initial implementation - the Github API really doesn't like when you do this, it gets pretty confused when the repo has changed between when you've fetched it and you write to it.

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+      project,
+      {
+        targetBaseTop: false,
+        propagateErrorsToParent: false,
+        breakLoops: false,
+        nonBlockingAlertsAndPrompts: false,
+      },
+    );
+    return github.getRepo(userName, repoName).
+      writeFile(MASTER_BRANCH, 'index.html', preview, COMMIT_MESSAGE, {});
```
alexpelan: Now we create the gh-pages branch upfront, before creating any files in master, and then write the preview only to gh-pages

alexpelan: (A side effect of this is we now create the repo with `auto_init` set to true - this touches an empty README. Otherwise the API doesn't like you creating and pushing a branch in a repo with nothing in it.

src/clients/repos.js
```diff
@@ -0,0 +1,122 @@
...
+    updateRepository({
+      name: repoName,
+      description: url,
+      homepage: url,
+    });
+}
+
+export async function createRepoFromProject(project, user) {
+  const github = gitHub.clientForUser(user);
+  const REPO_NAME = `popcode-repo-${Date.now()}`;
```
alexpelan: I took the repo name from the title, naively replacing spaces with '-' for now - I'm sure there are all sorts of other invalid github repository names though, how do we want to handle that?

And I left the timestamp in for now, in the future we'll somehow want to link repositories and send new exports are updates to the existing repo but at this point I don't want to touch that

alexpelan:
Okay,  this is ready for another review. The tests are failing, but I am out of time to work on this for probably a week or so so wanted to pass the baton back to you. Something in the saga refactor to share one between export gist and export repo is completely blowing the tests up, but both are working when I try them in the app - if you could take a look at that part that would be helpful.

src/clients/github.js
```diff
@@ -0,0 +1,280 @@
...
+function _canUpdateGist(user) {
+  return Boolean(_getGithubToken(user));
+}
+
+export {
+  createGistFromProject,
+  createRepoFromProject,
+  EmptyGistError,
+  loadFromId,
+};
```
outoftime: I know this style is still used in some places in the code, but my preference is for the `export` statements to go inline, e.g.:

```js
export async function createGistFromProject() {
  …
```

(That solves the problem of being able to easily tell which functions are part of the module’s public interface without needing the `_` notation)

src/sagas/clients.js
```diff
@@ -1,22 +1,44 @@
...
+  createRepoFromProject,
+} from '../clients/github';
+
+import {
+  gistExported,
+  gistExportError,
+  repoExported,
+  repoExportError,
+  } from '../actions/clients';
+import {getCurrentProject} from '../selectors/';
```
outoftime: Why the trailing `/` here?

Also let’s be consistent about whitespace between `import` statements (at least in the context of the same file)

src/sagas/ui.js
```diff
@@ -3,7 +3,11 @@ import debounceFor from 'redux-saga-debounce-effect/src/debounceFor';
 import {TextEncoder} from 'text-encoding';
 import base64 from 'base64-js';
 import {userDoneTyping as userDoneTypingAction} from '../actions/ui';
-import {gistExportDisplayed, gistExportNotDisplayed} from '../actions/clients';
+import {
+  gistExportDisplayed,
+  gistExportNotDisplayed,
+  repoExportDisplayed,
+  repoExportNotDisplayed} from '../actions/clients';
```
outoftime:
```js
  repoExportNotDisplayed,
} from '../actions/clients';
```

src/clients/github.js
```diff
@@ -0,0 +1,280 @@
...
+}
+
+async function _createHtmlFile(github, project, userName, repoName) {
+  if (trim(project.sources.html)) {
+    return _createSourceFile(
+      github,
+      userName,
+      repoName,
+      MASTER,
+      'raw_index.html',
```
outoftime: Why not just `index.html`?

alexpelan:
Yeah, I just figured out the tests, not sure why it was a problem in the test suite but not in the application.

src/clients/github.js
```diff
@@ -0,0 +1,280 @@
...
+
+  const {data} =
+    await _performWithRetryNotFoundOrNetworkErrors(
+      () => github.getUser().createRepo({name: REPO_NAME, auto_init: true}),
+    );
+
+  const userName = data.owner.login;
+
+  await _performWithRetryNotFoundOrNetworkErrors(
+    () => _createBranch(github, userName, REPO_NAME, MASTER, GH_PAGES),
```
alexpelan: I couldn't figure out how to make it work, if you look at the source it's not really 'optional' in that you don't need an old branch, there's just a default if you don't fill it in: http://github-tools.github.io/github/docs/3.1.0/Repository.js.html#line514

src/clients/github.js
```diff
@@ -0,0 +1,280 @@
...
+
+function EmptyGistError(message) {
+  this.name = 'EmptyGistError';
+  this.message = message;
+  this.stack = new Error().stack;
+}
+EmptyGistError.prototype = Object.create(Error.prototype);
+
+function _getPageTitle(preview) {
+  return _parser.parseFromString(
```
alexpelan: Is there? I looked to see if #386 had been resolved yet, figuring I could use whatever they did, but it has not been.

src/clients/github.js
```diff
@@ -0,0 +1,280 @@
+import get from 'lodash/get';
+import GitHub from 'github-api';
+import isEmpty from 'lodash/isEmpty';
+import trim from 'lodash/trim';
+import performWithRetries from '../util/performWithRetries';
+import generatePreview from '../util/generatePreview';
+
+const _anonymousGithub = new GitHub({});
```
alexpelan: Think I got them all

src/clients/github.js
```diff
@@ -0,0 +1,280 @@
...
+
+async function _createJsFile(github, project, userName, repoName) {
+  if (trim(project.sources.javascript)) {
+    return _createSourceFile(
+      github,
+      userName,
+      repoName,
+      MASTER,
+      'script.js',
+      project.sources.js,
```
alexpelan: yes, yikes.

src/clients/github.js
alexpelan: Vestigial from when we created the raw_index.html and index.html (preview) on the same branch. Fixed

alexpelan:
I wasn't able to reproduce the 404 error that was causing a promise failure, it was either from a prior version of the code where a 404 was causing promise rejection, or I just saw 404 errors happening (there are some 404 GETs in the network tab if you watch while exporting a repo, i think the code checks for existence of a file before creating it via a separate call) and thought that they had to be retried when they do not

So I think the only outstanding issue now is the multiple branch / `auto_init` thing?

src/clients/github.js
```diff
@@ -0,0 +1,276 @@
...
+export async function createRepoFromProject(project, user) {
+  const github = clientForUser(user);
+  const preview = generatePreview(project);
+  const title = getPageTitle(preview);
+
+  const REPO_NAME = `${title}-${Date.now()}`;
+
+  const {data} =
+    await performWithRetryNetworkErrors(
+      () => github.getUser().createRepo({name: REPO_NAME, auto_init: true}),
```
outoftime: @alexpelan so what happens if we just remove `auto_init` and also remove the `createBranch` below? Will GitHub let us implicitly create the branch when we first call `writeFile`? If I’m understanding the GH library source code directly I think it might work?

alexpelan: Doesn't seem to work.

First it does a GET for `https://api.github.com/repos/alexpelan/Page-Title-1499912981497/contents/index.html?ref=gh-pages`

response: 

    {
      "message": "No commit found for the ref gh-pages",
      "documentation_url": "https://developer.github.com/v3/repos/contents/"
    }

Then a PUT for `https://api.github.com/repos/alexpelan/Page-Title-1499912981497/contents/index.html`

response:

    {
       "message": "Branch gh-pages not found",
       "documentation_url": "https://developer.github.com/v3/repos/contents/"
    }

outoftime: Rats! okay can we just delete the README after initializing the repo…?

src/sagas/ui.js
```diff
@@ -12,26 +17,39 @@ export function* userDoneTyping() {
   yield put(userDoneTypingAction());
 }
 
-export function* exportGist() {
+function* _githubExport(
```
alexpelan: Fixed

src/sagas/ui.js
```diff
@@ -12,26 +17,39 @@ export function* userDoneTyping() {
...
       exportWindow.location.href = url;
-      yield put(gistExportDisplayed());
+      yield put(displayedAction());
     }
   } else {
     yield call([exportWindow, 'close']);
   }
 }
 
+export function exportGist() {
```
alexpelan: I don't think so...I originally had these as generators which yielded githubExport, but the spinner didn't render right and the tests failed because there was an extra `.next()` needed to call - oddly enough the gist/repo still got exported properly at the end, don't really know enough about sagas/generators to know why that was. Anyway, this function and `exportRepo` instead return a generator, instead of being a generator themself

src/clients/github.js
alexpelan: yep, this seems to work, I guess there just needs to be a commit, even if the repo is empty?

alexpelan:
@outoftime okay, I think we are good? unless you are in disagreement about the generators.

apologies to any subscribers to the repo for the now 90something notifications you've gotten about this

src/sagas/ui.js
outoftime: I think what you want is a `yield* githubExport`—the `yield*` basically says “here is another generator, put that in charge of generating until it completes”.

I think having a non-generator function that `return`s here might only work as an artifact of the transpilation process—and even if it’s not the case, I’d argue that it’s generally pretty confusing to see a non-generator function `return` a generator.

alexpelan: Yep, that still seems to work

src/clients/github.js
```diff
@@ -0,0 +1,280 @@
...
+
+async function createSourceFileInRepo(
+  github,
+  userName,
+  repoName,
+  branchName,
+  fileName,
+  source,
+) {
+  return github.getRepo(userName, repoName).
```
outoftime: So this `return` here works fine and does the right thing, because if an `async` function body returns a Promise object explicitly, then that becomes the await-able Promise and all is good.

HOWEVER, I think from a semantic standpoint, we should `await` rather than `return`, because the resolved *value* of the function (which is to say, the resolved value of the promise returned by `writeFile`) is not meaningful. `await` clearly communicates that “this asynchronous operation needs to complete for me to be considered done”, rather than `return` which implies that there is a result value here that someone might care about. It also helps make the `createHtmlFile` etc. functions a bit more concise, as I will mention below…

src/clients/github.js
```diff
@@ -0,0 +1,280 @@
...
+    return createSourceFileInRepo(
+      github,
+      userName,
+      repoName,
+      MASTER,
+      'index.html',
+      project.sources.html,
+    );
+  }
+  return Promise.resolve();
```
outoftime: Changing the `return` on line 188 to an `await` means we don’t need this line either, which seems right.

src/clients/github.js
outoftime: Well not _all_ of them…

src/clients/github.js
```diff
@@ -0,0 +1,280 @@
...
+  const REPO_NAME = `${title}-${Date.now()}`;
+
+  const {data} =
+    await performWithRetryNetworkErrors(
+      () => github.getUser().createRepo({name: REPO_NAME, auto_init: true}),
+    );
+
+  const userName = data.owner.login;
+
+  await performWithRetryNetworkErrors(
```
outoftime: It occurs to me that elsewhere in this module, we generally put the `performWithRetryNetworkErrors` directly around the call to the `github` SDK, which I think makes sense. Seems like that would make this method less verbose and reduce the overall number of times we need to write `performWithRetryNetworkErrors` in this file (since several of the below methods end up actually calling the SDK in `createSourceFileInRepo`)

src/clients/github.js
outoftime: yeah, according to [this Stack Overflow thread](https://stackoverflow.com/questions/18565425/how-do-i-create-an-orphan-branch-from-the-github-api) it isn’t possible to create an orphan branch via the GitHub API. obviously it’s not ideal to have the two commits adding and removing the README in the history, but i’m much less concerned about that than about having the README actually present in the repo’s files. so I think we’re good here : )

outoftime:
Oh, one other thing @alexpelan, can you update the PR description to describe the changes herein for posterity? Thank you!!

alexpelan:
Ok, ready for another look

alexpelan:
Oh, and if you want me to rewrite some git history let me know

alexpelan:
@outoftime alright, we are good to go!

PR #873, opened 2017-06-11
Use presentational/container pattern in Output component subtree

outoftime:
Applies the [Presentational and Container Components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) pattern to the component subtree starting with `<Output>`. Summary of changes:

* Introduces `PreviewContainer` and `ErrorReportContainer`, which are container components constructed purely out of the `redux-saga` `connect()` function. These control the `<Preview>` and `<ErrorReport>` (formerly `<ErrorList>`) presentational components.
* Renames `<ErrorList>` to `<ErrorReport>` and `<ErrorSublist>` to `<ErrorList>`
* Moves responsibility for popping preview out into a window to the `ui` saga, rather than handling it directly in the `<Preview>` component
* Consolidates all errors—both validation and runtime—under the `errors` key in the store. Each error list can now have either a `validation-error` or a `runtime-error` state (vs. previously just having a `failed` state for validation errors); the UI various behavior accordingly.
* Makes `<Preview>` and `<ErrorList>` pure functional components

This is proposed as the first step to a complete refactor that gives responsibility for React-Redux mediation entirely to connected container components. The end result of this would be for `<Workspace>` to become a purely presentational component. The minimal remaining steps would be to make a `DashboardContainer` and an `EditorsContainer`.

So far I am finding the use of container components very satisfying—we’re removing substantial complexity from elsewhere in the code but the containers themselves don’t feel complex at all. It’s also nice to be able to unwind the massive callback chains that currently lead us back to the `<Workspace>` from all corners of the code.

First step in #881


outoftime:
@inlinestyle this needs some cleanup/a few more tests but I’d be interested in your feedback on the broad outline of the refactor. I am feeling p. good about it!

src/components/notifications/GistExportNotification.jsx
```diff
@@ -15,5 +15,6 @@ export default function GistExportNotification(props) {
 
 GistExportNotification.propTypes = {
   payload: PropTypes.object.isRequired,
+  // eslint-disable-next-line react/no-unused-prop-types
```
outoftime: Hmm, good question—the reason I left it in there is that the abstract `ExportNotification` component type does take a `type` key. But it occurs to me that neither `PropTypes` nor `eslint-plugin-react` complains if you pass _extra_ propTypes to a component (although I wish they did!) so I guess there’s no reason to keep this in.

outoftime:
> Minor suggestion: for the containers in `src/containers`, it's a bit redundant to have "container" in the name. For all practical purposes it's fine to have `src/containers/Preview` and `src/containers/ErrorList`.

@inlinestyle so we’re OK with having e.g. `src/containers/Preview` and `src/components/Preview`?

Practically speaking I doubt we would ever import both into the same file, and we have the same sort of parallel naming schemes with e.g. actions and sagas anyway, so I don’t see much problem with it.

PR #874, opened 2017-06-11
Updated gist export with shareable link though rawgit and bit.ly

pwjablonski:
Had fun trying to do this. May not make sense if we are trying to do it through gh-pages but seems like a viable temporary alternative. 

The current flow...
- on export of gist, use gistData to make raw-git-url
- uses bit.ly to shorten the url. (I looked for custom urls but you need an enterprise account)
- Appends the shareable URL to the gist description (may not be the most elegant place for this but definitely the easiest)

Also, I've haven't written tests before so would be interested in learning how to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,8 @@ for all the sordid details.
 * [Frederic Brodbeck](http://www.freder.io/)
 * [Ben Yelsey](https://github.com/inlinestyle)
 * [Aaron Greenberg](https://github.com/ajgreenb)
+* [Peter Jablonski](https://github.com/pwjablonski)
+* [Ten-Young Guh](https://github.com/tenyoung795)
 * [Katie Conneally](http://www.katieconneally.com/) created the name Popcode
 * Logo design, "Pop" concept, and UI by the team at
   [Red Peak](http://redpeakgroup.com): Andrew Haug, Aya Kawabata, Jieun Lee,
--- a/package.json
+++ b/package.json
@@ -69,6 +69,15 @@
     {
       "name": "Aaron Greenberg",
       "url": "https://github.com/ajgreenb"
+    },
+    {
+      "name": "Peter Jablonski",
+      "email": "pwjablonski@gmail.com",
+      "url": "https://github.com/pwjablonski"
+    },
+    {
+      "name": "Ten-Young Guh",
+      "url": "https://github.com/tenyoung795"
     }
   ],
   "keywords": "code editor education learning",


pwjablonski:
@outoftime 

Yeah I see that now. I think creating a separate preview.html makes the most sense. Seems like a relatively simple fix. 

Definitely, has problems as we head into multiple pages but can work for now. I think repos and gh-pages will most likely make the most sense long term.

outoftime:
@pwjablonski word up.

in terms of tests—i’d definitely like to have a good system for testing client logic, but as of now it’s tough because they directly pull in external dependencies and there is no DI system set up. so the only option is to mock the dependencies which doesn’t feel great. for that reason (among others) there currently aren’t any tests for the clients. interested in what @inlinestyle @ajgreenb @alexpelan @jwang1919 think about that issue

pwjablonski:
@outoftime Put the preview together. Handling libraries this way doesn't seem the most elegant way of doing it. Also the preview isn't formatted correctly.

src/clients/gists.js
```diff
@@ -81,16 +87,78 @@ function canUpdateGist(user) {
...
+  if (project.enabledLibraries.includes('bootstrap')) {
+    const bootstrapLink = document.createElement('link');
+    bootstrapLink.setAttribute('href', 'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css');
+    bootstrapLink.setAttribute('rel', 'stylesheet');
+    previewHtml.getElementsByTagName('head')[0].appendChild(bootstrapLink);
+
+    const bootstrapScript = document.createElement('script');
+    bootstrapScript.setAttribute('src', 'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js');
+    previewHtml.getElementsByTagName('head')[0].appendChild(bootstrapScript);
+  }
```
outoftime: So is the reason we’re not using [`generatePreview`](https://github.com/popcodeorg/popcode/blob/master/src/util/generatePreview.js) the fact that we want to use CDN URLs for the library dependencies rather than inlining them?

If so—and we feel like it’s important to not just inline the libraries—I think we should enhance `generatePreview` so that it can use either inline or external libraries. This would presumably entail updating [`libraries.js`](https://github.com/popcodeorg/popcode/blob/master/src/config/libraries.js) to include CDN URLs for each configured library. We’d also want to make sure we have the same version of each library internally and externally (side note: Bower dependencies are woefully out of date; we should probably use exact versions in `bower.json` since Bower doesn’t have a lockfile mechanism)

pwjablonski:
Hey @outoftime. Yeah Alex is doing some great stuff there. I think it makes sense to sunset this PR. The gh-pages route is definitely the better route.

My questions, though they probably don't belong on this thread, are... 
- how does the repo export fit into classroom workflows? 
- Does it exist alongside or instead of gist exports?

src/clients/gists.js
pwjablonski: Yeah that was the thought. Definitely an ugly solution.  Anyway, I think it makes sense to stop working on this, but it was fun learning more about this process!

Issue #378, opened 2016-10-22
No errors when high-level document structure is wrong


outoftime:
Things that Popcode currently allows:
### `head` inside `body`

``` html
<!doctype html>
<html>
<body>
  <head><title></head>
</body>
</html>
```
### `body` inside `head`

``` html
<!doctype html>
<html>
  <head>
    <body></body>
  </head>
</html>
```
### Text outside root

``` html
<!doctype html>
Hello, world!
<html>
  <head>
    <body></body>
  </head>
</html>
```
### Text inside `head` tag:

``` html
<!doctype html>
<html>
<head>
        <title>Page Title</title>
        asdf
</head>
<body>
</body>
</html>
```
### Text directly inside root node

``` html
<!doctype html>
<html>
Hello, world!
<head>
  <title>Something</title>
</head>
<body>
</body>
</html>
```
### Content tag inside `<head>`

``` html
<!doctype html>
<html>
<head>
  <title>Something</title>
  <p>Hello, world!</p>
</head>
<body>
</body>
</html>
```
### Content tag between `<head>` and `<body>`

``` html
<!doctype html>
<html>
<head>
  <title>Something</title>
</head>
<p>Hello, world!</p>
<body>
</body>
</html>
```
### Content tag outside document structure

``` html
<!doctype html>
<p>Hello, world!</p>
<html>
<head>
  <title>Something</title>
</head>
<body>
</body>
</html>
```

Unfortunately, none of our linters/parsers actually complain about these problems.

A couple of possible approaches come to mind; neither is easy:
- Submit an upstream patch to [slowparse](https://github.com/mozilla/slowparse) to add errors when this happens. I haven’t looked at slowparse’s code much, but philosophically this seems like the sort of error it should catch. It also may be abandoned at this point (but we can always just target a fork)
- Use the `Parser` class in [`html-tokenizer`](https://github.com/greim/html-tokenizer). This generates a DOM tree structure from an input string. The good news is, unlike the parsers built into browsers, it _doesn’t_ enforce a valid DOM structure in its output—so if your HTML has a `<body>` tag inside a `<head>` tag, it’ll give you a DOM tree with a `body` element inside a `head` element. Obviously we’d need to hand-roll the actual validations, but that seems tolerable, and we could probably get some low-hanging fruit by implementing some very simple rules (e.g. `<html>` contains only `<head>` and `<body>`; `<head>` and `<body>` are direct descendants of `<html>`
[...]

PR #876, opened 2017-06-12
Deprecate mismatched-close-tag into more precise errors

### LINKED ISSUE #562: Misleading error when a closing tag is missing
Submitter: Kate Miller
Here is the code:

```html
<!DOCTYPE html>
<html>
    <head>
        <title>Page Title</title>
    </head>
    <body>
     <h1>My Web Page</h1>
        <p><a href="//google.com">
            Go to Google
    </a>   
    </body>
</html>
```

Here is the error:

```
11
Your <p> tag needs a closing tag, but instead there’s a </body> tag where the </p> tag should be.
```

This error was confusing, because in reading it without thinking about it, my first instinct was to replace the closing body tag with a closing p tag which then of course led me to another error...

```
12
Your <body> tag needs a closing tag, but instead there’s a </html> tag where the </body> tag should be.
```

Which led me to replace the closing html tag with a closing body tag...

I think the wording of that original error is confusing. By saying there is a closing body tag where the closing p tag should be immediately make you think replacement. A better error message might be "closing p tag is missing"

tenyoung795:
An attempt to fix #562, although goes a bit farther by deprecating `mismatched-close-tag` in favor of three distinct errors:

- `misplaced-close-tag`: **(new)** suggests moving a spare unopened close tag
- `unclosed-tag-by-mismatched-close-tag`: **(new)** suggests adding a close tag right before the mismatched close tag
- `unexpected-close-tag`: preexisting

This pull request introduces a new HTML `Validator` that performs its own lexical analysis using the [SAX](https://en.wikipedia.org/wiki/Simple_API_for_XML) parser from [`parse5`](https://github.com/inikulin/parse5) to parse HTML tokens.

The code is a bit complex, so feel free to offer readability suggestions. Thanks!

### DIFF
--- a/src/validations/html/rules/MismatchedTag.js
+++ b/src/validations/html/rules/MismatchedTag.js
@@ -1,0 +1,86 @@
+// If any spare unopened tag exists afterward, diagnose it and only yield
+// MISPLACED_CLOSE_TAG.
+//
+// Otherwise:
+//
+// A mismatched close tag that has been opened means the case
+//
+// <div>
+//   <span>
+// </div>
+//
+// so yield UNCLOSED_TAG.
+//
+// A mismatched close tag that has yet to be opened means the case
+//
+// <div>
+//   </span>
+// </div>
+//
+// so yield UNOPENED_TAG.
+
+import findLastIndex from 'lodash/findLastIndex';
+import Code from '../rules/Code';
+
+export default class {
+  constructor() {
+    this._openTagStack = [];
+    this._mismatchStacksByOpenName = new Map();
+    this._closeTagMisplacements = [];
+    this._unopenedTags = [];
+  }
+
+  openTag(location, {name}) {
+    this._openTagStack.push({location, name});
+  }
+
+  closeTag(location, name) {
+    const openIndex = findLastIndex(
+      this._openTagStack,
+      openTag => openTag.name === name,
+    );
+
+    const isOpened = openIndex >= 0;
+    if (isOpened) {
+      for (let i = openIndex + 1; i < this._openTagStack.length; ++i) {
+        const openTag = this._openTagStack[i];
+        const mismatch = {
+          openTag,
+          closeTag: {location, name},
+        };
+        const mismatches = this._mismatchStacksByOpenName.get(openTag.name);
+        if (mismatches) {
+          mismatches.push(mismatch);
+        } else {
+          this._mismatchStacksByOpenName.set(openTag.name, [mismatch]);
+        }
+      }
+      this._openTagStack.splice(openIndex);
+    } else {
+      const mismatchStack = this._mismatchStacksByOpenName.get(name) || [];
+      const matchingMismatch = mismatchStack.pop();
+      if (matchingMismatch) {
+        const {openTag, closeTag} = matchingMismatch;
+        this._closeTagMisplacements.push({match: location, openTag, closeTag});
+      } else {
+        this._unopenedTags.push({location, name});
+      }
+    }
+  }
+
+  * done() {
+    for (const mismatches of this._mismatchStacksByOpenName.values()) {
+      for (const {openTag, closeTag} of mismatches) {
+        yield {code: Code.UNCLOSED_TAG, openTag, closeTag};
+      }
+    }
+
+    for (const {openTag, closeTag, match} of this._closeTagMisplacements) {
+      yield {code: Code.MISPLACED_CLOSE_TAG, openTag, closeTag, match};
+    }
+
+    for (const closeTag of this._unopenedTags) {
+      yield {code: Code.UNOPENED_TAG, closeTag};
+    }
+  }
+}
--- a/src/validations/html/rules/index.js
+++ b/src/validations/html/rules/index.js
@@ -1,0 +1,51 @@
+import Validator from '../../Validator';
+import runRules from '../runRules';
+import Code from './Code';
+import MismatchedTag from './MismatchedTag';
+
+const errorMap = {
+  [Code.MISPLACED_CLOSE_TAG]: ({openTag, closeTag}) => ({
+    reason: 'misplaced-close-tag',
+    payload: {
+      open: openTag.name,
+      close: closeTag.name,
+      mismatch: closeTag.location.row + 1,
+    },
+  }),
+  [Code.UNCLOSED_TAG]: ({openTag: {name}}) => ({
+    reason: 'unclosed-tag',
+    payload: {tag: name},
+  }),
+  [Code.UNOPENED_TAG]: ({closeTag: {name}}) => ({
+    reason: 'unexpected-close-tag',
+    payload: {tag: name},
+  }),
+};
+
+class RuleValidator extends Validator {
+  constructor(source) {
+    super(source, 'html', errorMap);
+  }
+
+  _keyForError({code}) {
+    return code;
+  }
+
+  async _getRawErrors() {
+    return Array.from(await runRules([new MismatchedTag()], this._source));
+  }
+
+  _locationForError(error) {
+    switch (error.code) {
+      case Code.MISPLACED_CLOSE_TAG:
+        return error.match;
+      case Code.UNOPENED_TAG:
+      case Code.UNCLOSED_TAG:
+        return error.closeTag.location;
+      default:
+        throw new Error(`Unexpected code in ${JSON.stringify(error)}`);
+    }
+  }
+}
+
+export default source => new RuleValidator(source).getAnnotations();
--- a/src/validations/html/runRules.js
+++ b/src/validations/html/runRules.js
@@ -1,0 +1,40 @@
+import {SAXParser} from 'parse5';
+import voidElements from 'void-elements';
+
+// Runs `rules` on `source` and promises an iterable of all errors.
+//
+// `rules` must be an array of objects that satisfy the interface:
+// * done() returns an iterable of errors
+// * optionally,
+//   openTag({row: number, column: number}, {name: string})
+//   to take in open tags that are neither void nor self-closing
+// * optionally,
+//   closeTag({row: number, column: number}, name: string)
+//   to take in close tags
+export default (rules, source) => {
+  const parser = new SAXParser({locationInfo: true});
+  return new Promise((resolve) => {
+    parser.on('startTag', (name, attrs, selfClosing, {line, col}) => {
+      for (const rule of rules) {
+        if (rule.openTag && !selfClosing && !(name in voidElements)) {
+          rule.openTag({row: line - 1, column: col - 1}, {name});
+        }
+      }
+    });
+    parser.on('endTag', (name, {line, col}) => {
+      for (const rule of rules) {
+        if (rule.closeTag) {
+          rule.closeTag({row: line - 1, column: col - 1}, name);
+        }
+      }
+    });
+    parser.write(source);
+    parser.end(() => {
+      resolve(function* () {
+        for (const rule of rules) {
+          yield* rule.done();
+        }
+      }());
+    });
+  });
+};
--- a/locales/en/translation.json
+++ b/locales/en/translation.json
@@ -102,7 +102,6 @@
       "empty-title-element": "You have an empty <title> tag. A <title> tag should always contain text.",
       "html-in-css-block": "It looks like you typed some HTML code inside a <style> tag. Remember that only CSS code can go inside a <style> tag.",
       "invalid-attribute-name": "{{attribute}} isn’t allowed as an attribute name.",
-      "mismatched-close-tag": "Your <{{open}}> tag needs a closing tag, but instead there’s a </{{close}}> tag where the </{{open}}> tag should be.",
       "self-closing-non-void-element": "{{tag, en-handle-an|capitalize}} <{{tag}}> tag always needs a closing </{{tag}}> tag. Writing <{{tag}}/> is not allowed.",
       "unclosed-tag": "Your <{{tag}}> tag needs to be closed by an </{{tag}}> tag.",
       "unexpected-close-tag": "You have a closing </{{tag}}> tag that doesn’t match any opening <{{tag}}> tag.",
--- a/locales/en/translation.json
+++ b/locales/en/translation.json
@@ -112,7 +112,8 @@
       "unterminated-comment": "You have a comment that never ends. End the comment with -->",
       "unbound-attribute-value": "The value {{value}} isn’t attached to an attribute. Are you missing an = sign?",
       "invalid-tag-name": "<{{tag}}> isn’t a valid HTML tag. If you want to create a custom tag, make sure the name has a - in it, like <my-tag>",
-      "space-before-tag-name": "You can’t put any space after the < in an HTML tag. Write <{{tag}}> instead of < {{tag}}>."
+      "space-before-tag-name": "You can’t put any space after the < in an HTML tag. Write <{{tag}}> instead of < {{tag}}>.",
+      "misplaced-close-tag": "This closing </{{open}}> tag should go to before the </{{close}}> tag on line {{mismatch}}."
     },
     "css": {
       "missing-opening-curly": "You should have a { at the end of this line.",
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     "bugsnag-js": "^3.0.1",
     "classnames": "^2.1.5",
     "css": "^2.2.1",
+    "enumify": "^1.0.4",
     "es6-set": "^0.1.4",
     "esprima": "^3.0.0",
     "firebase": "^4.1.3",
--- a/package.json
+++ b/package.json
@@ -120,6 +120,7 @@
     "loop-breaker": "^0.1.0-alpha.7",
     "moment": "^2.14.1",
     "offline-plugin": "^4.8.1",
+    "parse5": "^3.0.2",
     "pify": "^3.0.0",
     "promise-retry": "^1.1.1",
     "prop-types": "^15.5.10",
@@ -140,7 +141,8 @@
     "rxjs": "^5.0.2",
     "slowparse": "^1.1.4",
     "stylelint": "^7.10.1",
-    "text-encoding": "^0.6.0"
+    "text-encoding": "^0.6.0",
+    "void-elements": "^3.1.0"
   },
   "engines": {
     "node": "7.6.0",
--- a/src/validations/html.js
+++ b/src/validations/html.js
@@ -2,10 +2,12 @@ import trim from 'lodash/trim';
 import mergeValidations from './mergeValidations';
 import validateWithHtmlInspector from './html/htmlInspector';
 import validateWithHtmllint from './html/htmllint';
+import validateWithRules from './html/rules';
 import validateWithSlowparse from './html/slowparse';
 
 export default source => mergeValidations([
   validateWithHtmlInspector(source),
   validateWithHtmllint(source),
   validateWithSlowparse(trim(source)),
+  validateWithRules(source),
 ]);
--- a/src/validations/html/htmllint.js
+++ b/src/validations/html/htmllint.js
@@ -64,7 +64,6 @@ const errorMap = {
     return {
       reason: 'unclosed-tag',
       payload: {tag},
-      suppresses: ['mismatched-close-tag'],
     };
   },
 };
--- a/src/validations/html/rules/Code.js
+++ b/src/validations/html/rules/Code.js
@@ -1,0 +1,8 @@
+import {Enum} from 'enumify';
+
+export default class Code extends Enum {}
+Code.initEnum([
+  'MISPLACED_CLOSE_TAG',
+  'UNCLOSED_TAG',
+  'UNOPENED_TAG',
+]);
--- a/src/validations/html/slowparse.js
+++ b/src/validations/html/slowparse.js
@@ -34,6 +34,7 @@ const errorMap = {
         return {
           reason: 'space-before-tag-name',
           payload: {tag: tagMatch[1]},
+          suppresses: ['unexpected-close-tag'],
         };
       }
     }
@@ -55,22 +56,6 @@ const errorMap = {
     payload: {attribute: error.attribute.name.value},
     suppresses: ['lower-case-attribute-name'],
   }),
-
-  MISMATCHED_CLOSE_TAG: (error) => {
-    const openTagName = error.openTag.name;
-    const closeTagName = error.closeTag.name;
-    if (openTagName === '#document-fragment') {
-      return {
-        reason: 'unexpected-close-tag',
-        payload: {tag: closeTagName},
-      };
-    }
-
-    return {
-      reason: 'mismatched-close-tag',
-      payload: {open: openTagName, close: closeTagName},
-    };
-  },
 
   SELF_CLOSING_NON_VOID_ELEMENT: error => ({
     reason: 'self-closing-non-void-element',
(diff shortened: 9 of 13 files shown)


outoftime:
@tenyoung795 this is cool as hell!!

I haven’t gone over it in detail yet but one bit of high-level feedback—I think it would be best, instead of thinking of this as a “custom validation”, rather to simply think of it as another validation backend that happens to have been written in-house. In particular, I would propose splitting out the nitty-gritty implementation of the DOM traversal from the adapter code in `src/validators`.

I think the ideal approach would be to create a new NPM module and set it up as a dependency of Popcode (much like [`loop-breaker`](https://github.com/popcodeorg/loop-breaker)). I’d be happy to create a new repo under the `popcodeorg` namespace and make you an owner if you’d like to go that route. As a more practical compromise, I’d also be down to just extract the parsing/traversing code into its own module in the popcode application.

LMK your thoughts!

tenyoung795:
Thanks, @outoftime! :smiley: 

Either an external or internal module sounds good to me, although I wonder what would the module look like? Specifically, what exactly would the module be abstracting out from [`custom.js`](https://github.com/tenyoung795/popcode/blob/b0439b7594d30e71b0def3aba3de0933553f76a7/src/validations/html/custom.js)?

Technically, `CustomValidator` itself performs no parsing but indeed registers various parse-event listeners in various ways. Here is the best API I can think of currently, in pseudocode:

```
interface Rule {
  openTag({name, attributes, selfClosing}, location);
  closeTag(name, location);
  done() => Iterable<Error>;
}

function runRules(source) {
  const parser = ...;
  const rules = ...;
  return new Promise((resolve) => {
    register each rule on parser;
    parser.write(source);
    parser.end(() => {
      resolve(chain all rules' errors);
    });
  });
}
```

Personally, I'd prefer the internal module unless we could very confidently devise an externally useful API; roughly speaking, it's easier to recover from under-design than over-design. If this proves externally useful later on, then it could easily be moved out. Hiding a module back into Popcode would be tricky once an external client exists.

outoftime:
Hey @tenyoung795, I’m convinced by your argument about the lower costs of moving to an external module later rather than moving in the opposite direction. I think the interface you propose above is a good one—specifically I would expect the `validator` to just call the `runRules(source)` function and get back a list of errors (synchronously or asynchronously) and then figure out how to turn the errors into the Popcode format. Seem reasonable?

tenyoung795:
Yep, that sounds reasonable! I'll try to redo the branch and then ping this PR once ready.

(If this takes too long, anyone should feel free to announce their own version of this PR.)

src/validations/html/rules/MismatchedTag.js
```diff
@@ -0,0 +1,139 @@
...
+    // Process the remaining errors.
+    for (const [location, {name, mismatchedTags}]
+         of this._mismatchedTagErrorsByLocation) {
+      if (this._unopenedTagLocations.has(location)) {
+        yield {
+          code: Code.UNOPENED_TAG,
+          closeTag: {
+            name,
+            location: {
+              row: location.get('row'),
```
tenyoung795: According to the `Record` [documentation](https://facebook.github.io/immutable-js/docs/#/Record), `location.row` and `location.column` wouldn't work on IE 8, so this (attempts to) stick to Immutable's API subset that works on IE 8. If Popcode shall not support IE 8, then this could use more idiomatic property access instead.

tenyoung795:
Okay, I think it's ready now!

FYI, I think `runRules` and friends could be used to fix #378. For example, a rule could check the descendant tag names of the `head` element to ensure no content was inside the `head`.

src/validations/html/custom.js
```diff
@@ -0,0 +1,144 @@
+import {List, Map, Record, Set} from 'immutable';
+import {SAXParser} from 'parse5';
+import voidElements from 'void-elements';
+import Validator from '../Validator';
+
+// An immutable.Record wrapper of parse5.MarkupData.Location, whose line and
+// col are 1-indexed.
+const Location = Record({line: 1, col: 1});  // eslint-disable-line new-cap
+
+function findMismatchedCloseTags(parser) {
```
outoftime: At a high level, I’d propose restructuring the way you store the state during the tree walk. In particular I would hold on to three collections, each of which corresponds to one of the errors this module can generate:

1. Map of opening tags that don’t have a properly-placed closing tag, and for which we haven’t yet seen a misplaced closing tag, keyed by tag name, with a queue of locations (`unclosed-tag-by-mismatched-close-tag`)
2. Set of pairs of opening/closing tags where the closing tag is in the wrong place (`misplaced-close-tag`)
3. Set of closing tags that don’t match any opening tag (`unexpected-close-tag`)

Any time we encounter a mismatch:

* Add all unclosed tags to collection 1
* If the closing tag we have encountered does not match an opening tag in the stack (`!open`):
  * If the closing tag we have encountered matches an entry in collection 1, remove that entry from collection 1, and add the open/close pair to collection 2
  * Otherwise, add the closing tag to collection 3

Obviously this structure has a lot in common with what you’ve got now, but I think it’ll be easier to reason about. In particular, once the initial tree walk is finished, we have a final accounting of the three error types we can generate—there is no need for a second pass to turn raw observations into specific errors. As a corollary, understanding the entire process requires understanding only the types of errors the module emits; there is no need to additionally understand the structure and intent of a set of intermediate data structures (something I found fairly challenging when taking this code in just now).

Definitely possible I am missing some nuance here that my proposal doesn’t account for, let me know what you think!

src/validations/html/custom.js
```diff
@@ -0,0 +1,144 @@
...
+              close: name,
+              misplacement: spareUnopenedTag.location.line,
+            },
+            suppresses: ['unexpected-close-tag'],
+          };
+        } else if (opened) {
+          yield {
+            row: location.get('line') - 1,
+            column: location.get('col') - 1,
+            reason: 'unclosed-tag-by-mismatched-close-tag',
```
outoftime: Is there a reason not to just report this as a normal `unclosed-tag`? Is it really useful to include information about where we found a mismatched closing tag? Would it make more sense to still locate the error at the location of the opening tag? (These are all sincere/non-rhetorical questions!)

src/validations/html/custom.js
```diff
@@ -0,0 +1,144 @@
...
+      for (const mismatchedTag of mismatchedTags) {
+        const spareUnopenedTag =
+          unopenedTagQueuesByName.get(mismatchedTag.name, List.of()).first();
+        if (spareUnopenedTag) {
+          unopenedTagQueuesByName = unopenedTagQueuesByName.update(
+            spareUnopenedTag.name, tags => tags.shift());
+
+          yield {
+            row: location.get('line') - 1,
+            column: location.get('col') - 1,
```
outoftime: I would make a slightly stronger claim that the location of this error should be the location of the `spareUnopenedTag` rather than the location of the mismatched closing tag, e.g.:

```html
<!DOCTYPE html>
<html>
    <head>
        <title>Page Title</title>
    </head>
    <body>
        <div>
            <p>
        </div>
        </p> <!-- fundamentally feels like the error is on this line -->
    </body>
</html>
```

I would change the message to e.g. `This closing </p> tag should go before the </div> on line 9`

outoftime:
@tenyoung795 oh man i am just wrapping up a monster detail review of the previous state of the code… let me submit it and see what happens

src/validations/html/rules/MismatchedTag.js
```diff
@@ -0,0 +1,139 @@
+import {List, Map, Record, Set} from 'immutable';
```
outoftime: What’s the thinking behind using `Immutable` in here rather than plain JavaScript `Object`, `Array`, `Map`, `Set`? I like `Immutable` and obviously it’s deeply baked into the Redux side of the app, but it’s not without its cost (a more verbose API, having to remember to reassign the variable every time you want to modify the collection) and it doesn’t seem like we’re gaining a lot (reassigning immutable variables vs. mutating mutable constants).

src/validations/html/rules/MismatchedTag.js
```diff
@@ -0,0 +1,139 @@
...
+import Code from '../rules/Code';
+
+const Location = Record({row: 0, column: 0});  // eslint-disable-line new-cap
+
+export default class {
```
outoftime: At a high level, I’d propose restructuring the way you store the state during the tree walk. In particular I would hold on to three collections, each of which corresponds to one of the errors this module can generate:

1. Map of opening tags that don’t have a properly-placed closing tag, and for which we haven’t yet seen a misplaced closing tag, keyed by tag name, with a queue of locations (`unclosed-tag-by-mismatched-close-tag`)
2. Set of pairs of opening/closing tags where the closing tag is in the wrong place (`misplaced-close-tag`)
3. Set of closing tags that don’t match any opening tag (`unexpected-close-tag`)

Any time we encounter a mismatch:

* Add all unclosed tags to collection 1
* If the closing tag we have encountered does not match an opening tag in the stack (`!open`):
  * If the closing tag we have encountered matches an entry in collection 1, remove that entry from collection 1, and add the open/close pair to collection 2
  * Otherwise, add the closing tag to collection 3

Obviously this structure has a lot in common with what you’ve got now, but I think it’ll be easier to reason about. In particular, once the initial tree walk is finished, we have a final accounting of the three error types we can generate—there is no need for a second pass to turn raw observations into specific errors. As a corollary, understanding the entire process requires understanding only the types of errors the module emits; there is no need to additionally understand the structure and intent of a set of intermediate data structures (something I found fairly challenging when taking this code in just now).

Definitely possible I am missing some nuance here that my proposal doesn’t account for, let me know what you think!

src/validations/html/runRules.js
```diff
@@ -0,0 +1,40 @@
...
+// * optionally,
+//   openTag({row: number, column: number}, {name: string})
+//   to take in open tags that are neither void nor self-closing
+// * optionally,
+//   closeTag({row: number, column: number}, name: string)
+//   to take in close tags
+export default (rules, source) => {
+  const parser = new SAXParser({locationInfo: true});
+  return new Promise((resolve) => {
+    parser.on('startTag', (name, attrs, selfClosing, {line, col}) => {
```
outoftime: `selfClosing` as in `<br />`? That’s no longer a thing in HTML5 (self-closing void elements are valid but the self-closing doesn’t actually do anything; self-closing non-void elements are not valid and are ignored, e.g. `<div />` does not give you an empty div element)

src/validations/html/rules/MismatchedTag.js
```diff
@@ -0,0 +1,139 @@
...
+  openTag(location, {name}) {
+    this._openTagStack = this._openTagStack.push({location, name});
+  }
+
+  closeTag(location, name) {
+    const locationKey = new Location(location);
+
+    const openIndex = this._openTagStack.findLastIndex(
+      openTag => openTag.name === name);
+    const opened = openIndex >= 0;
```
outoftime: I generally prefer e.g. `isOpened` for boolean variable names

src/validations/html/rules/MismatchedTag.js
```diff
@@ -0,0 +1,139 @@
...
+
+    // Assume any open tags after the matching open tag are mismatched. If the
+    // source is valid HTML, then no other open tags should exist. If no
+    // matching open tag exists, then all open tags are mismatched.
```
outoftime: Generally I’m against inline comments, as they tend to get out of sync with the actual code and I’m of the opinion that code can and should be written so that its intent is clear without additional comment. In this case, I think it’s quite clear what the next few lines of code are doing!

src/validations/html/rules/MismatchedTag.js
```diff
@@ -0,0 +1,139 @@
...
+  //
+  // so yield UNCLOSED_TAG.
+  //
+  // A mismatched close tag that has yet to be opened means the case
+  //
+  // <div>
+  //   </span>
+  // </div>
+  //
+  // so yield UNOPENED_TAG.
```
outoftime: This comment is useful but I think it should be module-level documentation rather than an inline comment.

src/validations/html/rules/MismatchedTag.js
```diff
@@ -0,0 +1,139 @@
...
+  //
+  // Note: MISPLACED_CLOSE_TAG suggestions invalidate UNCLOSED_TAG and
+  // UNOPENED_TAG suggestions by moving other mismatched close tags.
+  // Diagnosing UNCLOSED_TAG and UNOPENED_TAG accurately would require another
+  // pass after all the MISPLACED_CLOSE_TAG suggestions are resolved.
```
outoftime: I *think* my proposed structure above would remove the need for this invalidation step.

src/validations/html/rules/MismatchedTag.js
outoftime: @tenyoung795 Popcode doesn’t support _any_ version of IE (because IE doesn’t provide a way to make sandboxed client-side iframes), and even if it did, it sure as heck wouldn’t support IE8 ; )

The source of truth for “what browsers do we support” lives in [`browsers.json`](https://github.com/popcodeorg/popcode/blob/master/config/browsers.json)

src/validations/html/rules/MismatchedTag.js
```diff
@@ -0,0 +1,139 @@
...
+          this._foundMisplacedCloseTag = true;
+          this._unopenedTagQueuesByName = this._unopenedTagQueuesByName.update(
+            spareUnopenedTag.name, tags => tags.shift());
+
+          yield {
+            code: Code.MISPLACED_CLOSE_TAG,
+            openTag: mismatchedTag,
+            closeTag: {
+              name,
+              location: {
```
outoftime: I would make a slightly stronger claim that the location of this error should be the location of the `spareUnopenedTag` rather than the location of the mismatched closing tag, e.g.:

```html
<!DOCTYPE html>
<html>
    <head>
        <title>Page Title</title>
    </head>
    <body>
        <div>
            <p>
        </div>
        </p> <!-- fundamentally feels like the error is on this line -->
    </body>
</html>
```

I would change the message to e.g. `This closing </p> tag should go before the </div> on line 9`

src/validations/html/rules/index.js
```diff
@@ -0,0 +1,50 @@
...
+  [Code.MISPLACED_CLOSE_TAG]: ({openTag, closeTag, match}) => ({
+    reason: 'misplaced-close-tag',
+    payload: {
+      open: openTag.name,
+      close: closeTag.name,
+      misplacement: match.row + 1,
+    },
+  }),
+  [Code.UNCLOSED_TAG]: ({openTag, closeTag}) => ({
+    reason: 'unclosed-tag-by-mismatched-close-tag',
```
outoftime: Is there a reason not to just report this as a normal `unclosed-tag`? Is it really useful to include information about where we found a mismatched closing tag? Would it make more sense to still locate the error at the location of the opening tag? (These are all sincere/non-rhetorical questions!)

src/validations/html/rules/MismatchedTag.js
tenyoung795: This module uses containers keyed by locations. (AFAIK) the native JavaScript `Map` and `Set` cannot support custom equality for user-defined types. Their keys may be JavaScript objects but the native containers compare them by identity and (AFAIK) do not support custom equality.

So this uses `Map` and `Set` keyed by the `Record` type `Location`. This could've converted the `{row, col}` to string keys but I figured using a more direct `Record` type would be cleaner and more strongly-typed.

src/validations/html/rules/MismatchedTag.js
tenyoung795: Done.

src/validations/html/rules/MismatchedTag.js
tenyoung795: Okay, it's gone now!

src/validations/html/rules/MismatchedTag.js
tenyoung795: Okay; moved it to the top of the file.

tenyoung795:
I meant to address as much feedback as I could but there's a few spots I unintentionally missed (also a few I intentionally missed but need to explain); I'll hopefully get to all that tomorrow so you can withhold review till then!

src/validations/html/rules/index.js
tenyoung795: > Is there a reason not to just report this as a normal `unclosed-tag`?

Not really other than this was based on `mismatched-close-tag` but:

> Is it really useful to include information about where we found a mismatched closing tag?

Hmm, not really, since the error location already shows where the mismatch was found.

> Would it make more sense to still locate the error at the location of the opening tag?

Actually, the `RuleValidator` (and previous versions) always used the location of the closing tag as the error location. But I agree that the closing tag location + the existing `unclosed-tag` message should suffice.

Too much information in an error can easily overload students!

src/validations/html/rules/MismatchedTag.js
tenyoung795: Okay, I switched to direct property access.

Tangential: has ScriptEd ever encountered schools that only supported IE? Or were they forced to install to a supported browser?

src/validations/html/rules/MismatchedTag.js
tenyoung795: Okay, I reworded `misplaced-close-tag` to the template you described. It originally said the same information but I like your wording better. :smile:

src/validations/html/runRules.js
tenyoung795: On `'startTag`, `SAXParser`'s [event listener](http://inikulin.github.io/parse5/classes/saxparser.html#on) takes in a positional `selfClosing` parameter.

Originally, I figured that it would be easier for the `MismatchedTag` rule to throw out whatever self-closing elements (valid or not) that occur. On the other hand, I don't think students would know about self-closing elements (or at least volunteers should probably avoid bringing them up) so for non-void elements, would it be better to assume that the element was intended to be a normal opening tag?

src/validations/html/rules/MismatchedTag.js
tenyoung795: However, I converted `this._openTagStack` from `List` to a native `Array` since `Array` performs naturally well for a stack, the `List` appears to have no advantages.

Meanwhile, `this._unopenedTagQueuesByName` represent queues, and `List` [promises](https://facebook.github.io/immutable-js/docs/#/List) "efficient addition and removal from both the end (`push`, `pop`) and beginning (`unshift`, `shift`)." Therefore, I've kept `List` for the queues.

src/validations/html/rules/MismatchedTag.js
tenyoung795: Unfortunately, I think the problem is that the misplaced close tag may appear _before_ its matching open tag.

In other words:

> Any time we encounter a mismatch:
> - If the closing tag we have encountered does not match an opening tag in the stack

...but the close tag may match a _later_ open tag in the stream of tags.

I'm not opposed to three collections that each correspond to one of the three errors, but how could they be used while avoiding this issue?

While drafting this PR, I initially tried a similar approach until I realized this issue, hence the more complex containers.

tenyoung795:
Okay, should be ready for review now!

src/validations/html/runRules.js
outoftime: @tenyoung795 I’d say so—that’s how the HTML5 parser treats them, so I think we should too.

src/validations/html/rules/MismatchedTag.js
outoftime: Hey @tenyoung795 so are we talking about something like the below?

```html
<body>
This is a paragraph.</p>

<p>Here is another one
</body>
```

In that case I think it’s probably better behavior to just treat this as two errors: first, an unmatched closing tag, then later, an unclosed opening tag. Try the above markup in this branch and see if you agree that the current error is a bit mystifying.

Obviously if a student generates this situation by cut-and-pasting they may have indeed intended to have the closing tag close the opening tag, but in that case I think it’s still better to show the two errors to get the students thinking more directly about what the correct HTML structure is.

src/validations/html/rules/MismatchedTag.js
outoftime: @tenyoung795 OK, that’s a good point about custom equality. As far as I can tell you‘re correct that the native data structures can only work on exact object/primitive value equality.

That said, a couple of further thoughts on this:

* The fact that `Immutable` collections do the right thing here is fairly tangential to the core competency of `Immutable`, i.e. being immutable! And my previous observation stands that actually dealing with immutability in this module does make some of the code harder to follow. How about using e.g. [collections.js](https://github.com/montagejs/collections), which provides normal mutable collections but does have the ability to define custom equality for keys?
* If we don’t want to introduce a new library to solve this problem, I think we should be consistent about the type of collections we use—meaning I would advocate changing `_openTagStack` back to an `Immutable.List` to be consistent with everything else.

src/validations/html/rules/MismatchedTag.js
outoftime: Looks great!

src/validations/html/rules/MismatchedTag.js
outoftime: (Also I’m pretty skeptical that `Immutable.List` would provide any tangible performance benefits over `Array` for the queue use case here, although I’m certainly down to be proven wrong with a jsperf etc.)

src/validations/html/runRules.js
outoftime: That said, putting e.g. `<div />` in the branch right now does correctly generate an error saying that doesn’t work, so maybe 🤷‍♂️

src/validations/html/rules/MismatchedTag.js
tenyoung795: > in that case I think it’s still better to show the two errors to get the students thinking more directly about what the correct HTML structure is.

Actually @outoftime, are two errors better even if the close tag proceeds the open tag?

```html
    <body>
        This is a paragraph.<p>

        Here is another one
    </body>
    </p>
```

image

Would these two errors be enough to let the average student realize that they misplaced the close tag? e.g. A `<p>` tag needs to be closed by a `</p>` tag, and I have a `</p>` that doesn't match any `<p>` tag; so I need to move that unmatched `</p>` to close the unclosed `<p>` tag.

If so, I'd be happy to get rid of the `misplaced-close-tag` handling, which would significantly reduce complexity.

outoftime: @tenyoung795 I definitely think the `misplaced-close-tag` error is an indispensable part of the functionality in this PR, but I think my proposed structure handles the example you give here, specifically:

> If the closing tag we have encountered matches an entry in collection 1, remove that entry from collection 1, and add the open/close pair to collection 2

Am I thinking about it wrong?

tenyoung795: Provided we only find misplaced closing tags after an open tag, I got a very similar algorithm to work:

- If the closing tag we have encountered matches an opening tag in the stack, add all unclosed tags to collection 1.
- Otherwise:
  - If the closing tag we have encountered matches an entry in collection 1, remove that entry from collection 1, and add the open/close pair to collection 2
  - Otherwise, add the closing tag to collection 3

I think adding all mismatches to collection 1, regardless of whether the mismatch already qualifies for either collection 2 or collection 3, leads to spurious unclosed tag errors since collection 2 or 3 would already "explain" that unclosed tag.

tenyoung795:
Okay @outoftime, after almost a month (I went into a bit of a vacation) it's ready again!

outoftime:
@tenyoung795 awesome! releasing now!

tenyoung795:
Yay, thanks! :smiley:

PR #883, opened 2017-06-23
Set experimental property in ui if url parameter passed

alexpelan:


alexpelan:
Travis is gonna fail because of the cyclomatic complexity of the ui reducer, let me know if you have any feedback

src/actions/applicationLoaded.js
```diff
@@ -2,5 +2,5 @@ import {createAction} from 'redux-actions';
 
 export default createAction(
   'APPLICATION_LOADED',
-  (gistId = null) => ({gistId}),
+  (gistId = null, isExperimental = false) => ({gistId, isExperimental}),
```
outoftime: IMO the action creator should probably just take a named args object at this point (probably should have even before this change)

alexpelan:
@outoftime is that what you meant? Also bumped up complexity to 30

outoftime:
Yep! @alexpelan gonna merge now but would you mind writing up a brief PR description (what the motivation for this is etc.) at your leisure?

PR #884, opened 2017-06-25
Reset button

ibrand:
This PR is the cleaned up version of this branch: https://github.com/popcodeorg/popcode/pull/862 and is a pairing collaboration between @kaylee42 and I.

The feature adds a refresh button to the preview pane that resets the state of the code to before the js has run. It looks like this:
refreshbutton

Often while in the classroom, I saw students either confused as to how to play an animation over that was js based. They would either open the popout window or applying a roundabout way like adding extra spaces to their code and then deleting them to get popcode to re-render.

This feature is meant to add that functionality without having the students editing their code. It uses a timestamp of the last time the refresh button was clicked and inserts that into a comment in the preview pane html to force React to re-render.

Closes #862

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -501,20 +501,28 @@ body {
   flex: 1 0 100%;
 }
 
-.preview__pop-out-button {
-  background: url(../images/popout.png);
+.preview__button {
+  font-family: 'FontAwesome';
   width: 16px;
   height: 18px;
   position: absolute;
-  right: 10px;
-  top: 10px;
   cursor: pointer;
   opacity: 0.25;
   z-index: 1;
 }
 
-.preview__pop-out-button:hover {
+.preview__button:hover {
   opacity: 0.5;
+}
+
+.preview__button_pop-out {
+  right: 10px;
+  top: 11px;
+}
+
+.preview__button_reset {
+  right: 35px;
+  top: 10px;
 }
 
 /** @define error-list */
--- a/.eslintrc
+++ b/.eslintrc
@@ -50,7 +50,7 @@
         ],
         "complexity": [
           2,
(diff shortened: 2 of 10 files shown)


src/css/application.css
```diff
@@ -501,32 +501,28 @@ body {
...
   height: 18px;
   position: absolute;
-  right: 10px;
-  top: 11px;
   cursor: pointer;
   opacity: 0.25;
   z-index: 1;
 }
 
+.preview__pop-out-button {
```
outoftime: So I think it makes sense for this and `preview__reset-button` to be modifiers for `preview__button`—e.g. this would be `preview__button_reset` and the other would be `preview__button_pop-out`—WDYT?

ibrand:
@outoftime This should be ready! Once you've reviewed it, my plan is to squash all these commits into one duet commit that says:

Add a reset button to the popcode preview pane to allow students to reset the state of their code.

How's that sound?

src/css/application.css
ibrand: Yes! That looks great

ibrand:
This PR is the cleaned up version of this branch: https://github.com/popcodeorg/popcode/pull/862 and is a pairing collaboration between @kaylee42 and I.

The feature adds a refresh button to the preview pane that resets the state of the code to before the js has run. It looks like this:
refreshbutton

Often while in the classroom, I saw students either confused as to how to play an animation over that was js based. They would either open the popout window or applying a roundabout way like adding extra spaces to their code and then deleting them to get popcode to re-render.

This feature is meant to add that functionality without having the students editing their code. It uses a timestamp of the last time the refresh button was clicked and inserts that into a comment in the preview pane html to force React to re-render.

PR #886, opened 2017-06-26
Created buttons in Editors to Toggle the text size 

pwjablonski:
Had Some Extra Time this weekend and some time off today so figured I'd give this a shot!

I remember there being a request for this at Hack Day and we talked about potentially including something for the redesign. The idea is to make it easier for instructors to quickly increase font size while presenting/increasing accessibility. 

A couple of points
1) The complexity is too high for the UI function
2) Is this the best place to put the buttons/button?
   Pros - You are able to toggle editors individually / You don't have to open the dashboard to toggle  
   Cons - Extra buttons take up space on the editor pane
   Alternatives - Put a single button on the sidebar that toggles every editor.
  (This will probably all change with the redesign anyway)

3) Is this something we would want to save the state of in an export (ala libraries / hiddencomponents). I'm not sure but seems easy enough to switch.


src/components/Editor.jsx
```diff
@@ -114,6 +116,14 @@ class Editor extends React.Component {
     }
   }
 
+  _toggleEditorTextSize(editorTextSizeIsLarge) {
+    if (editorTextSizeIsLarge === true) {
```
outoftime: You don’t need the `=== true` here

src/components/Editor.jsx
```diff
@@ -114,6 +116,14 @@ class Editor extends React.Component {
...
+      this._editor.setFontSize('18px');
```
outoftime: Probably best to extract these into `const`s at the top of the file (e.g. `normalFontSize` and `largeFontSize`)

src/components/Editor.jsx
```diff
@@ -142,6 +152,7 @@ class Editor extends React.Component {
 }
 
 Editor.propTypes = {
+  editorTextSizeIsLarge: PropTypes.bool.isRequired,
```
outoftime: Since this is the `Editor` component, we can probably just call this prop `textSizeIsLarge`

src/components/EditorContainer.jsx
```diff
@@ -28,6 +28,12 @@ function EditorContainer(props) {
       >
         {t(`languages.${props.language}`)}
       </div>
+      <div
+        className="environment__label_toggleTextSize label"
+        onClick={props.onToggleEditorTextSize}
+      >
+        Aa
```
outoftime: It’d probably be slick to use [fa-search-plus](http://fontawesome.io/icon/search-plus/) and [fa-search-minus](http://fontawesome.io/icon/search-minus/) respectively (we already have FontAwesome installed)

src/components/EditorsColumn.jsx
```diff
@@ -76,8 +77,12 @@ export default class EditorsColumn extends React.Component {
           style={{flex: editorsFlex[index]}}
           onHide={partial(onComponentHide, `editor.${language}`)}
           onRef={partial(this._storeEditorRef, index)}
+          onToggleEditorTextSize={partial(onToggleEditorTextSize,
+           `editor.${language}`)}
```
outoftime: From a product standpoint I think it would make the most sense to have a single size control for all the editors—I can’t think of a situation in which you’d want different editors at different sizes. WDYT?

src/components/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
outoftime: I don’t think this belongs in the diff?

outoftime:
Hey @pwjablonski just a heads up that https://github.com/popcodeorg/popcode/pull/873 is likely to cause some pretty serious merge conflicts—I’m offering free merge conflict fixing for this PR so just lmk if you’d like me to do that : )

outoftime:
@pwjablonski actually! i realized that the PR I just linked doesn’t overlap with what you’ve done here. carry on.

pwjablonski:
Hey! 

1) So I made the changes on this branch based on your suggestions. See below
screen shot 2017-07-08 at 11 14 51 am

I also created two other branches with the button in the sidebar that I can create pull requests for. (Not sure if this is the best way to collaborate but wanted to preserve them in separate branches to compare)

2) Toggles the text size with one button in the sidebar. The icon switches from a plus magnifying glass to a minus when clicked and vice-versa. All of the editors change size. Normal font size is 14px and Large is 20px
screen shot 2017-07-08 at 11 18 49 am
screen shot 2017-07-08 at 11 18 39 am


3)  Changes the text size with two icons in the sidebar. Minimum  font size is 10px and max is 25px. Increments by 1px. 
screen shot 2017-07-08 at 11 09 50 am


I am split between options 2 and 3. 2 is definitely more simple and unassuming but 3 gives more flexibility. 

A couple of notes and questions?
- I'm not too clear with CSS naming conventions so would love feedback
- Do magnifying lens work or do pure plus/minus buttons looks better
- Is the SideBar the right place to put the buttons? Its the only place that is visible when the dashboard is closed besides in the Editors.
- Are the icons the right size and in the best spot on the Sidebar?

outoftime:
This was done in #913!

PR #887, opened 2017-06-30
Revert "Insert popcode JS and CSS libraries on the top of the head tag"

outoftime:
Reverts popcodeorg/popcode#859

Bootstrap does not work properly after the change in #859. Reverting this PR pending investigation.


outoftime:
@jwang1919 FYI—if you have time to look into this awesome, if not I can

jessewang-arvatosystems:
Yep yep! Can do.

PR #892, opened 2017-07-03
Offline access

### LINKED ISSUE #315: Offline access
Once you’ve loaded Popcode once, you should be able to access it again with no network connection.

If you do have a network connection, though, you should immediately be presented with the latest version.

The [Webpack offline plugin](https://github.com/NekR/offline-plugin) looks like the move here.

outoftime:
Builds on #860, installing the Webpack `offline-plugin` to allow offline access to Popcode.

Pretty simple integration. One thing I don’t love is listing out all the fonts/images directly, but other than parsing the CSS I don’t have a great idea for how to keep that deduplicated.

Closes #860
Fixes #315

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,6 +115,7 @@
     "lodash": "^4.15.0",
     "loop-breaker": "^0.1.0-alpha.7",
     "moment": "^2.14.1",
+    "offline-plugin": "^4.8.1",
     "pify": "^2.3.0",
     "promise-retry": "^1.1.1",
     "prop-types": "^15.5.10",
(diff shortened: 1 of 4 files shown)


outoftime:
@quixotically hey, this is working now! the major hurdle was actually that we were serving compiled assets (including everything built by WebPack) from the `/compiled` path, which prevented the service worker from being able to cache stuff that _wasn’t_ under `/compiled`. I [rearranged things so that compiled assets are served from the root path](https://github.com/popcodeorg/popcode/pull/891) and now everything looks great.

let me know what you think!

PR #893, opened 2017-07-04
Add reselect and use it to extract data from store in containers

outoftime:
Adds the `reselect` library, which enables the creation of memoized selectors. The job of selector is to derive information from the state tree in a form that’s useful to the UI.

A few motivations here:

* `reselect` selectors are memoized, meaning expensive calculations can be done once and retained until the inputs change. This is good for runtime efficiency.
* Selector memoization also means that we can get all the benefits of `Immutable` outside the Redux store without having to actually pass around `Immutable` objects. In particular, as long as `toJS()` calls are happening inside a selector, we have a guarantee of object-level equality of the selectors’ output as long as the data hasn’t changed.
* Finally, selectors give us a first-class place to put code that derives useful information out of the Redux store, replacing the ad-hoc system we use now (e.g. `projectUtils` etc.)


inlinestyle:
@outoftime looks good! I... don't have anything particularly deep to add, just a solid set of changes. 👾

outoftime:
Awesome, thanks for looking over it @inlinestyle !

PR #894, opened 2017-07-04
Use Blendid as basis of build tooling

outoftime:
Blendid is a lightweight client-side app framework that provides a build toolchain for building apps using preferred tools like Webpack, Babel, BrowserSync for local development, etc. This PR ports Popcode to using Blendid instead of a hand-rolled gulpfile for build tooling.


outoftime:
I opened this for archival purposes, but after exploring it, I don’t actually want to do it. The main problem is dependencies: Blendid adds dependencies we don’t actually need (e.g. `node-sass`, which annoyingly rebuilds the native C SASS library basically any time you run `yarn`); also there’s no guarantee things will be up to date (e.g. right now Blendid is on Webpack 2, which we’d be locked into if we used it).

PR #913, opened 2017-07-08
Toggle text size sidebar

pwjablonski:
Toggles the text size with one button in the sidebar. The icon switches from a plus magnifying glass to a minus when clicked and vice-versa. All of the editors change size. Normal font size is 14px and Large is 20px
screen shot 2017-07-08 at 11 18 49 am
screen shot 2017-07-08 at 11 18 39 am


src/components/Editor.jsx
```diff
@@ -114,6 +118,14 @@ class Editor extends React.Component {
     }
   }
 
+  _toggleEditorTextSize(textSizeIsLarge) {
```
outoftime: This method name is a little misleading, as it sounds like it’s something to do with changing the current state of the font size rather than just syncing the editor with the underlying state. I would call it e.g. `_applyFontSize` (also no need for `Editor` as we are already in the `Editor` component)

src/components/Sidebar.jsx
```diff
@@ -35,6 +35,13 @@ class Sidebar extends React.Component {
       },
     );
 
+    let plusMinus;
+    if (this.props.textSizeIsLarge) {
+      plusMinus = <span className="sidebar__plusMinus">&#xf010;</span>;
+    } else {
+      plusMinus = <span className="sidebar__plusMinus">&#xf00e;</span>;
```
outoftime: The only thing that’s varying here is the actual escape sequence of the FontAwesome character, why not just store that in the variable? (Or you could maybe even just inline it in the JSX using a conditional)

src/css/application.css
```diff
@@ -194,6 +194,18 @@ body {
...
+  text-align: center;
+  padding-bottom: 1rem;
+  font-size: 1.5rem;
+}
+
+.sidebar__plusMinus {
+  font-family: 'FontAwesome';
+  position: relative;
+  margin: auto;
+}
```
outoftime: I would factor the CSS thusly:

* Create a `u__icon` utility element that applies the FontAwesome font, as well as a `cursor: pointer` and `user-select: none`. Use it here and also apply it to the other FontAwesome icons in the UI.
* Combine the `sidebar__component` and `sidebar__plusMinus` elements into a single element (keep the `sidebar__plusMinus` name)

Seem reasonable?

src/css/application.css
```diff
@@ -397,6 +409,13 @@ body {
   z-index: 1;
 }
 
+.environment__label_toggleTextSize {
```
outoftime: I think this might be a straggler from the first implementation?

static/compiled/application.css
```diff
@@ -0,0 +1,1171 @@
+/*! normalize.css v4.1.1 | MIT License | github.com/necolas/normalize.css */
```
outoftime: These files shouldn’t be under source control—this used to be where compiled asset files lived, now moved to `dist/`

pwjablonski:
Ahh didn't run tests first! Will update!

static/compiled/application.css
pwjablonski: Not sure why this is happening. How would I go about fixing this?

outoftime: @pwjablonski make sure you’ve merged/rebased in the latest master, then do `git rm -r static`

outoftime:
@pwjablonski I will merge https://github.com/popcodeorg/popcode/pull/952 in a few minutes and then, when you pull from master, all your complexity problems should be solved

src/css/application.css
outoftime: @pwjablonski doesn’t look like this is used anywhere?

pwjablonski: Missed that one. Thanks!

src/components/Sidebar.jsx
```diff
@@ -50,6 +50,16 @@ class Sidebar extends React.Component {
...
           onClick={this.props.onToggleDashboard}
         />
+        <div
+          className="sidebar__plusMinus"
+          onClick={this.props.onToggleEditorTextSize}
+        >
+          {
+            this.props.textSizeIsLarge ?
+              <icon className="u__icon">&#xf010;</icon> :
+              <icon className="u__icon">&#xf00e;</icon>
```
outoftime: @pwjablonski so sorry I missed this before, but… I’m pretty sure [HTML doesn’t have an `<icon>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element)? Typically I think we would use a `<span>` for this.

PR #939, opened 2017-07-16
Create Snapshot

### LINKED ISSUE #879: Add “Snapshot” feature, deprecate gist-based submission flow
### Abstract

Create a new **Create Snapshot** feature, which allows students to generate a link back to their work that they can use to submit assignments. Treat this as the preferred way to submit and share work, in favor of gists.

### Motivation

In order to complete and submit even the most basic assignment with Popcode, a student must use two tools: Popcode and GitHub gists. When a student completes work, they export that work as a gist, and then copy the URL of the gist (or the import-back URL in the gists’s description) and submit it.

Including gists in this process adds extra workflow complexity (gist opens in a separate window, student has to copy the URL from the address bar) and steepens the initial learning curve for Popcode. In order to understand Popcode’s most basic classroom workflow, students now must be introduced to two different tools simultaneously, understand how they are different, etc.

The motivation for using gists in the current workflow was primarily a practical one: at the time I initially developed the export feature, Popcode had not yet integrated Firebase, and had no way to store student work on its own. The choice to use gists was not educationally motivated, and I don’t think there is a lot of educational value in the way gists are used here. While GitHub is undoubtedly an important tool for professional developers, gists are not; the current workflow supported by Popcode does not really intersect with the way a professional developer uses Git and GitHub at all.

So, let’s cut out the extra step and the extra instructional load, and just build a feature that allows students to get a link to a “snapshot” of their current code, which they can then submit for assignments, and which instructors can use to distribute starter code.

### How it would work

1. Student completes assignment (or instructor completes starter code)
2. Student clicks **Create Snapshot** button
3. Message appears saying “Your snapshot is ready! Click here to copy it.”
4. Student clicks the prompt and pastes the snapshot link into assignment submission.
5. When instructor clicks the link, a new Popcode environment opens with the student’s code loaded (exactly how gist link-backs work today)

### Implementation

Creating a snapshot would store a copy of the current project in Firebase, publicly accessible to anyone with the snapshot key (probably a UUID). Accessing Popcode with a `snapshot=` query parameter would instruct Popcode to load the given snapshot out of Firebase, much in the same way it currently does with a `gist=` parameter.

### Advantages

* Simplifies submission flow by removing the extra step of opening the Gist page and copying the URL from there
* Students can perform basic Popcode assignment workflow without needing to be exposed to GitHub
* Snapshot feature works the same way logged out or logged in; there is no need for students to sign up for GitHub accounts in order to do work in the classroom
* No longer need to deal with Gist 404 errors (these are caused by GitHub being slow to update and can’t be directly fixed on Popcode’s side)
* Removing GitHub from the core classroom workflow makes it easier to switch to using Google accounts as the main login provider in the future if we want to (good for Google Classroom integration)
* We’d be able to introduce [full GitHub repo export](https://github.com/popcodeorg/popcode/pull/864) without worrying about the ambiguity of “which of these do students use in day-to-day classroom workflows?” (neither)

### Things that would not change

* Popcode would still support Gist export; it just wouldn’t be the default flow for assignment submission
* Popcode would still support Gist import (i.e. existing starter code links will always work)
* Popcode would still use GitHub as the login provider (though we may want to change this in the future)

outoftime:
Adds a new Create Snapshot feature, which allows students to generate a link back to their work that they can use to submit assignments. This is similar to the preexisting workflow of exporting a gist and then submitting a link to it, but snapshots simplify the workflow and the conceptual background needed (specifically, you no longer need to learn about GitHub in order to work with Popcode assignment submission). See #879 for a longer discussion of motivation.

This PR also makes some significant changes to notifications, in order to support a notification that updates its own state when a button to copy the link is clicked:

* There is now a `NotificationList` container component
* Notifications in the store are stored keyed by notification type. This brings the actual structure in line with various places in the code that implicitly assume that each notification type can occur at most once.
* Notifications now have a `metadata` property that specialized notifications can use to store their own state; its structure is opaque to the broader application.

Fixes #879


jwang1919:
Code looks good but I have a few comments on the functionality:

1. I suggest a more noticeable change after snapshot link is clicked, like changing the text to `Your snapshot has been copied!`. I found the checkmark on click a little too subtle. I got confused when I copied/pasted the link and realized that it was a snapshot that I took earlier.

2. When there is a HTML/CSS/JS error in Popcode and a snapshot is saved, going back to the snapshot URL will not show any errors. Clicking on the refresh button will not fix this. Only way to show the error is to type something in the problem panel.
IE: http://localhost:3000/?snapshot=29c927aa-9f12-4548-95d3-8a3c1a43071a

Issue #779, opened 2017-04-17
Layout Overhaul


outoftime:
Popcode has an assortment of existing or anticipated UI/UX problems which I believe would be best addressed with a high-level refresh of the tool’s layout. Some prior art can be found in [this issue](https://github.com/popcodeorg/popcode/issues/638) and the [original design workups](https://github.com/popcodeorg/popcode-assets/blob/master/Popcode_identity_051616.pdf) (look at the fourth slide).

A layout refresh should make the following UX improvements:

- [x] Popcode’s login functionality is discoverable
- [ ] It’s clear that projects are automatically saved when you’re logged in; it’s also clear and explicit that projects are *not* automatically saved when you’re logged out
- [x] The “Load Project” view is legible (use the title tag; probably show the project list in a modal rather than inline)
- [ ] There’s a persistent visual indicator of which libraries are enabled
- [x] It’s discoverable that you can minimize editors/the output pane

The improved layout should also contemplate and accommodate these future features:

- [x] There is a space for project instructions, separate from the code and output
- [x] It’s possible to [switch between multiple pages](https://github.com/popcodeorg/popcode/issues/786) in the HTML editor
- [x] There is a JavaScript console built in to the tool (N.B. the layout/UI should make this unmistakably distinct from the output/preview pane)

Finally, we want to maintain these current good things about Popcode’s design:

* Avoids unnecessary visual distractions; keeps the focus on the code and the output
* Usable on very small screens (e.g. Chromebooks)
* Fun colors

## Action Item Plan

- [x] Add top bar with logo/wordmark; make sidebar dark gray and dashboard light gray; remove wordmark from sidebar
- [x] Remove logo from dashboard
- [x] Add hamburger menu button (opens dashboard)
- [x] Move login/logout state into top bar
- [x] Move text size toggle into top bar
- [x] Move export button into top bar
- [x] Move libraries button into top bar
- [x] Border between top bar and notifications
- [x] “New Project” button in top bar
- [x] My Projects in top bar (maybe release along with New Project)
- [x] Put editor labels in the top-right, chevron to minimize, full-width bar that can be maximized
- [x] Project preview top bar that displays title, refresh and pop-out buttons
- [x] Move remaining functionality from Dashboard into actual hamburger menu (designs TK); temporarily remove sidebar until project instructions functionality is complete

PR #975, opened 2017-08-12
Move text size toggle to the top bar

outoftime:
Next step in the move everything to the top bar project—moves the text size toggle.

localhost-3001- laptop with hidpi screen 4

Also consolidates the `u__fontawesome` element into the better `u__icon`, and removes a pesky projection.

Another step in #779


outoftime:
@pwjablonski relevant to your interests

PR #985, opened 2017-08-16
Overhaul editor labels and show/hide

outoftime:
Editor labels are moved to the top-right, lighter color scheme, no transparency, and no rounded corners. Also chevrons.

When clicked, the editor is minimized. Minimized editor appears as a bar at the bottom of the editors column, basically a full-width version of the editor label, with no actual editor. The chevron points up. Clicking this bar expands the editor.

This did involve removing the label on the output, which does not exist in the redesign. This means that there is now no way to minimize the output. I think that might be okay since you can now drag the column divider to give the editors more space, although ideally it would be good to still be able to fully minimize it.

localhost-3001- smallest common laptop 8

### DIFF
--- a/src/components/EditorContainer.jsx
+++ b/src/components/EditorContainer.jsx
@@ -23,10 +23,12 @@ function EditorContainer(props) {
       style={props.style}
     >
       <div
-        className="environment__label label"
+        className="environment__label"
         onClick={props.onHide}
       >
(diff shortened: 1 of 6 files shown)


outoftime:
@inlinestyle very cool (and a little unexpected) that dragging to resize the two expanded editors works 100% perfectly when there is a collapsed editor in the flex column!

Issue #780, opened 2017-04-17
Project instructions


outoftime:
We would like for instructors to be able to distribute instructions with project starter code. Currently, this is done by way of inline comments in the starter code, but students find this confusing.

While project instructions will certainly be part of a larger project assignment-submission flow currently planned for spring of 2018, we’d like to do something simpler sooner.

Initial pitch would be:

* After exporting starter code to a gist, instructors can add a new file with a certain name e.g. `instructions.md`, with project instructions in Markdown
* When Popcode imports a gist, it’ll look for this entry and, if present, display the instructions in their own pane (probably above/below the output?)
* Instructions should be saved with the project, so that loading the project again later will retain the instructions
* When exporting a gist, the instructions should be included (mostly to make it easy for instructors to import a sample project, make some modifications, and then re-export for their class).

Not having the ability to actually write instructions in Popcode is obviously not amazing, but I’d like to see if it’s enough to tide us over until we build the full assignment workflow. If not, we can always follow up with instruction creation/editing functionality.

inlinestyle:
@outoftime regarding the name of the `.md` file - 
It looks like project import [as implemented](https://github.com/popcodeorg/popcode/blob/master/src/reducers/projects.js#L42) concatenates all available `.css` and `.js` files. Would it make sense to do the same with `.md` files, i.e. find, concatenate, and display them all?

outoftime:
@inlinestyle sure, that seems like a more satisfying approach than having a specific magic filename hard-coded in

outoftime:
Closed by #988

outoftime:
_Actually_ closed by #1033 #1000

PR #988, opened 2017-08-17
Project README

inlinestyle:
Implements https://github.com/popcodeorg/popcode/issues/780!

Notes:
- I went with the [remark](http://remark.js.org/) markdown->react library because it looks like it's being used in https://github.com/popcodeorg/popcode/pull/861.
- Adds markdown file import&export to the gist and snapshot features, but does not add it to the proposed repo export feature (this should be trivial though).
- Makes order of all file concatenation deterministic (including for CSS & JS, this seemed like a good idea).

Some sample gists if you want to check them out on the branch:

[Gist with small markdown sample](https://gist.github.com/inlinestyle/f72628405e55bea43267ed91f0bb92dc)
image

[Gist with long markdown sample](https://gist.github.com/inlinestyle/b2b6a8fa00f160e00ccf9ca0c1895994)
image

### DIFF
--- a/src/containers/Dashboard.js
+++ b/src/containers/Dashboard.js
@@ -13,12 +13,13 @@ import {
 } from '../selectors';
 
 function mapStateToProps(state) {
+  const project = getCurrentProject(state);
   return {
-    currentProject: getCurrentProject(state),
     currentUser: getCurrentUser(state),
     gistExportInProgress: isGistExportInProgress(state),
     isExperimental: isExperimental(state),
     isOpen: isDashboardOpen(state),
+    readme: project ? project.readme : '',
   };
 }
 
--- a/src/containers/Dashboard.js
+++ b/src/containers/Dashboard.js
@@ -17,9 +17,9 @@ function mapStateToProps(state) {
   return {
     currentUser: getCurrentUser(state),
     gistExportInProgress: isGistExportInProgress(state),
+    instructions: project ? project.instructions : '',
     isExperimental: isExperimental(state),
     isOpen: isDashboardOpen(state),
-    readme: project ? project.readme : '',
   };
 }
 
--- a/src/containers/Dashboard.js
+++ b/src/containers/Dashboard.js
@@ -5,7 +5,7 @@ import {
   exportRepo,
 } from '../actions';
 import {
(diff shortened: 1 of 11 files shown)


src/containers/Dashboard.js
```diff
@@ -13,12 +13,13 @@ import {
 } from '../selectors';
 
 function mapStateToProps(state) {
+  const project = getCurrentProject(state);
```
outoftime: I think it’s worth making an explicit `getCurrentProjectReadme` here to minimize container knowledge of the store structure

src/reducers/projects.js
```diff
@@ -28,6 +29,13 @@ function unhideComponent(state, projectKey, component, timestamp) {
   ).setIn([projectKey, 'updatedAt'], timestamp);
 }
 
+function contentForLanguage(files, language) {
+  const filesForLanguage = sortBy(
+    filter(files, {language}),
+    file => file.filename);
```
outoftime: Prefer the `);` on a newline here : )

src/records/Project.js
```diff
@@ -14,20 +14,23 @@ export default class Project extends Record({
   enabledLibraries: new Set(),
   hiddenUIComponents: new Set(),
   updatedAt: null,
+  readme: '',
```
outoftime: I don’t feel strongly but I think `instructions` might be a more straightforward description of the intent of this?

outoftime: (I’m definitely down to call it `README.md` in the gist/repo exports though)

outoftime:
@inlinestyle actually I’m cooling on the idea of “automatically opening the instructions pane”—I don’t think that’s the right mental model. Instead I think basically the default state should be that the instructions pane is visible if there are instructions; then the coder has the option of *hiding* the instructions, using the same overall model as hiding other UI components (editors etc.)

outoftime:
which is to say I don’t think we need to worry about it here and instead should worry about it when the `Dashboard` officially ceases to exist and just becomes the instructions area

inlinestyle:
@outoftime updated!
- New `getCurrentProjectInstructions` selector
- `project.readme` -> `project.instructions`
- Newline before closing `);`

PR #991, opened 2017-08-17
Use container div to correctly size preview iframe

### LINKED ISSUE #989: The screen view is terrible
Submitter: Ruhika
Email: rchatterjee188@gmail.com
The screen view is terrible. I can't view my entire screen to do my tests. I had no trouble with this yeasterday.

### LINKED ISSUE #990: alert() display is off
Submitter: alert() display is off
Email: peter@scripted.org
The alert function model does not fill the entire preview screen

screen shot 2017-08-17 at 3 30 33 pm

outoftime:
Fun fact: [you can’t use `top`/`left`/`right`/`bottom` to size an absolutely-positioned iframe!](https://www.w3.org/TR/CSS21/visudet.html#abs-replaced-height).  You need to give iframes explicit `height` and `width`, full stop.

The solution is pretty simple—add a container `<div>` that we size using the absolute position boundaries, then give the iframe a `height` and `width` of 100%.

Fixes #989
Fixes #990

### DIFF
--- a/src/components/PreviewFrame.jsx
+++ b/src/components/PreviewFrame.jsx
@@ -110,11 +110,13 @@ class PreviewFrame extends React.Component {
     }
 
     return (
-      <iframe
-        className="preview__frame"
-        sandbox={sandboxOptions}
-        {...srcProps}
-      />
+      <div className="preview__frame-container">
+        <iframe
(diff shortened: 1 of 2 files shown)


outoftime:
@pwjablonski FYI

Issue #499, opened 2016-10-22
Confirm identity after logging in


outoftime:
Inspired by some of the issues brought up in #216.

Since GitHub login sessions are unavoidably long-lived, we often have situations where a student will open a shared laptop, click “Log In” in Popcode, and be unintentionally logged in as a different student (because that different student has an active GitHub session in the browser). Unfortunately, there is no way to make GitHub confirm the upstream identity as part of the login process (I believe some OAuth providers—maybe Twitter?—do provide this option).

The best workaround I can think of is:
- After initial login, immediately pop up a modal with the avatar and login of the account that just authed, asking “Is this you?”
- At this point, the user session should _not_ yet be considered authenticated (in particular, if there is a current project, it should not be consumed into the authed account).
- If the student says yes, complete the login process.
- If the student says no, the modal switches to a set of instructions that directs the student to https://github.com/logout (maybe in a pop-up).
- When the student has logged out of GitHub (maybe we can watch the state of the window we created?), restart the auth process.

outoftime:
No longer needed with Google auth login

PR #999, opened 2017-08-20
Identity confirmation

ericliu121187:
#### Motivation

Here is a first pass at #499 . If a Github user was logged in previously, there is currently no check for the current user before authenticating the previously logged-in user.

Hopefully, by prompting the user to confirm if the previously logged-in user is the correct identity, it is one step to mitigating incorrect authentications.

#### Approach

Instead of authenticating the logged-in Github user, I first create an `unconfirmedIdentity` in the `state.user`. This seems a bit icky.

Then, `IdentityConfirmation` would render conditionally based on that which is a modal for the user to confirm identity.

If confirmed, an action would be dispatched then update `state.user`, which would resume the flow.

#### Concerns

- I didn't style anything. Is there a particular guide or design I should follow for the modal?
- There was mention of potentially tracking if the user logged out of Github, but I'm not sure how that would be possible. I support the solution of writing out instructions on the modal after the user clicks "no".

#### Screenshots

modal

#### TODO
- [x] Style IdentityConfirmation
- [x] Update IdentityConfirmation with copy (and user flow if needed)
- [ ] Test

### DIFF
--- a/src/components/IdentityConfirmation.jsx
+++ b/src/components/IdentityConfirmation.jsx
@@ -3,13 +3,11 @@ import PropTypes from 'prop-types';
 import AuthenticationStates from '../enums/AuthenticationStates';
 import config from '../config';
 
-function IdentityConfirmation(props) {
-  const {
-    currentUser,
-    onConfirmIdentity,
-    onRejectIdentity,
-  } = props;
-
+function IdentityConfirmation({
+  currentUser,
+  onConfirmIdentity,
+  onRejectIdentity,
+}) {
   if (currentUser.authenticationState !== AuthenticationStates.AUTHENTICATED) {
     return null;
   }
--- a/src/components/IdentityConfirmation.jsx
+++ b/src/components/IdentityConfirmation.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import AuthenticationStates from '../enums/AuthenticationStates';
+import config from '../config';
 
 function IdentityConfirmation(props) {
   const {
@@ -16,14 +17,47 @@ function IdentityConfirmation(props) {
   return (
     <div className="identity-confirmation">
       <div className="identity-confirmation__modal">
-        {`Are you ${currentUser.displayName}?`}
+        <h1 className="identity-confirmation__title">
+          {`Are you ${currentUser.displayName}?`}
+        </h1>
 
-        <button type="button" onClick={onRejectIdentity}>
-          No
-        </button>
+        <p>
+          {`If you are not ${currentUser.displayName}:`}
+        </p>
 
-        <button type="button" onClick={onConfirmIdentity}>
-          Yes
+        <ol>
+          <li>
+            Click the link below to be taken to the GitHub logout page.
+          </li>
+
+          <li>
+            On the GitHub logout page, click the log out button.
+          </li>
+
+          <li>
+            Sign in with your own username and password.
+          </li>
+
+          <li>
+            {'Come back here and click "Log in" again.'}
+          </li>
+        </ol>
+
+        <a
+          href={config.gitHubLogoutUrl}
+          target="_blank"
+          onClick={onRejectIdentity}
+        >
+          {`If you are not ${currentUser.displayName}, click here to  be taken
+          to the GitHub logout page.`}
+        </a>
+
+        <button
+          className="identity-confirmation__confirm-btn"
+          type="button"
+          onClick={onConfirmIdentity}
+        >
+          {`Yes, I am ${currentUser.displayName}. Log me into Popcode!`}
         </button>
       </div>
     </div>
--- a/src/components/IdentityConfirmation.jsx
+++ b/src/components/IdentityConfirmation.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import {t} from 'i18next';
 import AuthenticationStates from '../enums/AuthenticationStates';
 import config from '../config';
 
@@ -18,38 +19,15 @@ function IdentityConfirmation({
     <div className="identity-confirmation">
       <div className="identity-confirmation__modal">
         <h1 className="identity-confirmation__title">
-          {`Are you ${displayName}?`}
+          {t('identity-confirmation.title', {displayName})}
         </h1>
-
-        <p>
-          {`If you are not ${displayName}:`}
-        </p>
-
-        <ol>
-          <li>
-            Click the link below to be taken to the GitHub logout page.
-          </li>
-
-          <li>
-            On the GitHub logout page, click the log out button.
-          </li>
-
-          <li>
-            Sign in with your own username and password.
-          </li>
-
-          <li>
-            {'Come back here and click "Log in" again.'}
-          </li>
-        </ol>
 
         <a
           href={config.gitHubLogoutUrl}
           target="_blank"
           onClick={onRejectIdentity}
         >
-          {`If you are not ${displayName}, click here to  be taken
-          to the GitHub logout page.`}
+          {t('identity-confirmation.github-log-out-prompt', {displayName})}
         </a>
 
         <button
@@ -57,7 +35,7 @@ function IdentityConfirmation({
           type="button"
           onClick={onConfirmIdentity}
         >
-          {`Yes, I am ${displayName}. Log me into Popcode!`}
+          {t('identity-confirmation.confirm-identity', {displayName})}
         </button>
       </div>
     </div>
--- a/src/components/IdentityConfirmation.jsx
+++ b/src/components/IdentityConfirmation.jsx
@@ -1,0 +1,43 @@
+import isNull from 'lodash/isNull';
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function navigateToGithubLogOut() {
+  const GITHUB_LOGOUT_URL = 'https://github.com/logout';
+  window.open(GITHUB_LOGOUT_URL, '_blank');
+}
+
+function IdentityConfirmation({currentUser, onConfirmIdentity}) {
+  const {unconfirmedIdentity} = currentUser;
+
+  function handleConfirmIdentity() {
+    onConfirmIdentity(currentUser.unconfirmedIdentity);
+  }
+
+  if (isNull(unconfirmedIdentity)) {
+    return null;
+  }
+
+  return (
+    <div className="identity-confirmation">
+      <div className="identity-confirmation__modal">
+        {`Are you ${unconfirmedIdentity.displayName}?`}
+
+        <button type="button" onClick={navigateToGithubLogOut}>
+          No
+        </button>
+
+        <button type="button" onClick={handleConfirmIdentity}>
+          Yes
+        </button>
+      </div>
+    </div>
+  );
+}
+
+IdentityConfirmation.propTypes = {
+  currentUser: PropTypes.object.isRequired,
+  onConfirmIdentity: PropTypes.func.isRequired,
+};
+
+export default IdentityConfirmation;
--- a/src/components/IdentityConfirmation.jsx
+++ b/src/components/IdentityConfirmation.jsx
@@ -2,12 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import AuthenticationStates from '../enums/AuthenticationStates';
 
-// TODO: Move to saga
-function navigateToGithubLogOut() {
-  const GITHUB_LOGOUT_URL = 'https://github.com/logout';
-  window.open(GITHUB_LOGOUT_URL, '_blank');
-}
-
 function IdentityConfirmation(props) {
   const {
     currentUser,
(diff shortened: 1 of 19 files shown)


outoftime:
@ericliu121187 this is a great first crack!

Very interesting the way you decided to represent the intermediate state in the store. My initial thought was to basically keep everything the same, but change the `user.authenticated` boolean to essentially an enum, e.g. `user.authenticationState` which could be `unauthenticated`, `unconfirmed`, or `authenticated`. I think this would require changing marginally more existing application code (anywhere that looks at `user.authenticated` basically), but I still lean toward that approach because the overall data structure is less complex (you’re always going to find the user’s name at `user.displayName`, etc.)

Another motivation for that approach is that ideally there would be a _fourth_ `authenticationState`, namely `unknown`, which would be the initial state until Popcode gets feedback from Firebase, and which could be reflected in the UI by not showing either a log in prompt or the user’s badge. (All this would of course belong in a different PR though)

So, what do you think?

Also to answer your questions:

This is the first time we’re putting a modal in the interface, so there’s no prior art on styling. I’d say just do what feels reasonable and try to use existing styles elsewhere in the interface as a guide where you can!

> There was mention of potentially tracking if the user logged out of Github, but I'm not sure how that would be possible. I support the solution of writing out instructions on the modal after the user clicks "no".

Yeah, I think you’re right that it’s not possible, just wanted to make sure we had turned over all rocks. I think spelling out instructions is a good call. Which could be represented by yet a _fifth_ `authenticationState`, something like `rejected` or `denied`!

ericliu121187:
> I think this would require changing marginally more existing application code (anywhere that looks at user.authenticated basically), but I still lean toward that approach because the overall data structure is less complex (you’re always going to find the user’s name at user.displayName, etc.)

That's why I opted for this approach first, but having enums is much cleaner and more robust. Let's do it!

Do you prefer the enums to be hard-coded? I don't think I saw app-wide constants?

outoftime:
@ericliu121187 I’m fine either way—if we just want to use hard-coded strings that works, I’d also be down to introduce something like [enumify](https://github.com/popcodeorg/popcode/pull/999) (it would be useful elsewhere in the application as well)

ericliu121187:
Here's a pass on UI:

identity confirmation modal

#### "Confirm" and "Reject"

We could potentially swap the "reject" and "confirm" styles. This could potentially reduce false positives, the chance that the user accidentally clicks "confirm".

#### UX

Right now, after the user clicks either button, the modal goes away removing the instructions. On a "confirm", we could keep that current flow.

On a "reject", we can keep the instructions and modify copy slightly. However, we could update the actions to "cancel" and "I have signed in. Log me in!" which would take them back to the first step of the modal.

#### Copy and basic styling
Lmk copy changes and any styles you'd like to try out!

ericliu121187:
Do you prefer having the `window.open` and `signOut` (Firebase) onClick of the "reject" being handled in a saga rather than directly on the onClick?

outoftime:
@ericliu121187 awesome! on the UX tip I think the current behavior—closing the window in either case—is fine. I do worry that that list of instructions is a lot of text for students to read (I note students usually just try to get past messages like this as quickly as possible). That said, I don’t have a great suggestion for how to trim it down.

As far as UI goes, I’d be inclined to put the two buttons next to each other, with similar styling, except giving the “log me out” button a gray background so that the happy path still sits at the top of the visual hierarchy.

Gonna do a full pass on line-level code review as well—I feel like we are getting quite close on functionality so this is the time!

src/css/application.css
```diff
@@ -73,6 +73,90 @@ body {
...
+  width: 0;
+}
+
+.identity-confirmation__confirm-btn:active:before {
+  padding-top: 120%;
+  transition: width .2s ease-out, padding-top .2s ease-out;
+  width: 120%;
+}
+
+.identity-confirmation__modal {
```
outoftime: So I think we actually want two separate BEM blocks here—a `modal` which deals with the basic modal layout, and then an `identity-confirmation` which styles the stuff in this particular modal. So it would look something like:

```html
<div class="modal>
  <div class="modal__contents identity-confirmation">
    <h1 class="identity-confirmation__title">Are you Eric Liu?</h1>
    <!-- etc. -->
  </div>
</div>
```

src/enums/AuthenticationStates.js
```diff
@@ -0,0 +1,9 @@
+import {Enum} from 'enumify';
+
+export default class AuthenticationStates extends Enum {}
+AuthenticationStates.initEnum([
+  'AUTHENTICATED',
+  'CONFIRMED',
+  'REJECTED',
```
outoftime: I would propose instead:

* `AUTHENTICATED` (meaning fully authenticated and confirmed)
* `PENDING_CONFIRMATION`
* `UNAUTHENTICATED`

Allows us to undo a bunch of the `authenticated` -> `confirmed` changes in the PR (I think `isUserAuthenticated` is generally more descriptive than `isUserConfirmed`) and keeps the concern of confirmation more local to just the confirmation process

src/reducers/user.js
```diff
@@ -1,19 +1,32 @@
 import Immutable from 'immutable';
 import get from 'lodash/get';
+import AuthenticationStates from '../enums/AuthenticationStates';
 
-const init = new Immutable.Map({authenticated: false});
+const init = new Immutable.Map({
+  authenticationState: AuthenticationStates.UNKNOWN,
+});
 
 function user(stateIn, action) {
```
outoftime: Does this do the right thing if you access Popcode while already logged in (e.g. by opening a second tab)? I suspect the current behavior will be to ask you for confirmation, which isn’t what we want.

src/components/IdentityConfirmation.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+  if (currentUser.authenticationState !== AuthenticationStates.AUTHENTICATED) {
+    return null;
+  }
+
+  return (
+    <div className="identity-confirmation">
+      <div className="identity-confirmation__modal">
+        <h1 className="identity-confirmation__title">
+          {`Are you ${currentUser.displayName}?`}
```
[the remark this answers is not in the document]
ericliu121187: Will do - does all of the copy below also go there as well?

src/css/application.css
```diff
@@ -73,6 +73,90 @@ body {
...
+  color: #ecf0f1;
+
+  transition: background-color .3s;
+}
+
+.identity-confirmation__confirm-btn:hover, .identity-confirmation__confirm-btn:focus {
+  background-color: #27ae60;
+}
+
+.identity-confirmation__confirm-btn {
```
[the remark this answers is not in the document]
ericliu121187: Ah, don't think the linter is catching that - also there's 16 errors with the linter that seem unrelated to this PR. Is that normal?

```
/Users/ericliu/Projects/popcode/src/components/EditorContainer.jsx
  1:23  error  Missing file extension for "inline-style-prefixer/static"        import/extensions
  1:23  error  Unable to resolve path to module 'inline-style-prefixer/static'  import/no-unresolved

/Users/ericliu/Projects/popcode/src/components/EditorsColumn.jsx
  3:23  error  Missing file extension for "inline-style-prefixer/static"        import/extensions
  3:23  error  Unable to resolve path to module 'inline-style-prefixer/static'  import/no-unresolved

/Users/ericliu/Projects/popcode/src/components/Output.jsx
  3:23  error  Missing file extension for "inline-style-prefixer/static"        import/extensions
  3:23  error  Unable to resolve path to module 'inline-style-prefixer/static'  import/no-unresolved

/Users/ericliu/Projects/popcode/src/util/markdown.js
  4:28  error  Missing file extension for "remark-react-lowlight"                        import/extensions
  4:28  error  Unable to resolve path to module 'remark-react-lowlight'                  import/no-unresolved
  5:17  error  Missing file extension for "highlight.js/lib/languages/css"               import/extensions
  5:17  error  Unable to resolve path to module 'highlight.js/lib/languages/css'         import/no-unresolved
  6:17  error  Missing file extension for "highlight.js/lib/languages/xml"               import/extensions
  6:17  error  Unable to resolve path to module 'highlight.js/lib/languages/xml'         import/no-unresolved
  7:16  error  Missing file extension for "highlight.js/lib/languages/javascript"        import/extensions
  7:16  error  Unable to resolve path to module 'highlight.js/lib/languages/javascript'  import/no-unresolved

/Users/ericliu/Projects/popcode/src/validations/Validator.js
  6:27  error  Missing file extension for "strip-markdown"        import/extensions
  6:27  error  Unable to resolve path to module 'strip-markdown'  import/no-unresolved
```

ericliu121187:
@outoftime Thanks for the in-depth feedback! Reviewed and don't see any that warrant a discussion right now. I will be going through and making changes.

src/css/application.css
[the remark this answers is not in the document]
ericliu121187: I retract my previous statement. Needed to install packages. 😅

outoftime:
@ericliu121187 sounds good! lmk when it’s ready for another review pass : )

src/css/application.css
```diff
@@ -73,6 +73,75 @@ body {
...
+  top: 50%;
+  transform: translate(-50%, -50%);
+  width: 0;
+}
+
+.identity-confirmation__confirm-button:active::before {
+  padding-top: 120%;
+  transition: width 0.2s ease-out, padding-top 0.2s ease-out;
+  width: 120%;
+}
```
ericliu121187: Regarding all the button CSS rules, it's to imitate the Google Material Design feel. Thought it fit in with the look and feel of popcode. Lmk if you think it should go in a different direction.

[Check it out here!](https://cl.ly/nEAF)

ericliu121187:
Per suggestion, trimmed copy - 

copy

ericliu121187:
Updated UI:

buttons

I'd prefer them to be top/buttom of each other, since it saves screen real estate.

src/reducers/user.js
ericliu121187: I understand the issue and would like to discuss before tackling it.

1. How do we convey to Firebase that the user is confirmed or not? Should we update the user data in Firebase after the user has confirmed identity? Maybe make a call similar to `saveProviderInfo()` in `firebase.js` with a boolean flag `isUserAuthenticated` on `additionalUserInfo` object.

1. When `applicationLoaded`, we can then check `isUserAuthenticated` and set the correct `authenticationState` in the store.

outoftime:
@ericliu121187 UI looks good to me!

src/reducers/user.js
outoftime: @ericliu121187 we talked about this a bit on Slack but just to have it all in one place (and also I don’t think I was expressing myself very clearly):

I don’t think confirmation state is a Firebase-level concern (it’s not a persistent fact about the _user_; it’s a fact about the current user session).

I do think we want to vary the behavior on the application level based on what sequence of events yields to the user being authenticated via Firebase:

* If the user is authenticated in Firebase at application load time, just set their state directly to `AUTHENTICATED` (this should only happen if they already have a Popcode session open, or did very recently, so confirmation shouldn’t be needed)
* If the user clicks “login” and then successfully authenticates via Firebase, set the state to `PENDING_CONFIRMATION` and show the dialog
* If the user’s Firebase auth state changes to authenticated in any other scenario (meaning they logged in in a different Popcode window), set the state directly to `AUTHENTICATED`

The tricky thing about distinguishing between the latter two is that we simply [listen for changes in the Firebase auth state](https://github.com/popcodeorg/popcode/blob/3603fc88adc5b2f166832e6c5d10c11eea70c4ca/src/components/Workspace.jsx#L197-L202) and dispatch actions accordingly, totally agnostic to what initiated the auth state change.

I think probably the ideal solution would be to refactor the auth system to use [channels](https://github.com/redux-saga/redux-saga/blob/master/docs/advanced/Channels.md) to be able to consume Firebase auth changes with sagas. Then I think we could just set up a suitable system of `take`s to do the right thing in various situations.

Let me know your thoughts…!

outoftime:
@ericliu121187 OK! So basically the idea I have is to add some extra spice to the [saga that handles non-initial auth state changes](https://github.com/popcodeorg/popcode/blob/master/src/sagas/user.js#L42-L51), to wit, within the saga loop:

* Use a [`race`](https://redux-saga.js.org/docs/api/#raceeffects) to [`take`](https://redux-saga.js.org/docs/api/#takepattern) *either* an auth state change from the channel, or the `LOG_IN` action
* If we get an auth state change, it’s implicit, so just dispatch an action that changes their state to `AUTHENTICATED` or `UNAUTHENTICATED` respectively, and we’re done
* If, however, we get a `LOG_IN` first, then we need another `race`. This one listens for either an auth state change (implying that login was successful) or an `AUTHENTICATION_FAILED` action (this does not currently exist but could easily be dispatched in the `logIn()` saga)
  * If this one was an auth state change, log in was successful, and we should dispatch something to set the authentication state to `PENDING_CONFIRMATION`
  * If it was an `AUTHENTICATION_FAILED`, we’re done

Let me know if that makes sense! I admit it all seems rather complex, I think it’s mostly down to the fact that Firebase only gives us an interface to listen to all auth state changes, whether they are initiated by user action or not, so at some point we need to do some fancy footwork to distinguish between the user-initiated and the implicit ones.

outoftime:
@ericliu121187 thank you for all your hard work on this—at this point though we have switched to using Google for primary auth, which always includes a step where the user clicks on the account they want to use for login. So, unfortunately I don’t think we need this feature on the Popcode end anymore! I’m going to go ahead and close the PR.

PR #1001, opened 2017-08-21
Update project picker

pwjablonski:
Switched order and font size for project list item
From

screen shot 2017-08-20 at 11 50 41 pm

To
screen shot 2017-08-20 at 11 49 42 pm


outoftime:
@pwjablonski thanks for taking the initiative on this! but………are you sure it’s better? I consciously decided to put the date at the top of the hierarchy because it’s a much more reliable signal than the project title, which is very often *Project Title* (as in your screenshot!) But perhaps with the curriculum materials consistently using descriptive `<title>`s, and the title bar in the UI, that’ll be less of a concern? lmk what you think : )

pwjablonski:
@outoftime Hey! I think the project title gives the most information about what is inside the project and is the first thing people will look at to determine which project they want to navigate to, so my thought is that should have most prominence. You are right that Page Title is often the project title and can be confusing if the majority say Page Title. That being said, I believe all the popcode starter code have been updated with titles which should help address that problem.

outoftime:
@pwjablonski OK I’m down!

PR #1013, opened 2017-09-01
Fix unecessary ugly scrollbar in hamburger menu

alexpelan:
I noticed this during training and it bothered me.

Before:
screen shot 2017-08-31 at 8 25 02 pm

After:
screen shot 2017-08-31 at 8 25 18 pm

### DIFF
--- a/src/components/TopBar/HamburgerMenu.jsx
+++ b/src/components/TopBar/HamburgerMenu.jsx
@@ -56,19 +56,19 @@ const HamburgerMenu = createMenu({
           key="social"
         >
           <a
-            className="u__icon"
+            className="u__icon top-bar__menu-item-icon"
             href="https://github.com/popcodeorg/popcode"
             rel="noopener noreferrer"
             target="_blank"
           >&#xf09b;</a>
           <a
-            className="u__icon"
+            className="u__icon top-bar__menu-item-icon"
             href="https://twitter.com/popcodeorg"
             rel="noopener noreferrer"
             target="_blank"
           >&#xf099;</a>
           <a
-            className="u__icon"
+            className="u__icon top-bar__menu-item-icon"
             href="https://slack.popcode.org/"
             rel="noopener noreferrer"
             target="_blank"
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -195,7 +195,7 @@ body {
   min-width: 100%;
   max-height: 50vh;
   max-width: 30em;
-  overflow: scroll;
+  overflow: auto;
 }
 
 .top-bar__menu-item {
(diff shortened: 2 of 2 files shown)


outoftime:
Oof, I’m spoiled by Mac OS with the default unobtrusive scrollbars! 😬

src/css/application.css
```diff
@@ -194,7 +194,6 @@ body {
   min-width: 100%;
   max-height: 50vh;
   max-width: 30em;
-  overflow: scroll;
```
outoftime: We do need scrolling for the project picker once you get more than a handful of projects, though—I think this should probably be `overflow: auto`?

src/css/application.css
```diff
@@ -211,8 +210,7 @@ body {
 }
 
 .top-bar__menu-item_icons {
-  letter-spacing: 1.5em;
-  margin-right: -1.5em;
+  letter-spacing: 1em;
```
outoftime: Do we need to change this?

src/css/application.css
```diff
@@ -211,8 +210,7 @@ body {
 }
 
 .top-bar__menu-item_icons {
-  letter-spacing: 1.5em;
-  margin-right: -1.5em;
```
outoftime: If we do change the `letter-spacing`, we should still have an offsetting negative margin so that the icons are centered (because `letter-spacing` creates space *after* every letter, not just between them, which is DUMB)

alexpelan:
Oh yeah, I purposefully turn on obtrusive scrollbars to catch bugs like this one

src/css/application.css
alexpelan: yep

alexpelan:
So, I ended up changing the 1.5em of `letter-spacing` to 0.75em of `padding` on each side - this is more intuitive to me and we don't have to do the negative margin to compensate.

I am not a BEM expert - I believe we lint for violations? and it looks good there, but wasn't sure if I should give the _icon class a different name since it's a descendent of the _icon class.

alexpelan:
Also, I want to interactive rebase before merge, so don't merge before I do that

src/components/TopBar/HamburgerMenu.jsx
```diff
@@ -47,19 +47,19 @@ export default function HamburgerMenu({
       </a>
       <div className="top-bar__menu-item top-bar__menu-item_icons">
         <a
-          className="u__icon"
+          className="u__icon top-bar__menu-item_icon"
```
outoftime: @alexpelan in BEM-speak, this means “the icon variant of the menu item element of the top bar block”, which I don’t think is what we mean (each `<a>` is not itself a menu item). I would probably call this `top-bar__menu-item-icon` (“the menu item icon element of the top bar block”)

src/css/application.css
```diff
@@ -211,11 +211,14 @@ body {
...
   text-align: center;
 }
 
+.top-bar__menu-item_icon {
+  padding-left: 0.75em;
+  padding-right: 0.75em;
```
outoftime: 👍

alexpelan:
@outoftime good to go with me if it is for you

PR #1016, opened 2017-09-04
Clicking outside menus closes them

### LINKED ISSUE #1003: Clicking outside drop-down menu should close it
Submitter: Erica
Email: erica@scripted.org
When you click the Libraries drop-down menu and select a library like jQuery, you must click the Libraries link to make the drop-down disappear. Would like to see a click-outside handler for the Libraries menu!

outoftime:
The proximate/user-facing effect of this change is that clicking outside open menus closes them, which is an expected/good UX pattern.

Toward that end, we refactor the top bar menus to use a common component factory, `createMenu`. This looks something like a higher-order component, although I’m not sure it really qualifies as an HOC because the component passed into the HOC is just the label, whereas the output component is a fully-functional menu.

In the process of this refactor, I also discovered that using the `components/index` and `containers/index` modules to provide named exports of components/containers, while aesthetically pleasing, causes circular dependencies which result in imported components that are undefined at the time of module evaluation. This hasn’t been an issue when imported components are only dereferenced in render functions, but when creating a higher-order component, we pass a component at module evaluation time, so it can’t be undefined. Anyway, the behavior of the module loader in this situation is deeply unintuitive (I spent several hours trying to figure out what was going on).

So, introduces the `webpack-circular-dependencies` plugin and eliminates the usage of the `index` modules, which were the cause of the only circular dependencies in our code. The plugin will cause compilation to fail if circular dependencies are introduced in the future.

Fixes #1003


src/components/TopBar/HamburgerMenu.jsx
```diff
@@ -1,73 +1,84 @@
...
+  name: 'hamburger',
 
+  renderItems({
+    isExperimental,
+    isGistExportInProgress,
+    isUserAuthenticated,
+    onExportGist,
+    onExportRepo,
+  }) {
+    return tap([], (items) => {
```
inlinestyle: I'm curious, are you using `tap` here to make it clear from the top that `items` is what we'll be returning? Or is there some other functional purpose?

src/components/TopBar/createMenu.jsx
```diff
@@ -0,0 +1,107 @@
...
+        dispatch(closeTopBarMenu(name));
+      },
+
+      onToggle() {
+        dispatch(toggleTopBarMenu(name));
+      },
+    };
+  }
+
+  return function createMenuWithMappedProps(Label) {
```
inlinestyle: A name like `LaunchButton` or `MenuLaunchButton` or `MenuLauncher` might be a little more helpful for figuring out what this argument is for (assuming that I've got it figured out 😑).

src/components/TopBar/ProjectPicker.jsx
```diff
@@ -0,0 +1,38 @@
...
+      <MenuItem
+        isEnabled={projectKey === currentProjectKey}
+        key={projectKey}
+        onClick={partial(onChangeCurrentProject, projectKey)}
+      >
+        <ProjectPreview projectKey={projectKey} />
+      </MenuItem>
+    ));
+  },
+})(ProjectPickerButton, ProjectPreview);
```
inlinestyle: Hmm does the second argument get used? It looks like these args go to `createMenuWithMappedProps`, which seems to have an arity of 1.

TEST_PLAN.md
```diff
@@ -22,20 +22,46 @@ parts that seem relevant to your code changes.
...
+  contains the content that you exported.
 
-### Gist import
+### Gist import & project instructions
 
 * Open [a gist import
   URL](http://localhost:3000/?gist=339c841617fb50c98420d9f37654039d) and ensure
   that the gist is imported into the environment
+* Open [a gist with
+  instructions](http://localhost:3000/?gist=911a82a17a280545858d2d8ecc557ef3).
```
inlinestyle: Oh hey that's me! Note to self: update these instructions with an example of a gist that has non-included syntax in https://github.com/popcodeorg/popcode/pull/1000.

src/components/TopBar/HamburgerMenu.jsx
outoftime: Yeah, maybe because I’ve written a lot of Ruby where it’s a very common pattern, but my instinct is usually to `tap` when the situation is “create some object/data structure, mutate it a bunch, and then return it”. Does it feel awkward to you?

src/components/TopBar/ProjectPicker.jsx
outoftime: OOPS YOU ARE RIGHT

outoftime: (My initial implementation took a second component argument which was used to render items, but that proved insufficiently flexible)

PR #1019, opened 2017-09-05
Fix #584 text-shadow with multiple values

### LINKED ISSUE #584: text-shadow with multiple values generates erroneous error
The following does not work:

```css
.gold-on-gold {
   text-shadow: rgba(0,0,0,0.1) -1px 0, rgba(0,0,0,0.1) 0 -1px,
   rgba(255,255,255,0.1) 1px 0, rgba(255,255,255,0.1) 0 1px,
   rgba(0,0,0,0.1) -1px -1px, rgba(255,255,255,0.1) 1px 1px; 
   color: gold;
   background: gold;
}
```

jwang1919:
Fixes bug https://github.com/popcodeorg/popcode/issues/584

Gave commas between CSS parenthesis statements like `rgba` leeway in the PrettyCSS validation.

I found that doing so inadvertently created another bug where multiple commas would flag as valid. I added another check to deal with this as well as added the requisite validations in the test case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
   "bugs": "https://trello.com/b/ONaFg6wh/popcode",
   "license": "MIT",
   "dependencies": {
-    "PrettyCSS": "^0.3.13",
+    "PrettyCSS": "^0.3.15",
     "ajv": "^5.2.2",
     "array-to-sentence": "^1.1.0",
     "bowser": "^1.4.5",
--- a/src/validations/css/prettycss.js
+++ b/src/validations/css/prettycss.js
@@ -53,7 +53,8 @@ const errorMap = {
   },
 
   'extra-tokens-after-value': (error, source) => {
-    const lineNumber = error.token.line;
+    const errorToken = error.token;
+    const lineNumber = errorToken.line;
 
     if (lineNumber > 1) {
       const lines = source.split('\n');
@@ -61,7 +62,7 @@ const errorMap = {
       const thisLine = lines[lineNumber - 1];
 
       if (
-        error.token.charNum - 1 === /\S/.exec(thisLine).index &&
+        errorToken.charNum - 1 === /\S/.exec(thisLine).index &&
         !endsWith(trim(previousLine), ';')
       ) {
         return {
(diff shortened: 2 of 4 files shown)


outoftime:
@jwang1919 so I think we might be at a point where we should just contribute a fix upstream to `PrettyCSS`. We’ve [done so](https://github.com/fidian/PrettyCSS/pull/44) [before](https://github.com/fidian/PrettyCSS/pull/45) and the process was pretty painless—the library is well-factored and it’s easy to add tests and such.

I would lean in that direction because PrettyCSS itself has a full tokenizer and lexer built in, so adding support for a given syntactic construct is much more robust and expressive than trying to work around a parser error using regular expressions. (I admit to being a little lost looking at the regexps you introduce here, and I consider myself reasonably competent with regexps).

WDYT?

jwang1919:
@outoftime Yeah that sounds like a better option to me as well.
I'll revisit this branch once I submit the PR and it gets accepted in the PrettyCSS library.

The regexp was a difficult to come up with, but not too hard to interpret.
It looks for any strings that has a comma followed by another comma with any number of blank spaces in between.
```
, // look for comma
(?= // lookahead assertion, only look for these things
\s* // zero to many blank spaces
, // a comma
) // close lookadhead assertion
```

outoftime:
@jwang1919 gotcha. yeah I’ve only worked with lookaheads a couple times, that’s what threw me off.

FWIW, we don’t necessarily need to wait for PrettyCSS to integrate your patch—we could temporarily target https://github.com/popcodeorg/PrettyCSS (which I think I did in the past while waiting for PRs to get merged)

jwang1919:
Okay, made the changes and submitted PRs to both the [original branch](https://github.com/fidian/PrettyCSS/pull/51) and the [popcode fork](https://github.com/popcodeorg/PrettyCSS/pull/1). Once those are dealt with, I can change this branch's code.

jwang1919:
When possible, please review again. I have pushed the fix onto the PrettyCSS library (the maintainer merged it pretty quickly) and I've updated the PrettyCSS library.

Issue #348, opened 2016-10-22
Reduce load times


outoftime:
The goal here is to optimize load time **for a fresh session**, i.e. navigating to `https://popcode.org` without having been recently logged in, no gist or snapshot URL, etc.

- [ ] Investigate lodash’s contribution to bundle size—seems like we’re not getting the benefits of a la carte module loading. Should we use the a la carte packages instead?
- [ ] Replace or trim down `moment`
- [x] Replace or trim down `text-encoding`
- [ ] Asynchronously load libraries (e.g. the inlined copy of jQuery that’s used when it’s enabled for a project)
- [ ] Asynchronously load vendor code that isn’t required to bootstrap a fresh session (e.g. GitHub, Firebase clients)
- [x] Vendor split to improve cacheability

[Webpack visualizer](https://chrisbateman.github.io/webpack-visualizer/) is a great tool for finding the worst offenders in our bundle size.

outoftime:
https://github.com/siddharthkp/bundlesize looks like a pretty good way to make sure we lock in improvements here

outoftime:
Cross-post from #1031:

I feel like it was broken at some point but the [Webpack Visualizer](https://chrisbateman.github.io/webpack-visualizer/) is working great now, ton of low-hanging fruit in here. I’d be inclined to try pulling out some of the big libraries that we don’t need for minimal bootstrap (`firebase/database` and `github` are two obvious ones) into async loads, but feel free to follow your nose/heart as well!

outoftime:
@ajgreenb also one other thing I’m noticing that may be worth investigating—when I do a cold load and inspect the Network tab, many resources seem to be spending an alarming amount of time in a “Request to ServiceWorker” phase. Not sure what’s up with it but might be worth investigating…

ajgreenb:
Cool, I'll check in on both of those things. Earlier tonight, I tried taking a recording using the Performance tab of the developer tools from a refresh but the recording never loaded (the tab just hung.) Have you tried that/had any luck?

outoftime:
@ajgreenb hmm yep, was able to just run a Performance recording of the production app loading now—[here’s a dump](https://cl.ly/232w2L3d1P0t/Profile-20171011T092344.json) if it’s useful!

ajgreenb:
[...]

PR #1021, opened 2017-09-06
Reduce bundle size a bit

outoftime:
Step 1 of N in reducing the size of the initial bundle load. This removes non-English `moment` locales (since English is currently the only supported localization across the application). Trims bundle size by about 7%.

Also adds `source-map-explorer`, which is proving to be a really useful tool for tracking down the biggest contributors to bundle size.

Relevant to #348

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -206,6 +206,11 @@ module.exports = {
         ),
         use: ['null-loader'],
       },
+      {
+        test: /\.js$/,
+        include: matchModule('moment/locale'),
+        use: ['null-loader'],
+      },
     ],
   },
 
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,7 +1,6 @@
 /* eslint-env node */
 /* eslint-disable import/unambiguous */
 /* eslint-disable import/no-commonjs */
-/* eslint-disable comma-dangle */
 
 const fs = require('fs');
 const path = require('path');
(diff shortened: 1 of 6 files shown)


outoftime:
@ajgreenb started poking at bundle size a couple weeks ago, hadn’t bothered to cut a PR, but here’s what I came up with. LMK what you think and feel free to take it from here!

webpack.config.js
```diff
@@ -202,6 +202,11 @@ module.exports = {
           ]
         ),
         use: ['null-loader'],
+      },
+      {
+        test: /\.js$/,
```
outoftime: N.B. the English locale does not live in `moment/locale` for some reason (presumably American exceptionalism)

ajgreenb: Out of curiosity, any reason why you use null-loader for this instead of webpack's [IgnorePlugin](https://webpack.js.org/plugins/ignore-plugin/)?

outoftime: lol good call, ignoring Moment locales is literally the example use

ajgreenb:
I'll get a PR up in the next day or two with a couple updates. Was seeing some odd behavior where webpack would split a bundle when using `System.import()` but not `import()` so I have to dig into that a little more.

webpack.config.js
outoftime: @ajgreenb ok so I played around with IgnorePlugin and I ran into require errors when using it to remove unneeded parts of the `stylelint` library, which we currently do with the null loader. Given that the two seem to be roughly a wash in terms of interface/structure, I’m inclined to just keep using the null loader.

PR #1023, opened 2017-09-07
Code improvements to click-outside

outoftime:
Followup to #1016 based on post-merge code review:

* Better parameter name for wrapped component in `createMenu`
* Remove unused second argument to `createMenu` in `ProjectPicker` component


outoftime:
@inlinestyle I’m also down to remove the `tap` if you think it’d be better!

PR #1025, opened 2017-09-10
Element highlighter

pwjablonski:
Hey! This is my first pass. Would love to get feedback if I'm on the right track in approaching this.

This feature detects if the cursor is located in within a CSS block and determines the associated selector.  It then sends the selector to the preview window and adds a highlighter div to cover the corresponding elements based on the elements' positions. 

screen shot 2017-09-10 at 4 39 29 pm

I used [C9 core](https://github.com/c9/core) as a starting point specifically
[livecss,js](https://github.com/c9/core/blob/2c74d6b799544490a6d3f87862ef3a8cda802e9f/plugins/c9.preview/static/livecss.js)

Couple Points
- Works for the CSS only
- Highlight scheme (shadowing/ color)  is borrowed from C9. Might want to update.
- Doesn't include the selector label with the highlight as C9
- Doesn't display padding margin on elements which would be great to have.
- Currently, works only if cursor between braces of CSS block

For HTML
This is a lot more tricky and still wrapping my head around it. Quite a few more steps to think through. 
- C9 used an HTML Instrumentor and Simple DOM.


pwjablonski:
Also this bug... Doesn't update if the HTML is updated through javascript. For example

screen shot 2017-09-10 at 5 57 31 pm

outoftime:
@pwjablonski wow!! this is so cool!!!

In terms of high-level approach this definitely makes sense—I would propose some fairly big changes to the tactics at various levels (how we identify the selector in the CSS in particular) but the basic mechanics of picking up the selector from the current cursor position and then posting a message over to the preview window seem exactly right. Gonna post a bunch of line-level (but still fairly high-level) feedback in a few minutes!

src/components/Editor.jsx
```diff
@@ -136,12 +136,60 @@ class Editor extends React.Component {
...
+      this._sendSelectorToHighlighter(selector);
+    });
     session.setAnnotations(this.props.errors);
     this._editor.setSession(session);
     this._editor.moveCursorTo(0, 0);
     this._resizeEditor();
   }
 
+  _locateSelectorCSS(session) {
+    const reCssQuery = /(^|.*\})(.*)\{|\}/;
```
outoftime: So my high level feedback on this is that regular expressions don’t make very good language parsers. This is both because regexp-based parsing tends to be hard to understand, and also because it tends to be brittle and miss edge cases. I can tell you that at least the first concern applies here ; )

My pitch would be to instead use a bona fide CSS parser to do the job. I looked at a few options and I think [`postcss`](https://github.com/postcss/postcss) is the move. `postcss` is already a dependency of `stylelint`, so using it here won’t mean incorporating another library and increasing bundle size (although I do think we should load it asynchronously with `System.import()`).

I think the task at hand is fairly simple with `postcss`, to wit:

* Iterate over every rule in the parsed stylesheet using [`root.walkRules()`](http://api.postcss.org/Root.html#walkRules)
* Check to see if the cursor position falls within the rule boundaries specified in the `source` property of the rule. If not, move on.
* Calculate the bounds of the rule’s selector. The beginning of the selector is imply the beginning of the rule. You can calculate the end of the selector by getting the length of the `selector` property of the rule, and then adding it to the beginning. (Note that properly handling multi-line selectors would be a mildly more complicated version of the same process)
* See if the cursor position falls within the selector bounds you calculated

Let me know if this sounds reasonable! I think ultimately it’ll be no more code than what you’ve got here, but probably considerably easier to take in and more robust.

src/components/Editor.jsx
```diff
@@ -136,12 +136,60 @@ class Editor extends React.Component {
...
+      }
+      this._sendSelectorToHighlighter(selector);
+    });
     session.setAnnotations(this.props.errors);
     this._editor.setSession(session);
     this._editor.moveCursorTo(0, 0);
     this._resizeEditor();
   }
 
+  _locateSelectorCSS(session) {
```
outoftime: I would propose moving this logic into a separate module, probably just a function that takes `lines` and `cursor` parameters.

src/components/Editor.jsx
```diff
@@ -136,12 +136,60 @@ class Editor extends React.Component {
...
+      }
+    }
+    return query;
+  }
+
+  _sendSelectorToHighlighter(selector) {
+    window.frames[0].postMessage(JSON.stringify({
+      type: 'highlightElement',
+      selector: {selector},
+    }), '*');
```
outoftime: So, here we’re bypassing Redux and passing state directly from a React component through the DOM. I am a stickler for materializing all state in the Redux store.

Ergo: I think we need a property somewhere in the Redux store (I don’t have an amazing proposal for where) that holds the current selector; then the preview component should consume that property, watch it for changes, and use `postMessage` to notify the preview frame via `postMessage`.

src/util/generatePreview.js
```diff
@@ -39,6 +39,63 @@ const errorHandlerScript = `(${function() {
   };
 }.toString()}());`;
 
+const elementHighlighterScript = `(${function() {
```
outoftime: The amount of code in here is getting pretty intense and it’s making me want to find a way to organize it properly. I think my ideal situation would be to have something like `src/previewFrame/index.js` which then can load submodules like `src/previewFrame/elementHighlighter.js` all with nice `imports` and everything.

I *think* the cleanest(?) way to do this would be a Webpack loader that evaluates the source file as an entry point and does a full-on Webpack compilation of it. As far as I can tell such a thing does not currently exist, probably because the concept is pretty diabolical.

It’s not clear to me that the proposed reorganization should block this PR, though.

src/util/generatePreview.js
```diff
@@ -39,6 +39,63 @@ const errorHandlerScript = `(${function() {
...
+    removeCovers();
+    generateCovers();
+  });
+
+  window.addEventListener('resize', () => {
+    removeCovers();
+    generateCovers();
+  });
+
+  window.addEventListener('scroll', () => {
```
outoftime: What’s the thinking here? Attaching basically _anything_ to a scroll handler without throttling/debouncing is a recipe for performance problems.

src/util/generatePreview.js
```diff
@@ -39,6 +39,63 @@ const errorHandlerScript = `(${function() {
...
+      highlighterElements.forEach((highlighterElement) => {
+        highlighterElement.remove();
+      });
+    }
+  }
+
+  function generateCovers() {
+    if (selector !== '') {
+      const elements =
+      window.document.querySelectorAll(selector);
```
outoftime: We don’t need all these `window.`s

src/util/generatePreview.js
```diff
@@ -39,6 +39,63 @@ const errorHandlerScript = `(${function() {
...
+          window.document.createElement('div'));
+        cover.className = 'popcode_HIGHLIGHTER';
+        cover.style.position = 'fixed';
+        cover.style.zIndex = 10000000000;
+        cover.style.border = '1px solid rgba(238, 238, 48, 0.5)';
+        cover.style.boxSizing = 'border-box';
+        cover.style.boxShadow = '0 0 20px rgba(0,0,0,0.5)';
+        cover.style.pointerEvents = 'none';
+        cover.style.color = 'black';
+        cover.style.background = 'transparent';
```
outoftime: I think it would make sense to put all these style declarations in a stylesheet in the preview document. (Obviously the below, being dynamically generated, should stay here)

src/util/generatePreview.js
```diff
@@ -155,6 +211,13 @@ class PreviewGenerator {
     this.previewBody.appendChild(scriptTag);
   }
 
+  _addElementHighlighterHandling() {
+    const scriptTag = this.previewDocument.createElement('script');
+    scriptTag.innerHTML = elementHighlighterScript;
+    this.previewBody.appendChild(scriptTag);
+  }
+
+
```
outoftime: Extra whitespace

src/util/generatePreview.js
```diff
@@ -39,6 +39,63 @@ const errorHandlerScript = `(${function() {
   };
 }.toString()}());`;
 
+const elementHighlighterScript = `(${function() {
+  let selector = '';
```
outoftime: Why declare this in the outer semi-global scope? Can’t we just assign it to a local inside the event listener and then pass it as an argument to `generateCovers`?

outoftime:
@pwjablonski OK just left a bunch of thoughts, some very high level, some literally concerning whitespace. I think I’ll probably have a decent amount more low-level feedback later in the process, but at this point I think the main goal should be to come to an agreement on the approach we use to the big technical challenges of this PR. Let me know what you think!

src/components/Editor.jsx
pwjablonski: Yes - this sounds good! Much simpler and easier to understand. Will try it out! Do you think it makes sense to do only if the cursor is within the bounds selector text or the entire rule?

src/components/Editor.jsx
pwjablonski: Yes this makes sense!

src/util/generatePreview.js
pwjablonski: Okay! I want to update the cover's position when the user scrolls or resizes the window since its location is based off its position relative to the frame.  There was no thinking here,  I hadn't used throttling or debouncing before. Looking into how to incoperate throttling and debouncing.

src/util/generatePreview.js
pwjablonski: Yup!

src/util/generatePreview.js
pwjablonski: Sounds cleaner! Where does it make sense to store the CSS source for this?

src/util/generatePreview.js
pwjablonski: :)

src/util/generatePreview.js
pwjablonski: This is to store the selector so the scroll and resizes listeners can update with the given selector.

src/components/Editor.jsx
pwjablonski: This sounds good! Where would the file for this module live?

src/util/generatePreview.js
pwjablonski: How would you suggest importing lodash into the preview frame?

src/components/Editor.jsx
outoftime: I don’t feel super-strongly about it but intuitively I think just the selector itself would feel the most natural.

Also, I might have forgotten to mention this, but I think we should only show the highlights when the editor is focused (I think that would also reduce the pressure on the frame-side code to keep the highlight boxes perfectly in sync at all times)

src/components/Editor.jsx
outoftime: Probably just in `util`, the junk drawer for useful but hard-to-categorize modules : )

src/util/generatePreview.js
outoftime: I would probably just hand-roll a throttle function, at least until we do [this](https://github.com/popcodeorg/popcode/pull/1025#pullrequestreview-62624580). That said, it’s not clear to me why we need to handle scroll at all—resize makes sense but given that the highlight boxes are absolutely positioned, why would they need to be recalculated on scroll?

src/util/generatePreview.js
outoftime: Probably in the `templates` directory, and then just import it as plain text?

src/util/generatePreview.js
outoftime: Sure, but we can accomplish the same thing by declaring a `const selector = selectorFromMessage` in the event listener, and then passing it as an argument to `generateCovers`, right?

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
...
+
+  function updateCovers(selector) {
+    const highlighterElements =
+    window.document.querySelectorAll('.popcode_HIGHLIGHTER');
+    if (highlighterElements.length !== 0) {
+      highlighterElements.forEach((highlighterElement) => {
+        highlighterElement.remove();
+      });
+    }
+
```
pwjablonski: Didn't include scroll or resizing handling / throttling this time around.  Removing highlights when editor not focused will remove the need for a scroll handler. Trying to find ways to prevent this from happening....

screen shot 2017-09-20 at 4 03 04 pm

templates/highlighterCss.txt
```diff
@@ -0,0 +1,11 @@
+.popcode_HIGHLIGHTER {
```
pwjablonski: Wanted to put this file in src/previewFrame but the import wouldn't work

src/util/cssSelectorAtCursor.js
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    const root1 = postcss.parse(source);
+    root1.walkRules((rule) => {
+      const ruleStartRow = rule.source.start.line;
+      const ruleStartCol = rule.source.start.column;
+      const ruleEndRow = rule.source.end.line;
+      const ruleEndCol = rule.source.end.column;
+      const cursorRow = cursor.row + 1;
+      const cursorCol = cursor.column + 1;
+      if ((cursorRow > ruleStartRow && cursorRow < ruleEndRow) ||
```
pwjablonski: Only works for the entire rule for now.

src/components/PreviewFrame.jsx
```diff
@@ -100,6 +110,13 @@ class PreviewFrame extends React.Component {
     });
   }
 
+  _postHighlighterSelectorToFrame(selector) {
+    window.frames[0].postMessage(JSON.stringify({
```
outoftime: Generally a React component should never make any assumptions about the structure of the DOM outside of that which it directly controls (and try to limit assumptions even of the DOM it does control). So, the preferred way to handle this is to add a [`ref` handler](https://facebook.github.io/react/docs/refs-and-the-dom.html) to the `<iframe>`, and then store it as a property of the component instance (e.g. `this._frame_element`).

src/components/PreviewFrame.jsx
```diff
@@ -122,9 +139,14 @@ class PreviewFrame extends React.Component {
 }
 
 PreviewFrame.propTypes = {
+  highlighterSelector: PropTypes.string,
   src: PropTypes.string.isRequired,
   onRuntimeError: PropTypes.func.isRequired,
 };
 
+PreviewFrame.defaultProps = {
+  highlighterSelector: '',
```
outoftime: I would probably make the default `null` here.

src/components/Workspace.jsx
```diff
@@ -145,6 +147,14 @@ class Workspace extends React.Component {
     this.props.dispatch(focusLine(language, line, column));
   }
 
+  _handleEditorCursorChange(selector) {
```
outoftime: Ugh, sorry we still haven’t containerized the editors…

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
+const elementHighlighterScript = `(${function() {
+  window.addEventListener('message', (message) => {
+    const data = JSON.parse(message.data);
+    const {selector: {selector: selectorFromMessage}} = data;
+    const selector = selectorFromMessage;
```
outoftime: You can just do `const {selector: {selector}} = data`.

(Or just `const {selector} = data` if you want to remove the extra nesting per my preceding comment)

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
+const elementHighlighterScript = `(${function() {
+  window.addEventListener('message', (message) => {
+    const data = JSON.parse(message.data);
+    const {selector: {selector: selectorFromMessage}} = data;
```
outoftime: We should check `data.type` to make sure this is an element highlighter message.

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
...
+    const selector = selectorFromMessage;
+    updateCovers(selector);
+  });
+
+  function updateCovers(selector) {
+    const highlighterElements =
+    window.document.querySelectorAll('.popcode_HIGHLIGHTER');
```
outoftime: Completely aesthetic, but I’d be inclined to use a class name like `__popcode-highlighter` to more closely fit the naming convention we use in the main application CSS (with the `__` prefix serving to notate that it’s “internal”)

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
...
+    if (highlighterElements.length !== 0) {
```
outoftime: It’s fairly idiomatic to just do `if (highlighterElements.length)` since zero is falsy

outoftime: Also you don’t really need to check the length at all since the `forEach` is just a no-op if it’s empty.

src/previewFrame/elementHighlighterScript.js
outoftime: How does one cause that to happen?

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
...
+      highlighterElements.forEach((highlighterElement) => {
+        highlighterElement.remove();
+      });
+    }
+
+    if (selector !== '') {
+      const elements =
+      document.querySelectorAll(selector);
```
outoftime: Indent (also this probably doesn’t need to be on its own line?)

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
...
+    const {selector: {selector: selectorFromMessage}} = data;
+    const selector = selectorFromMessage;
+    updateCovers(selector);
+  });
+
+  function updateCovers(selector) {
+    const highlighterElements =
+    window.document.querySelectorAll('.popcode_HIGHLIGHTER');
+    if (highlighterElements.length !== 0) {
+      highlighterElements.forEach((highlighterElement) => {
```
outoftime: I prefer `for (const highlighterElement of highlighterElements)` in modern JS

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
...
+        highlighterElement.remove();
+      });
+    }
+
+    if (selector !== '') {
+      const elements =
+      document.querySelectorAll(selector);
+      elements.forEach((element) => {
```
outoftime: Ibid. `for`…`of`

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
...
+        const rect = element.getBoundingClientRect();
+        const cover =
+        document.documentElement.appendChild(
```
outoftime: Ibid. indentation

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
+const elementHighlighterScript = `(${function() {
```
outoftime: This is really nicely factored!

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
...
+      elements.forEach((element) => {
+        const rect = element.getBoundingClientRect();
+        const cover =
+        document.documentElement.appendChild(
+          document.createElement('div'));
+        cover.className = 'popcode_HIGHLIGHTER';
+        cover.style.left = `${rect.left - 1}px`;
+        cover.style.top = `${rect.top - 1}px`;
+        cover.style.width = `${rect.width + 2}px`;
+        cover.style.height = `${rect.height + 2}px`;
```
outoftime: What’s the reason for the additional pixels? Can we avoid having to do this by using a `border-box`?

src/components/PreviewFrame.jsx
```diff
@@ -100,6 +110,13 @@ class PreviewFrame extends React.Component {
     });
   }
 
+  _postHighlighterSelectorToFrame(selector) {
+    window.frames[0].postMessage(JSON.stringify({
+      type: 'highlightElement',
```
outoftime: I think we should stick with the convention established for posting errors e.g. use `org.popcode.highlightElement`

src/previewFrame/elementHighlighterScript.js
```diff
@@ -0,0 +1,36 @@
+const elementHighlighterScript = `(${function() {
+  window.addEventListener('message', (message) => {
+    const data = JSON.parse(message.data);
```
outoftime: We should add a try/catch here—in theory there’s no guarantee that the message is JSON.

src/previewFrame/index.js
```diff
@@ -7,57 +7,16 @@ import uniq from 'lodash/uniq';
 import loopBreaker from 'loop-breaker';
 import libraries from '../config/libraries';
 import previewFrameLibraries from '../config/previewFrameLibraries';
+import highlighterCss from '../../templates/highlighterCss.txt';
+import errorHandlerScript from './errorHandlerScript';
+import elementHighlighterScript from './elementHighlighterScript';
+import alertAndPromptReplacementScript
+  from './alertAndPromptReplacementScript';
```
outoftime: I really like this refactor—particularly that the modules are exported as strings of source code. Would make me very nervous otherwise. But a very nice 80/20 vs. the more radical proposal I had made initially.

src/previewFrame/index.js
```diff
@@ -7,57 +7,16 @@ import uniq from 'lodash/uniq';
 import loopBreaker from 'loop-breaker';
```
outoftime: I love that you’ve broken the preview generator into multiple submodules, but I think the whole thing should probably still live in `src/util` (e.g. `src/util/previewFrame/index` etc.). The general principle is that top-level directories inside `src` should group together architecture-level categories of code (reducers, sagas, components, etc.) rather than specific concerns. `util` is kind of the “junk drawer” for stuff that doesn’t fit neatly into a category with other extant modules (not to say the code in there is junk!)

templates/highlighterCss.txt
outoftime: I think that’s basically correct—webpack is configured to treat everything in here as a raw text file, whereas it would assume files elsewhere are JavaScript. I would probably name this `highlighter.css` though?

src/previewFrame/index.js
pwjablonski: Moved to util

src/previewFrame/elementHighlighterScript.js
pwjablonski: This was a remnant of the cloud9 implementation. Took them out and replaced with border-box

src/previewFrame/elementHighlighterScript.js
pwjablonski: yup!

src/previewFrame/elementHighlighterScript.js
pwjablonski: updated

src/previewFrame/elementHighlighterScript.js
pwjablonski: yup!

src/previewFrame/elementHighlighterScript.js
pwjablonski: updated

src/previewFrame/elementHighlighterScript.js
pwjablonski: Removed

src/previewFrame/elementHighlighterScript.js
pwjablonski: Updated

src/previewFrame/elementHighlighterScript.js
pwjablonski: Added conditional

src/previewFrame/elementHighlighterScript.js
pwjablonski: Oops forgot to add! will add in another commit

src/previewFrame/elementHighlighterScript.js
pwjablonski: :)

src/components/PreviewFrame.jsx
pwjablonski: Changed to null

src/components/PreviewFrame.jsx
```diff
@@ -100,6 +110,13 @@ class PreviewFrame extends React.Component {
...
+      selector: {selector},
```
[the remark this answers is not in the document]
pwjablonski: changed to just selector

src/components/PreviewFrame.jsx
pwjablonski: changed to org.popcode.highlightElement

src/components/PreviewFrame.jsx
pwjablonski: This makes sense. I added the ref but had to add allow-same-origin. Is this okay?

src/components/Preview.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import isNil from 'lodash/isNil';
 import partial from 'lodash/partial';
 import classnames from 'classnames';
-import generatePreview, {generateTextPreview} from '../util/generatePreview';
+import generatePreview, {generateTextPreview} from '../previewFrame/index';
```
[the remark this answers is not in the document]
pwjablonski: Updated to ../previewFrame

src/components/Editor.jsx
```diff
@@ -136,6 +137,14 @@ class Editor extends React.Component {
     session.on('change', () => {
       this.props.onInput(this._editor.getValue());
     });
+    session.selection.on('changeCursor', () => {
+      if (this.props.language === 'css') {
+        const cursor = session.selection.lead;
+        const highlighterSelector =
+        cssSelectorAtCursor(this._editor.getValue(), cursor);
+        this.props.onCursorChange(highlighterSelector);
```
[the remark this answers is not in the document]
pwjablonski: I went the first route here adding the logic of converting in a selector to the saga.  I haven't used saga's before so not sure if I implemented it correctly.

src/components/Editor.jsx
```diff
@@ -136,6 +137,14 @@ class Editor extends React.Component {
     session.on('change', () => {
       this.props.onInput(this._editor.getValue());
     });
+    session.selection.on('changeCursor', () => {
+      if (this.props.language === 'css') {
+        const cursor = session.selection.lead;
+        const highlighterSelector =
+        cssSelectorAtCursor(this._editor.getValue(), cursor);
```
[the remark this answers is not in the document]
pwjablonski: Indented line 144

src/components/Editor.jsx
```diff
@@ -136,6 +137,14 @@ class Editor extends React.Component {
     session.on('change', () => {
       this.props.onInput(this._editor.getValue());
     });
+    session.selection.on('changeCursor', () => {
+      if (this.props.language === 'css') {
```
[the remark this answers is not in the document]
pwjablonski: Added language as a parameter of selectorAtCursor

src/components/Editor.jsx
```diff
@@ -136,6 +137,14 @@ class Editor extends React.Component {
     session.on('change', () => {
       this.props.onInput(this._editor.getValue());
     });
+    session.selection.on('changeCursor', () => {
```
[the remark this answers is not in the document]
pwjablonski: This is fired only when the editor is focused I believe

src/clients/github.js
```diff
@@ -3,7 +3,7 @@ import GitHub from 'github-api';
 import isEmpty from 'lodash/isEmpty';
 import trim from 'lodash/trim';
 import performWithRetries from '../util/performWithRetries';
-import generatePreview from '../util/generatePreview';
+import generatePreview from '../previewFrame/index';
```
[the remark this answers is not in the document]
pwjablonski: Changed to '../previewFrame'

src/actions/ui.js
```diff
@@ -42,6 +42,11 @@ export const userDismissedNotification = createAction(
   type => ({type}),
 );
 
+export const updateHighlighterSelector = createAction(
+  'UPDATE_HIGHLIGHTER_SELECTOR',
+  selector => ({selector}),
```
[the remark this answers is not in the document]
pwjablonski: Removed from action

src/actions/ui.js
```diff
@@ -42,6 +42,11 @@ export const userDismissedNotification = createAction(
   type => ({type}),
 );
 
+export const updateHighlighterSelector = createAction(
+  'UPDATE_HIGHLIGHTER_SELECTOR',
```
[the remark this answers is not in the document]
pwjablonski: Changed to CURRENT_CURSOR_CHANGED

package-lock.json
```diff
@@ -0,0 +1,10511 @@
+{
```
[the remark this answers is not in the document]
pwjablonski: Removed

templates/highlighterCss.txt
pwjablonski: Okay thanks! Changed to highlighter.css

src/previewFrame/elementHighlighterScript.js
pwjablonski: Since the positions of the highlighted elements are fixed,  when the screen is scrolled the elements move down screen but the highlighter elements remain fixed.

src/components/PreviewFrame.jsx
outoftime: I’m a frayed knot—this opens up a security hole where an attacker could send a snapshot or gist that attempts to break out of the preview iframe and hijack Popcode’s access to the coder’s GitHub account. As far as I know, though, setting up a `ref` should behave exactly the same as `window.frames[0]` here—what happened that spurred you to add `allow-same-origin`?

src/components/Preview.jsx
```diff
@@ -3,7 +3,8 @@ import PropTypes from 'prop-types';
 import isNil from 'lodash/isNil';
 import partial from 'lodash/partial';
 import classnames from 'classnames';
-import generatePreview, {generateTextPreview} from '../util/generatePreview';
+import generatePreview, {generateTextPreview}
+  from '../util/previewFrame';
```
outoftime: Does this need to be on a second line?

src/components/PreviewFrame.jsx
outoftime: @pwjablonski just checked out the branch and removed the `allow-same-origin`—seems to work fine?

pwjablonski: I figured it out.
I had `console.log(this._frame_element.contentWindow);`
Which gave the error "Blocked a frame with origin "http://localhost:3000" from accessing a cross-origin frame"
Just tried it again without allow-same-origin and it works just fine. Sorry for the confusion.

src/components/Preview.jsx
pwjablonski: nope it fits on one line

outoftime:
@pwjablonski playing around with it, a few thoughts:

* The actual highlighter boxes are pretty visually intense. Maybe chill out the shadow a bit?
* I am confident the scrolling problem is solvable by using absolute positioning rather than fixed. It does mean we have to do a little more math to figure out what the `top` and `left` values of the boxes should be. I have gotten pretty close messing around with it locally. The main thing is to, first, attach the box to `document.body` rather than `document.documentElement`. Then, call `document.body.getBoundingClientRect()`, and subtract the `top` and `left` values of that rect from each element’s rect to get the offset of the element within the body. However this seems to still be a _bit_ off in some circumstances, which I will continue poking at
* I notice that if your cursor is at the very beginning or end of a rule, there’s no highlight. Would it be easy to change that?
* I think it would be worth it to remove the highlights when the editor is unfocused (I think this should be pretty straightforward)

This is really cool!!

pwjablonski:
@outoftime awesome!

1. Yeah definitely can chill out the highlighter shadow. Another remnant of cloud9 inspiration. Any preference on the border color?
2. That seems like a better alternative if we can avoid handling the scroll. Interested in what you've got so far.
3. Yes that should be easy enough
4. Great will look into removing the highlights when the editor is unfocused

outoftime:
@pwjablonski OK I think I figured out a robust way to calculate the right absolute (not fixed) position for the highlighter elements. It eschews use of `getBoundingClientRect()` altogether in favor of using `offsetTop` and `offsetLeft`. [Here’s a diff that shows what I changed to get it popping](https://gist.github.com/outoftime/fdfe9c2718ffda742b1a2526ee83bbb0).

Also of note is that I had to change back to traditional `for` loops rather than `for…of`—I discovered that the latter breaks in older browsers because we don’t import the Babel polyfill into the preview environment, and `for…of` relies on `Symbol`. Sad!

Anyway, lmk what you think / when you want me to do another pass!

pwjablonski:
This is great! 

1. We will still need to handle the window resize, correct?
2. It seems as though the body in the preview is given an extra 8px of margin so the cover for the rule
`body {
}` is off by a bit
3. If an element is given a position of fixed by the User ie. `p{postion:fixed}` the cover does not appear. I think it is because in this case, the values of offsetTop, offsetHeight etc. for the element are null when the position is set to fixed.

Looking to push some of the other changes soon!

pwjablonski:
Sorry I haven't forgotten about this! Looking to get some changes in soon.

pwjablonski:
@outoftime Made some updates so that the highlighter is removed when the editor is not focused but still have to update several things...

1. Will need to handle window resize
2. The body in the preview is given an extra 8px of margin so the cover for the rule
body { } is off by a bit. 
3. If an element is given a position of fixed by the User ie. p{postion:fixed} the cover does not appear. I think it is because in this case, the values of offsetTop, offsetHeight etc. for the element are null when the position is set to fixed.
4. Minor - but highlighter is not added if the cursor is moved from a rule in CSS, to another editor, and back into the same rule.

outoftime:
@pwjablonski hey, let me know if you want me to help out with any of the above / when you’re ready for another review pass. Obviously this isn’t the most pressing thing we’re working on right now but it is a very sick feature and also an appealing return to Popcode’s core instructional focus!

pwjablonski:
@outoftime trying to update to use previewSupport.Not sure I am approaching this correctly. I don't fully grasp how webpack is being used here with the preview. 

 Currently, it looks like the `updateCovers` function is being called continuously in `channel.bind` in `handleElementHighlights.js` and the cover is never appended to the document body.

I would appreciate some guidance!

pwjablonski:
@outoftime I took a stab at handling these issues. Getting close! 

1. Will need to handle window resize
2. The body in the preview is given an extra 8px of margin so the cover for the rule
body { } is off by a bit.
3. If an element is given a position of fixed by the User ie. p{postion:fixed} the cover does not appear. I think it is because in this case, the values of offsetTop, offsetHeight etc. for the element are null when the position is set to fixed.
4. Highlighter is not added if the cursor is moved from a rule in CSS, to another editor, and back into the same rule.

src/reducers/ui.js
```diff
@@ -18,6 +18,8 @@ const defaultState = new Immutable.Map().
     typing: false,
     requestedFocusedLine: null,
     textSizeIsLarge: false,
+    focusedSelector: null,
+    focusedEditor: null,
```
outoftime: Do we actually need both of these pieces of information? From a standpoint of application state, it seems like all we _really_ need to know is the `focusedSelector`—we don’t care which editor the selector came from. It does make sense to have separate *actions* that represent changes to editor focus and cursor position, but it’s not clear that the state itself needs to capture both pieces of information (e.g. we can just clear `focusedSelector` when the editor is blurred).

src/sagas/compiledProjects.js
```diff
@@ -10,6 +10,7 @@ import {
 import every from 'lodash/every';
 
 import {getCurrentProject, getErrors} from '../selectors';
+
```
outoftime: Doesn’t seem like a strictly related change ; )

src/sagas/ui.js
```diff
@@ -50,10 +55,19 @@ export function* exportProject() {
...
   yield all([
     debounceFor('UPDATE_PROJECT_SOURCE', userDoneTyping, 1000),
     takeEvery('POP_OUT_PROJECT', popOutProject),
+    takeEvery('CURRENT_CURSOR_CHANGED', updateFocusedSelector),
+    takeEvery('EDITOR_FOCUSED', updateFocusedSelector),
     takeEvery('EXPORT_PROJECT', exportProject),
   ]);
 }
+
+
```
outoftime: Extra whitespace

src/selectors/getHighlighterSelector.js
```diff
@@ -0,0 +1,3 @@
+export default function getHighlighterSelector(state) {
+  return state.getIn(['ui', 'editors', 'focusedSelector']);
+}
```
outoftime: Doesn’t seem like this selector is being used?

src/util/selectorAtCursor.js
```diff
@@ -0,0 +1,27 @@
+import postcss from 'postcss';
+
+export function selectorAtCursor(source, cursor, language) {
+  let highlighterSelector = '';
+  if (language === 'css') {
+    try {
+      const root1 = postcss.parse(source);
```
outoftime: Why not just `root`?

src/util/selectorAtCursor.js
```diff
@@ -0,0 +1,27 @@
...
+      root1.walkRules((rule) => {
+        const ruleStartRow = rule.source.start.line;
+        const ruleStartCol = rule.source.start.column;
+        const ruleEndRow = rule.source.end.line;
+        const ruleEndCol = rule.source.end.column;
+        const cursorRow = cursor.row + 1;
+        const cursorCol = cursor.column + 1;
+        if ((cursorRow > ruleStartRow && cursorRow < ruleEndRow) ||
```
outoftime: A newline after the very first `(` on this line would be more consistent with project style

src/util/selectorAtCursor.js
```diff
@@ -0,0 +1,27 @@
...
+          (cursorRow === ruleStartRow && cursorCol >= ruleStartCol) ||
+          (cursorRow === ruleEndRow && cursorCol <= ruleEndCol)) {
```
outoftime: Ibid. newline before `)` here

src/util/selectorAtCursor.js
```diff
@@ -0,0 +1,27 @@
+import postcss from 'postcss';
+
+export function selectorAtCursor(source, cursor, language) {
+  let highlighterSelector = '';
```
outoftime: Echoing an earlier comment seems like this should be `null` if we don’t find a selector.

src/util/selectorAtCursor.js
```diff
@@ -0,0 +1,27 @@
...
+        const ruleStartRow = rule.source.start.line;
+        const ruleStartCol = rule.source.start.column;
+        const ruleEndRow = rule.source.end.line;
+        const ruleEndCol = rule.source.end.column;
+        const cursorRow = cursor.row + 1;
+        const cursorCol = cursor.column + 1;
+        if ((cursorRow > ruleStartRow && cursorRow < ruleEndRow) ||
+          (cursorRow === ruleStartRow && cursorCol >= ruleStartCol) ||
+          (cursorRow === ruleEndRow && cursorCol <= ruleEndCol)) {
+          highlighterSelector = rule.selector;
```
outoftime: Per the [API docs](http://api.postcss.org/global.html#childIterator) we can `return false` here to break iteration if we’ve found a match, which is a mild gain for runtime efficiency.

src/util/selectorAtCursor.js
```diff
@@ -0,0 +1,27 @@
...
+        }
+      });
+      return highlighterSelector;
+    } catch (e) {
+      return highlighterSelector;
```
outoftime: I’d just `return null` here for clarity.

src/util/selectorAtCursor.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+  }
+  return highlighterSelector;
```
outoftime: Ibid `return null`

templates/highlighter.css
```diff
@@ -0,0 +1,12 @@
+.__popcode-highlighter {
+  position:absolute;
```
outoftime: Space after the colon is project style (applies to several below lines as well

templates/highlighter.css
```diff
@@ -0,0 +1,12 @@
...
+  z-index: 10000000000;
+  border:1px solid rgba(238, 238, 48, 0.2);
+  box-sizing:border-box;
+  box-shadow:0 0 20px rgba(0,0,0,0.2);
+  pointer-events: none;
+  color: black;
+  background: transparent;
+  box-sizing: border-box;
+}
+
```
outoftime: Got an extra line of whitespace here

templates/highlighter.css
```diff
@@ -0,0 +1,12 @@
+.__popcode-highlighter {
+  position:absolute;
+  z-index: 10000000000;
+  border:1px solid rgba(238, 238, 48, 0.2);
+  box-sizing:border-box;
+  box-shadow:0 0 20px rgba(0,0,0,0.2);
+  pointer-events: none;
+  color: black;
```
outoftime: I don’t think we need to be setting the (text) color of this? No text, right?

templates/highlighter.css
```diff
@@ -0,0 +1,12 @@
+.__popcode-highlighter {
+  position:absolute;
+  z-index: 10000000000;
+  border:1px solid rgba(238, 238, 48, 0.2);
```
outoftime: Is this accessible?

templates/highlighter.css
```diff
@@ -0,0 +1,12 @@
...
+  box-sizing:border-box;
+  box-shadow:0 0 20px rgba(0,0,0,0.2);
+  pointer-events: none;
```
outoftime: 👍

test/unit/sagas/compiledProjects.js
```diff
@@ -4,6 +4,7 @@ import {
   validatedSource as validatedSourceSaga,
 } from '../../../src/sagas/compiledProjects';
 import {getCurrentProject, getErrors} from '../../../src/selectors';
+
```
outoftime: Also does not seem strictly related ; )

templates/highlighter.css
```diff
@@ -0,0 +1,12 @@
...
+  color: black;
+  background: transparent;
+  box-sizing: border-box;
```
outoftime: Hmm, won’t this mean the border itself sits on top of the inner edge of the highlighted element? Could we instead use `content-box` and just give it a -1px margin top/left?

yarn.lock
```diff
@@ -592,7 +592,7 @@ babel-cli@^6.23.0:
   optionalDependencies:
     chokidar "^1.6.1"
 
-babel-code-frame@^6.22.0, babel-code-frame@^6.26.0:
+babel-code-frame@^6.11.0, babel-code-frame@^6.22.0, babel-code-frame@^6.26.0:
```
outoftime: Are the changes in this file intentional? It doesn’t look like there are any new dependencies by way of `package.json` so unless you needed to upgrade an existing dependency I wouldn’t expect to see changes in this file.

src/sagas/ui.js
```diff
@@ -1,12 +1,17 @@
...
+  currentFocusedSelectorChanged,
+} from '../actions/ui';
 import {getCurrentProject} from '../selectors';
 import {
   projectExportDisplayed,
   projectExportNotDisplayed,
 } from '../actions/clients';
 import {openWindowWithContent} from '../util';
+
+import {selectorAtCursor} from '../util/selectorAtCursor';
```
outoftime: Right now `postcss` isn’t part of the main (blocking) bundle, and it would be good for load time to keep it that way! Let’s import this asynchronously, like we [do with validations](https://github.com/popcodeorg/popcode/blob/master/src/sagas/errors.js#L16-L23).

src/util/selectorAtCursor.js
```diff
@@ -0,0 +1,27 @@
+import postcss from 'postcss';
+
+export function selectorAtCursor(source, cursor, language) {
```
outoftime: Would be very nice to have some unit tests for this : )

src/components/Workspace.jsx
```diff
@@ -130,6 +138,28 @@ class Workspace extends React.Component {
     );
   }
 
+  _handleErrorClick(language, line, column) {
+    this.props.dispatch(focusLine(language, line, column));
+  }
```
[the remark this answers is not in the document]
pwjablonski: Yup!

src/components/Editor.jsx
```diff
@@ -99,6 +100,10 @@ class Editor extends React.Component {
       this._startNewSession(this.props.source);
       this._resizeEditor();
       this._editor.on('focus', this._resizeEditor);
+      this._editor.setOptions({
+        fontFamily: 'Inconsolata',
+        fontSize: '14px',
+      });
```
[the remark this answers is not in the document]
pwjablonski: Gotcha!

src/previewSupport/handleElementHighlights.js
```diff
@@ -0,0 +1,54 @@
...
+  if (element === document.body) {
+    return {top: 0, left: 0};
+  }
+  const {top, left} = getOffsetFromBody(element.offsetParent);
+  return {top: top + element.offsetTop, left: left + element.offsetLeft};
+}
+
+function removeCovers() {
+  const highlighterElements =
+  window.document.querySelectorAll('.__popcode-highlighter');
```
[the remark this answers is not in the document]
pwjablonski: +

src/previewSupport/handleElementHighlights.js
```diff
@@ -0,0 +1,54 @@
...
+  for (let i = 0; i < highlighterElements.length; i++) {
```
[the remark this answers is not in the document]
pwjablonski: Does this still apply?

https://github.com/popcodeorg/popcode/pull/1025#issuecomment-335042104

src/previewSupport/handleElementHighlights.js
```diff
@@ -0,0 +1,54 @@
...
+    const highlighterElement = highlighterElements[i];
+    highlighterElement.remove();
+  }
+}
+
+function addCovers(selector) {
+  const elements = document.querySelectorAll(selector);
+  for (let i = 0; i < elements.length; i++) {
```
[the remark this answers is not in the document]
pwjablonski: Same as above
https://github.com/popcodeorg/popcode/pull/1025#issuecomment-335042104

src/previewSupport/handleElementHighlights.js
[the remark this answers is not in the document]
pwjablonski: Yup!

src/previewSupport/handleElementHighlights.js
```diff
@@ -0,0 +1,54 @@
...
+    const element = elements[i];
+    const cover = document.createElement('div');
+    document.body.appendChild(cover);
+    const rect = element.getBoundingClientRect();
+    let offset = {top: rect.top, left: rect.left};
+    if (element.offsetParent === null) {
+      cover.style.position = 'fixed';
+    } else if (element !== document.body ||
+      element !== document.documentElement) {
```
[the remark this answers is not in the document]
pwjablonski: The body seems to have an additional 8px of padding. So the box is off by a bit using getOffsetFromBody

src/previewSupport/handleElementHighlights.js
```diff
@@ -0,0 +1,54 @@
...
+      offset = getOffsetFromBody(element);
+    }
+    cover.className = '__popcode-highlighter';
```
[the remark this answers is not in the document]
pwjablonski: nice

src/previewSupport/handleElementHighlights.js
```diff
@@ -0,0 +1,54 @@
...
+    cover.style.left = `${offset.left}px`;
+    cover.style.top = `${offset.top}px`;
+    cover.style.width = `${element.offsetWidth}px`;
+    cover.style.height = `${element.offsetHeight}px`;
+  }
+}
+
+function updateCovers(selector) {
+  removeCovers();
+  if (selector !== '') {
```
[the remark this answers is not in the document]
pwjablonski: agreed

src/reducers/ui.js
```diff
@@ -90,6 +97,24 @@ export default function ui(stateIn, action) {
...
+      return state.setIn(
+        ['editors', 'focusedEditor'],
+        null,
+      );
+
+    case 'EDITOR_RESIZED':
+      return state.setIn(
+        ['editors', 'focusedSelector'],
+        null,
+      );
```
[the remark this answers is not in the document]
pwjablonski: This is to remove the highlighter in the event of the resizing. Alternatively, we could just update cover with new window size on  resize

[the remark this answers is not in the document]
pwjablonski: This is more about the resize of the entire window instead of the editors but works for both

src/util/selectorAtCursor.js
pwjablonski: I think root is keyword

src/util/selectorAtCursor.js
pwjablonski: great!

src/reducers/ui.js
pwjablonski: I don't think so. There was a reason why I did it this way but it's not coming to mind. Will look into it.

src/util/selectorAtCursor.js
outoftime: Ah, maybe `rootNode` then?

src/previewSupport/handleElementHighlights.js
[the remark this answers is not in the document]
outoftime: Yes it does… I feel like we should just pull in the polyfill though?

templates/highlighter.css
pwjablonski: @kaylee42 I think we should talk about this one too. 

Its currently based on the cloud9 color scheme but thimble might better jumping off point https://thimble.mozilla.org

src/reducers/ui.js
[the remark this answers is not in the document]
outoftime: OK—in that case I think we should make it explicit rather than piggybacking on a resize listener that seems more or less unrelated.

Question: don’t we actually want to listen for resize inside the iframe? Seems like that’s the real question, and also I think that would catch certain cases we are missing now (e.g. moving the vertical divider changes the preview frame size).

As a corollary, I think right now a resize will fully remove the highlights; better I think to just recalculate their positions?

src/previewSupport/handleElementHighlights.js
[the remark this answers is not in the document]
outoftime: Actually looking again I think this condition will always evaluate to true, since it’s saying “if the element isn’t the body OR it isn’t the document root”? So maybe no need for the condition at all if it behaves correctly as written?

src/sagas/ui.js
pwjablonski: Looked at validations example but couldn't quite figure the right way to implement for selector at cursor. what should this look like? @outoftime

yarn.lock
pwjablonski: not sure why this happened

pwjablonski:
@outoftime made many of the suggested changes
1. Unit test needs review
2. Looked at validations example but couldn't quite figure the right way to implement for selector at cursor. what should this look like?
3. not quite sure what you mean [here](https://github.com/popcodeorg/popcode/pull/1025#discussion_r171726874) re getoffsetbody
4. I played around with the highlight to mimic thimbles highlighter. might be a bit intense but worth a look.

pwjablonski:
@outoftime brought this one up to date and added some tests.

Updated the highlight to fade into an outline, but may still be a bit invasive. Open to suggestions!

outoftime:
Closing as #1598 is the active continuation of this work.

PR #1031, opened 2017-09-17
Add vendor code-splitting

ajgreenb:
This PR introduces code-splitting for vendor libraries. It also adds scaffolding for caching in production. When `NODE_ENV === 'production'` in `webpack.config.js`, the files will be tagged with a hash of their contents. As far as I can tell, our build process currently doesn't allow us to set `NODE_ENV` in `webpack.config.js` since it gets `require`d before `NODE_ENV` gets set in the `env` gulp task. To address this, I could

1. Make `webpack.config.js` export a function that takes an environment parameter and use that to modify the webpack configuration.
2. Add a `build:production` script in `package.json` that looks something like `NODE_ENV=production gulp build --production`.

Thoughts?


ajgreenb:
Longer-term thinking: would you ever consider dropping gulp altogether and just using webpack for all building?

src/index.html
```diff
@@ -14,6 +14,5 @@
         <div class="pop-throbber__image"></div>
       </div>
     </div>
-    <script src="application.js"></script>
   </body>
 </html>
```
ajgreenb: Note that I moved `src/static/index.html` to `src/index.html` since it's now used as a template for html-webpack-plugin.

outoftime: OK—I would say let’s maybe move it to `src/html/index.html` (mirroring `src/css/application.css`)? Realistically all the JS should be in `src/js` but that’s a reorganization for another day…

webpack.config.js
```diff
@@ -253,6 +261,27 @@ module.exports = {
       ],
       ServiceWorker: {navigateFallbackURL: '/'},
     }),
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'vendor',
+      minChunks({ context }) {
```
outoftime: Project style would be no space inside the curly braces

webpack.config.js
```diff
@@ -253,6 +261,27 @@ module.exports = {
...
+        if (context) {
+          const isNodeModule = context.indexOf('node_modules') !== -1;
+          const isBowerComponent = context.indexOf('bower_components') !== -1;
+          return isNodeModule || isBowerComponent;
+        }
+        return false;
```
outoftime: *Very* minor / I don’t feel strongly about this at all, but I would invert this and return `false` early if `context` is not truthy, so that the “meat” of the function body describes the expected case.

webpack.config.js
```diff
@@ -66,8 +70,12 @@ module.exports = {
   entry: './src/application.js',
   output: {
     path: path.resolve(__dirname, './dist'),
-    filename: 'application.js',
-    sourceMapFilename: 'application.js.map',
+    filename: process.env.NODE_ENV === 'production'
```
outoftime: I think it’s worth declaring a local `const isProduction` since we’re asking this question several times.

ajgreenb:
In regard to pushing Firebase permissions changes on deploy: keeping gulp around just for that seems like overkill. If webpack doesn't cut it, maybe a simple `package.json` script would suffice? What does that gulp task actually do? I'm only have a surface-level understanding of Firebase.

ajgreenb:
In regard to caching, I'm sure there's a solution to the service worker holding onto every version of the asset files. I'll do a little research and to find out the right way.

ajgreenb:
In regard to the `DeprecationWarning`, I'll dig around and see what I can find out about that. I suspect it's caused by an out-of-date plugin.

ajgreenb:
Just pushed a few more changes. A couple of things I'd like to bring to your attention:

1. We are (and were) minifying in development. Is that the right call? It's causing some Chrome console warnings about minifying development versions of React and Redux.
    * We could add the `UglifyJsPlugin` only when `NODE_ENV === 'production'`.
    * If we ultimately decide to use just webpack for building the front-end, we could skip adding the `UglifyJsPlugin` altogether and just execute `webpack -p`, which automatically adds it and also "tree-shakes".
2. I added a `prod` task to the `package.json` scripts. Is that the right idea?
    * Will that require changes to the Travis CI configuration?

outoftime:
> We are (and were) minifying in development. Is that the right call?

Whoa, really! Totally unintentional. That would explain why the build in development is so freaking slow…

package.json
```diff
@@ -173,6 +173,7 @@
     "postinstall": "bower install",
     "pretest": "yarn run lint",
     "dev": "yarn install && gulp dev",
+    "prod": "yarn install && NODE_ENV=production gulp build",
```
outoftime: I don’t think there’s a need for the `yarn install` here—that’s already a step in Travis, and best to keep separate things separate imo (the `dev` thing is to make it more convenient for humans running the dev server, but no humans to inconvenience in production builds)

webpack.config.js
```diff
@@ -16,6 +20,8 @@ const babel = require('babel-core');
 const babelLoaderVersion =
   require('./node_modules/babel-loader/package.json').version;
 
+const isProduction = process.env.NODE_ENV === 'production';
```
outoftime: So, I just had a thought—`yargs` presumably is just looking at the arguments passed to the current process when it started. That information should be no different in a `webpack.config` module that’s `require`d into the `gulpfile` than it is in the `gulpfile` itself, so could we just make this line `const isProduction = yargs.argv.production` and roll back some of the other related changes?

webpack.config.js
```diff
@@ -253,6 +259,38 @@ module.exports = {
       ],
       ServiceWorker: {navigateFallbackURL: '/'},
     }),
+    new webpack.DefinePlugin({
+      'process.env': {
+        NODE_ENV: JSON.stringify(process.env.NODE_ENV || 'development'),
```
outoftime: Do we need the `stringify` here?

ajgreenb: Yup, I think we do:

> Note that because the plugin does a direct text replacement, the value given to it must include **actual quotes** inside of the string itself. Typically, this is done either with either alternate quotes, such as `'"production"'`, or by using `JSON.stringify('production')`.

From https://webpack.js.org/plugins/define-plugin/

outoftime: well then, in my face.

ajgreenb:
Seems like offline-plugin adds its own content-based hash for cache invalidation. See

https://github.com/NekR/offline-plugin/issues/127#issuecomment-257329938

It's a matter of adding the `updateStrategy: 'changed'` option to the configuration. So could we avoid doing webpack content hashes altogether? The problem I see with this is browsers that don't support ServiceWorkers or AppCache. Does Popcode target browsers that don't support either of those?

ajgreenb:
Okay cool, so I'll leave the content-hashing and possibly add `updateStrategy: 'changed'` to the offline-plugin configuration options.

That `DeprecationWarning` is caused by the webpack-chunk-hash plugin. I can remove it, but I've noticed that sometimes webpack will change a module's hash even if its contents haven't changed.

ajgreenb:
`webpack-chunk-hash` just published `0.5.0` which fixes that `DeprecationWarning` we were seeing, so that's fixed.

ajgreenb:
@outoftime okay, looks like the default `updateStrategy` for offline-plugin is `changed`, so I don't need to do anything to enable that.

I'm not too familiar with ServiceWorkers or offline-plugin—do you know a way to view which resources have been cached? I'd like to be able to test that the cache behavior is correct (i.e., not accumulating copies of every bundle version.)

ajgreenb:
Aaand, one more thing—you can see from the Travis output that the tests fail with `Uncaught ReferenceError: webpackJsonp is not defined`. This happens locally, too. [It seems like karma-webpack doesn't support the CommonsChunkPlugin][issue].

Any preference on how to work around this issue? I could

* Use lodash to pluck out the CommonsChunkPlugins in `karma.conf.js`, or
* Use some kind of flag in `webpack.config.js` to include or omit the CommonsChunkPlugins based on whether we're testing.

[issue]: https://github.com/webpack-contrib/karma-webpack/issues/22

outoftime:
> I'm not too familiar with ServiceWorkers or offline-plugin—do you know a way to view which resources have been cached? I'd like to be able to test that the cache behavior is correct (i.e., not accumulating copies of every bundle version.)

You can go into the **Application** tab of the Chrome Dev Tools, and then look in **Service Workers** and **Cache Storage**

> Any preference on how to work around this issue?

The second option sounds better but what did you have in mind for the flag? I think my ideal solution would basically have `webpack.config.js` have a named export along the lines of `generateWebpackConfig()` that takes options and then generates a config based on those options, as well as a default export that is the result of calling that function with default options. However given we’re in CommonJS-land, we can’t actually do that…

ajgreenb:
Nice, thanks! That's a great tip. I'll test this branch when I get home and make sure that the cache behaves as expected.

webpack supports exporting a function that returns a configuration object instead of just a plain object. Check out [this link](https://webpack.js.org/guides/environment-variables/). How about having `development`, `production`, and `test` modes that enable/disable various plugins for each environment? In `webpack.config.js`, we could do something like

```js
module.exports = (env = 'development') => {
    // ...
}
```

And then we could require it in different files and manually create the configuration object by passing the flag. If that sounds good, I'll open a small merge request that just changes `webpack.config.js` into a function but leaves everything else the same. Then I'll add the mode differentiation in this branch.

outoftime:
@ajgreenb just merged the webpack function PR!

ajgreenb:
After doing some more research, it looks like offline-plugin doesn't actually care what you call your files. From what I understand, it will use its own content-based hashes as the sole source of truth for cache updates (regardless of the name of the file.)

* nekr/offline-plugin#292
* [Pending PR to update offline-plugin's documentation for `updateStrategy`](https://github.com/NekR/offline-plugin/blob/93b5369cca7910ae36385dadd507ff99e4e88042/docs/update-strategies.md)

So I think the caching behavior is identical across this branch and `master`!

ajgreenb:
I think the only thing left to do is update the `purgeCache` gulp task to stop purging the JavaScript files. The Cloudflare API *does* allow you to granularly specify what to purge, but you have to give it a list of filenames. Should that list include anything other than `index.html`? `application.css` comes to mind. Anything else?

outoftime:
I think that probably covers it!

What about caching headers on S3? Right now the assets are set to expire fairly quickly—ideally everything that’s content-hashed should have ultra long lived headers right?

ajgreenb:
Yeah, ultra-long-lived sounds good or maybe even Cache-Control: immutable? Where do you set those headers?

ajgreenb:
This may be naive/silly (not too experienced here), but what about just not sending caching headers at all from S3 since we already cache in Cloudflare (also browser caching and maybe ServiceWorker caching)? Doesn't having lots of different caching layers make it easier to end up with the wrong caching behavior?

outoftime:
> This may be naive/silly (not too experienced here), but what about just not sending caching headers at all from S3 since we already cache in Cloudflare (also browser caching and maybe ServiceWorker caching)? Doesn't having lots of different caching layers make it easier to end up with the wrong caching behavior?

Sorry, I’m a bozo—caching behavior is already configured at the CloudFlare, not the S3, level. We can just update them to be more aggressive with the JavaScript after this is released.

ajgreenb:
Sweet, well, unless I'm forgetting something, I think I've addressed all your feedback thus far. Let me know if you have further comments!

webpack.config.js
```diff
@@ -105,8 +109,39 @@ module.exports = (env = 'development') => {
       ],
       ServiceWorker: {navigateFallbackURL: '/'},
     }),
+    new webpack.DefinePlugin({
+      'process.env': {
+        NODE_ENV: JSON.stringify(isProduction ? 'production' : 'development'),
+      },
+    }),
```
outoftime: I don’t think this is needed—we [already set `NODE_ENV` using the `EnvironmentPlugin`](https://github.com/ajgreenb/popcode/blob/1b29bb9ee61c613923c55112cdc48281ddecd560/webpack.config.js#L74) (note that the values given there are just defaults in case the key isn’t defined in `process.env` at compile time).

gulpfile.js
```diff
@@ -149,6 +149,11 @@ gulp.task('purgeCache', () =>
     key: process.env.CLOUDFLARE_KEY,
   }).zones.purgeCache(
     process.env.CLOUDFLARE_ZONE,
-    {purge_everything: true},
+    {
+      files: [
+        'https://popcode.org/index.html',
+        'https://popcode.org/application.css',
```
outoftime: Let’s grab the hostname from `process.env`? It can just be added to `.travis.yml` for that to work properly.

ajgreenb:
Aw yeahhh, thanks @outoftime! I'll continue picking away at #348.

outoftime:
Sweet! I feel like it was broken at some point but the [Webpack Visualizer](https://chrisbateman.github.io/webpack-visualizer/) is working great now, ton of low-hanging fruit in here. I’d be inclined to try pulling out some of the big libraries that we don’t need for minimal bootstrap (`firebase/database` and `github` are two obvious ones) into async loads, but feel free to follow your nose/heart as well!

PR #1046, opened 2017-09-21
Fix dynamic flex in older browsers

outoftime:
We are changing flex-related values in JavaScript, but we support browsers that only take flex properties with vendor prefixes. So, use the `inline-style-prefixer` library to add appropriate vendor prefixes to the JavaScript style objects.

Also drop support for Chrome 22, which additionally has an outdated and overly strict interpretation of percentage heights that breaks Popcode’s layout. Happily, Chrome 23 works fine. According to Google Analytics, Popcode has only been loaded once using Chrome 22 since May 1, and I’m guessing this was me in BrowserStack.

### DIFF
--- a/config/browsers.json
+++ b/config/browsers.json
@@ -1,5 +1,5 @@
 {
   "firefox": "28",
-  "chrome": "22",
+  "chrome": "23",
   "safari": "7.1"
 }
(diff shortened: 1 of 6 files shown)


outoftime:
@inlinestyle relevant to your interests

inlinestyle:
LGTM 👍

PR #1049, opened 2017-09-22
Smooth preview refresh

outoftime:
Typing in the HTML is unpleasant for two reasons:

1. A spinner briefly appears while the project is asynchronously validating. This is a fairly recent change, with the introduction of custom validation, but it’s not a problem with the validation itself.  Validation is expected to be asynchronous and can take a non-trivial amount of time. The UX around it is bad.
2. Even before the spinner issue, there was a noticeable flash as the preview iframe refreshed itself.

This is a rough-and-ready sketch of a fix for these issues. Instead of using a single iframe and updating its `srcdoc` property, we instead create a new frame every time the project changes. We also keep the frame from the previous version of the compiled project in the DOM, layering the two on top of each other with absolute positioning.  Further, we set the new frame’s visibility to hidden until it has loaded. The effect is that, for the most part, one can’t see any flashes because each new frame instance appears fully-loaded on top of the last.

There is a bunch of work to do around the behavior of error messages, which are basically broken at the moment.

In general this is more of a proof-of-concept right now—as well as dealing with error messages, I need to write tests, clean up code, restore functionality I deleted for expediency, etc.

### To do

- [x] Restore project title
- [x] Restore pop-out
- [x] Excise concept of “user is typing”
- [x] Fix layout of static errors
- [x] Nice UI for transition to error state (fade in)
- [x] Fix layout of runtime errors
- [x] Why are there two iframes when the environment first loads?
- [x] Tests
- [x] Code cleanup


jwang1919:
I never really paid attention to the preview screen while typing in Popcode but I can see what you mean with the screen flashing. I think what you have so far is a good first step, but doesn't completely address the solution.

If I set the background black and hold a key down, I can still see the flashing pretty prominently. 
http://localhost:3000/?snapshot=e67bf101-4ca4-4550-91fa-783ea297becc

Maybe only render the preview when no keystroke is made in the last X seconds?

outoftime:
> Maybe only render the preview when no keystroke is made in the last X seconds?

@jwang1919 So I was initially opposed to this as I generally don’t want there to be a perceptible delay between typing and preview showing up, but I do think for the case you’re mentioning (extremely rapid-fire changes) a very tight throttle (~50ms) could do the trick without causing a perceptible delay. I shall try it out!

src/components/PreviewFrame.jsx
```diff
@@ -113,7 +123,9 @@ class PreviewFrame extends React.Component {
       <div className="preview__frame-container">
         <iframe
           className="preview__frame"
+          ref={this._attachToFrame}
           sandbox={sandboxOptions}
+          style={{visibility: 'hidden'}}
```
outoftime: Use stylesheet

src/components/Preview.jsx
```diff
@@ -1,85 +1,42 @@
...
-        {u__hidden: !isSyntacticallyValid},
       )}
     >
       <div className="preview__title-bar">
-        <span
-          className="preview__button preview__button_pop-out u__icon"
-          onClick={partial(onPopOutProject, project)}
-        >&#xf08e;</span>
-        {generateTextPreview(project)}
+        TK TITLE
```
outoftime: Fix

src/components/TopBar/index.jsx
```diff
@@ -17,7 +17,7 @@ function uiVariants({validationState, isUserTyping}) {
     return {popVariant: 'neutral'};
   }
   if (validationState === 'validating') {
-    return {popVariant: 'thinking', modifier: 'top-bar_yellow'};
+    return {popVariant: 'thinking'};
```
outoftime: We can probably keep yellow and just add a CSS transition

src/sagas/compiledProjects.js
```diff
@@ -0,0 +1,36 @@
...
+
+import every from 'lodash/every';
+
+import {getCurrentProject, getErrors} from '../selectors';
+import generatePreview from '../util/generatePreview';
+import {projectCompiled} from '../actions';
+
+function* validatedSource() {
+  const errors = yield select(getErrors);
+  if (every(errors, errorList => errorList.state === 'passed')) {
```
outoftime: This should be a selector

PR #1083, opened 2017-10-07
Set NODE_ENV in actual environment

outoftime:
`NODE_ENV` should be set to `production` in the actual environment, rather than relying on `yargs` in Gulp to take care of it. That way the env is consistent across the entire build, and we don’t need to depend on the order in which things run.


outoftime:
@ajgreenb FYI

Issue #1080, opened 2017-10-05
Expanding an editor should focus it


outoftime:
If an editor is collapsed, and the coder expands it, it should automatically be focused (this is the natural expectation).

I think we can just use the existing [`ui.editors.requestedFocusedLine` mechanism](https://github.com/popcodeorg/popcode/blob/master/src/reducers/ui.js#L66-L76) to do this.

PR #1086, opened 2017-10-09
Focus on expanded editor

YiningChen:
Focuses editor when user expands a collapsed editor! (Issue #1080)

- to allow for extra data to be stored for each component in `hiddenUIComponents`, I turned it from a Set to a Map, with the key being referenced as `componentKey` in parameters. (before it was sometimes called `componentName` and other times just `component`) 
- Also, for the language editors, I just used the language as the key instead of `editor.${language}`, as "editor" is stored as `componentType` for each hidden component.
- and created a migration function within records/Project so that old snapshots and gists will still work seamlessly! (But I'm not sure if that's the correct place for such a function xD I mainly did so to avoid a circular file reference, as any migration function would need to import the `HiddenUIComponent` record)

Also, as I was working I was wondering if instead of `hiddenUIComponents` in state, it could just be `UIComponents` with a `hidden` property. It might also make it easier to tell if new components like the console and instructions were purposefully expanded, or if they just didn't exist for old projects.


src/reducers/ui.js
```diff
@@ -24,6 +24,29 @@ const defaultState = new Immutable.Map().
   set('topBar', new Immutable.Map({openMenu: null})).
   set('lastRefreshTimestamp', null);
 
+// i.e. restore to before editors were resized
```
outoftime: I’m generally not a fan of most “explanatory” comments—it’s very easy for them to diverge from the code they seek to explain, and if variables/functions/classes are named carefully, I don’t think comments are usually necessary. In this case I think the function names `restoreDefaultColumnFlex` etc. are entirely self-documenting!

src/reducers/ui.js
```diff
@@ -51,11 +74,13 @@ export default function ui(stateIn, action) {
...
     case 'HIDE_COMPONENT':
+      return restoreFlexOnComponentToggle(action.payload.componentName, state);
+
     case 'UNHIDE_COMPONENT':
-      if (action.payload.componentName === 'output') {
-        return state.setIn(['workspace', 'rowFlex'], DEFAULT_ROW_FLEX);
-      }
-      return state.setIn(['workspace', 'columnFlex'], DEFAULT_COLUMN_FLEX);
+      state = focusEditor(
+        action.payload.componentName.replace('editor.', ''),
```
outoftime: This is definitely the most low-impact way to do what we need to do here, but I’m not in love with the idea of parsing a string that’s just being used as an internal identifier. Seems like it might be better to change the structure of the action a bit so that the information we need is discretely available. Maybe something like `{componentType: 'editor', componentId: 'javascript'}`?

src/reducers/ui.js
```diff
@@ -51,11 +74,13 @@ export default function ui(stateIn, action) {
...
+        state);
```
outoftime: Very minor nitpick, but project style would be to put the `)` on the next line : )

src/reducers/ui.js
```diff
@@ -24,6 +24,29 @@ const defaultState = new Immutable.Map().
...
+  return state.setIn(['workspace', 'columnFlex'], DEFAULT_COLUMN_FLEX);
+}
+
+function restoreDefaultRowFlex(state) {
+  return state.setIn(['workspace', 'rowFlex'], DEFAULT_ROW_FLEX);
+}
+
+function restoreFlexOnComponentToggle(componentName, state) {
+  if (componentName === 'output') {
+    return restoreDefaultRowFlex(state);
```
YiningChen: It doesn't seem like the output component is ever hidden, but left this in here in case that's part of the roadmap. Happy to delete this functionality and simplify if that would be better!

outoftime: Yep, you’re right about that—it actually used to be possible, but the redesign didn’t have an output-minimization button, so I went ahead and dropped it from the initial implementation of the new layout. I think we could probably just put one in the preview top bar (next to either the refresh or the pop-out button) but also no one has mentioned missing it…

src/reducers/ui.js
YiningChen: Okay! That makes a lot of sense! Thanks :)

outoftime:
heyyy @YiningChen any idea when we might be able to get this puppy over the finish line?? it’s so close!

YiningChen:
ah!! Sorry @outoftime had a busy couple of weeks 😱, but planning on picking it back up tomorrow!

outoftime:
@YiningChen no worries, totally understand! can’t wait to ship it!

src/reducers/projects.js
```diff
@@ -88,6 +92,7 @@ export function reduceRoot(stateIn, action) {
 
 export default function reduceProjects(stateIn, action) {
   let state;
+  const {payload} = action;
```
YiningChen: I did this to get past the line length linter 😂 But let me know if this would cause issues!

outoftime: I have a policy of not really nitpicking the reducers until after I am able to refactor them to have a reasonable structure.

src/util/componentName.js
```diff
@@ -0,0 +1,8 @@
+export function makeComponentName({componentType, componentId}) {
+  if (componentId) {
+    return `${componentType}.${componentId}`;
+  }
+  return componentType;
+}
+
+
```
outoftime: Couple extra lines of whitespace here

src/util/componentName.js
```diff
@@ -0,0 +1,8 @@
+export function makeComponentName({componentType, componentId}) {
```
outoftime: @YiningChen can you share the reasoning for having this? In particular, why not just make the `{componentType, componentId}` structure the thing we use universally to describe a UI component, rather than still using the string representation in some cases?

src/reducers/projects.js
outoftime: I would actually say this is a good change on the merits—in fact, I’d probably argue for just fully destructuring `action` in the params list itself!

outoftime:
@YiningChen getting verrrrrrry close i think! sorry about the annoying component structure thing, that gives this a much bigger footprint than it would have otherwise, but i still think it’s good!

src/reducers/projects.js
YiningChen: Ah! Right that totally makes sense! thanks!

src/util/componentName.js
YiningChen: No good reasoning! I assumed there was already a reason it was a string, which is a bad reason xD Will change, thanks!

YiningChen:
No worries! It's been fun digging into more and more as well! :D

outoftime:
@YiningChen would love to ship this! it’s very close! now that the Britney Pickle Spears have been served, think you might have a chance to bring it home?

YiningChen:
lol! I have a working version that refactored every call to show/hide actions, but ran into a weird bug that I think was related to a previous saved project's schema not matching up to the new schema (where hiddenUIComponents is an array of objects instead of a set of strings). Seemed to work after forcing it to be an array with lodash's concat, but wanted to play around with it a bit more to see if there's a better solution!

outoftime:
@YiningChen gotcha! i figured that kind of issue would come up sooner or later (actually surprised it wasn’t sooner)—I think the move is probably to introduce the concept of “migrations” that would be applied to projects right when we dehydrate them from Firebase. Basically the idea I had in mind was to create a collection of functions, each of which takes a project, checks it for a particular outdated structure (in this case, strings for hidden components), and then “migrates” it to the newer structure. As we make more changes to how projects are structured, we’d just add more functions to the list. Does that seem reasonable?

YiningChen:
Oooo okay! yeah definitely! 😄 I'll aim to get that done latest soon after thanksgiving! :)

outoftime:
@YiningChen exciting! I think I’ll probably have time to really dig into this Wednesday but just perusing it a bit, I think it would be really helpful if you could update the PR description with a high-level outline of the code changes here—in particular I’m not quite grasping the idea of the “hidden component line” data structure and how it fits into the overall change we’re making here. Can’t wait to dive in deeper!

outoftime:
(Also it looks like there’s a conflict with master right now so would be great to resolve that if you have a chance)

YiningChen:
Ah yes! sorry! will do!

src/components/Editor.jsx
```diff
@@ -58,13 +58,15 @@ class Editor extends React.Component {
   }
 
   componentWillUnmount() {
+    const {onComponentHidden, language} = this.props;
+    const {row, column} = this._editor.getCursorPosition();
+    onComponentHidden(language, row, column);
```
YiningChen: Not sure if there is a better place for this action to be called, I would have ideally liked to call it from `EditorContainer.jsx`, where the hide function is called.

outoftime:
@YiningChen great writeup! looking forward to digging into this 😀

YiningChen:
Hi Matt!

That definitely makes a lot sense, and it didn't feel 100% right when I called that action from `componentWillUnmount` so I'm definitely in agreement! And storing line/column position definitely doesn't seem worth it just to restore position when an component is unhidden. Happy to dial back the implementation! How do you feel about the `HiddenUIComponent` record with `line` and `column` removed? I originally went down this route because we discussed having a `componentType` and `language` property for the editors, which I do really like having for situations like this: https://github.com/popcodeorg/popcode/pull/1086/files#diff-cf58228d41d0dc144163bbeb96dbe489R65

Also out of curiosity, how come the editors are destroyed and recreated when toggled as opposed to using CSS to hide them? I would love to follow up this in the future, and the best implementation I can think of to avoid storing line/column would to just refocus the editor using `focus()`. 

Thanks for the thorough review! :)

src/components/EditorsColumn.jsx
```diff
@@ -61,10 +62,7 @@ export default class EditorsColumn extends React.Component {
     const children = [];
     const [hiddenLanguages, visibleLanguages] = partition(
       ['html', 'css', 'javascript'],
-      language => includes(
-        currentProject.hiddenUIComponents,
-        `editor.${language}`,
-      ),
+      language => find(currentProject.hiddenUIComponents, {language}),
```
YiningChen: @outoftime My only hesitation with renaming the `language` property in `HiddenUIComponent` to `instance` is this line, since I really like the how the `language` property connects to how language is used in this file. `instance` also wouldn't be used by anything that isn't an editor, and I can't imagine we'd want multiple consoles or instructions.

src/components/Editor.jsx
```diff
@@ -63,8 +63,7 @@ class Editor extends React.Component {
   }
 
   _focusRequestedLine(requestedFocusedLine) {
-    if (get(requestedFocusedLine, 'component') !==
-      `editor.${this.props.language}`) {
+    if (get(requestedFocusedLine, 'componentKey') !== this.props.language) {
```
YiningChen: it doesn't feel 100% right here to use language as the `componentKey`, but I wanted to get rid of all `editor.${this.props.language}`, which also feels a bit arbitrary. Please let me know if you have better ideas!

YiningChen:
ready for review again!

outoftime:
@YiningChen awesome! will look at this as soon as i can!

outoftime:
@YiningChen any chance you have the bandwidth to pick this up and bring it to the finish line? Happy to help out with merge conflicts or if you’re too busy, I can take over, just let me know!

YiningChen:
ah sorry I totally forgot about this for awhile! started a new job and moved across the country xD I should have time this weekend! but if not totally fine with it being taken over :) Thanks!

outoftime:
Oh wow congrats! No wild rush at all, just wanted to see if you were up for finishing it : )

PR #1126, opened 2017-10-14
Add support for confirm and prompt functions

acingraham:
I've found confirm and prompt to be great ways to introduce user input and conditionals.  Ideally, we'd be able to support them using SweetAlert, but unfortunately, SweetAlert calls are asynchronous and use a Promise-based API which would only add to the complexity of introducing these topics.

### DIFF
--- a/src/components/PreviewFrame.jsx
+++ b/src/components/PreviewFrame.jsx
@@ -8,6 +8,7 @@ import {sourceDelimiter} from '../util/generatePreview';
 
 const sandboxOptions = [
   'allow-forms',
+  'allow-modals',
   'allow-popups',
   'allow-scripts',
   'allow-top-navigation',
--- a/src/util/generatePreview.js
+++ b/src/util/generatePreview.js
@@ -49,10 +49,6 @@ const alertAndPromptReplacementScript = `(${function() {
       },
       configurable: true,
     },
-    prompt: {
-      value: (message, defaultValue = '') => defaultValue,
-      configurable: true,
-    },
   });
 
   delete window.swal; // eslint-disable-line prefer-reflect


outoftime:
@acingraham thanks for this! happy for this to be the jumping off point for a discussion but it is quite intentional that we don’t allow any blocking dialogs in the preview environment—the behavior can be very frustrating when the preview is live-refreshing, because it throws up a dialog on every keystroke.

That’s not to say we don’t have options; a few that come to mind:

* “Debounce” the running of JavaScript in the preview (basically as-you-type we render an HTML/CSS-only version of the preview, then after ~2s without typing, we update it with the JS). Cute but seems really hard to get right and potentially confusing
* Only allow `alert()`, `prompt()`, and `confirm()` in non-initial execution contexts (so you could do it in a click handler, but not in the top-level script). Also seems potentially confusing
* Detect attempts to call `alert()` / `prompt()` in the preview environment. By default, if one of these functions is called, no-op it in the preview environment but show a message prompting the coder to “click here to allow” them. Clicking the prompt refreshes the preview with `alert()` / `prompt()` allowed (this would probably only take effect once)
* Statically detect calls to `alert()` / `prompt()` / `confirm()` in the code. When they’re there, replace the preview with a message saying something to the effect of, “since your code contains calls to a function that creates a dialog box, automatic preview refresh is disabled. Click here once you’re ready to see the preview”

I think these are pretty much in ascending order of how much I like them : )

WDYT?

acingraham:
Totally makes sense why we'd want to have some checks for them and I like your alternatives to just allowing prompt and confirm.  What do you think of a toggle, like a checkbox, that allows or ignores prompt and confirm?  Just tossing out ideas before we narrow in on a solution and get to coding!

outoftime:
@acingraham interesting idea! a couple of questions:

* Were you envisioning a control that’s always present, or displaying a checkbox specifically when we’ve detected `alert`/`confirm`/`prompt` in the code?
* How long would the box stay checked for? What are the use cases where the coder would keep it checked for a substantial period of time?

outoftime:
@acingraham since this was more of an exploration I’m going to go ahead and close it for now—feel free to pick it back up whenever you feel moved to, though!

PR #1133, opened 2017-10-16
Name anonymous webpack chunks

ajgreenb:
webpack supports [using "magic" comments][import] to name chunks created dynamically by using the `import()` function. If you don't use these comments, webpack names the files using a numeric "request ID" (e.g., `0.js`) Adding a name makes it easier to map chunks to the files webpack emits for them.

[import]: https://webpack.js.org/api/module-methods/#import-


outoftime:
@ajgreenb can you elaborate a bit on what the practical advantage is here / what we can do with the  magic comments that we can’t do without them? in general i find “machine-readable” comments to be fairly icky from a standpoint of code aesthetics, although we do use eslint comments where needed, which is fairly analogous (the comments control something about the build system that doesn’t have bearing on the way the code behaves).

src/validations/html/slowparse.js
```diff
@@ -124,7 +124,10 @@ class SlowparseValidator extends Validator {
   }
 
   async _getRawErrors() {
-    const {Slowparse} = await System.import('../linters');
+    const {Slowparse} = await import(
+      /* webpackChunkName: 'linters' */
+      '../linters',
+    );
```
outoftime: I’m not sure why this doesn’t use `importLinters` like every other validation—seems like I [just missed it in the original PR](https://github.com/popcodeorg/popcode/pull/587)

outoftime:
@ajgreenb if we do go this route, I’d propose we get rid of the `importLinters` / `linters` modules altogether, since the main purpose of those is to artificially force Webpack to load them all in one chunk. Seems like the magic comments are a more idiomatic way to do that.

We do probably need a more flexible version of the retry system, probably something that lightly wraps `promiseRetry` and lets us do something like:

```js
const htmllint = retryingFailedImports(() => import(
  /* webpackChunkName: 'linters' */
  'htmllint'
))
```

ajgreenb:
I find descriptive naming useful when running webpack's output through some of the bundle analysis tools. It's easier to tell which parts of the visualization correspond to which code when the chunks map clearly to names in the code base. In the following visualization, using names makes it obvious where in the code base these modules get imported.

webpack bundle analyzer

Granted, Popcode currently only has a single dynamically imported chunk, so this isn't much of a win right now. [Here's an issue that contains more debate about this topic.](https://github.com/webpack/webpack/issues/1949) On the whole, if you decide to pass on this, it won't be a big deal to me.

outoftime:
OK, given the module visualization argument you’ve laid out as well as a bunch of other good use cases in the linked issue, I’m convinced. I do think we should just wholesale do away with the `linters`/`importLinters` modules altogether in this PR—are you down?

ajgreenb:
Sure am! I'll update this PR with your suggestions.

ajgreenb:
Quick question: why are the linters imported asynchronously at all? According to [this line](https://github.com/popcodeorg/popcode/blob/master/src/validations/index.js#L6), it seems like they'll be imported as soon as the app loads, right? This must be necessary, since the app might load from a gist or snapshot and require immediate validation. Apart from breaking a chunk into two pieces, is there an advantage to doing it this way as opposed to just importing them normally as part of the `vendor` chunk?

Also, do you have any ideas about how to recreate the initial loading of all modules in a nice way without `importLinters`?

outoftime:
@ajgreenb great question! the reason that line is there is so that we rely on a network connection for as short a window as possible. basically we know that, at the time Popcode loads, there is a network connection (or available offline cached version), because otherwise Popcode couldn’t have loaded. so at that point we want to greedily load in all async resources, so we’ll have them when we need them, even if the network drops out later.

this is still better than just loading them as part of the main chunk because basic application bootstrap isn’t blocked on loading those resources.

in terms of how to recreate the effect without a module that bundles them all together, let me poke around a bit : )

outoftime:
@ajgreenb none of my immediate hunches panned out, although I am wondering whether the service worker will actually handle this for us? on phone now but I’m interested to know whether the SW just loads all the resources it knows about out of the gate… seems plausible that it might?

outoftime:
Woop woop, looks like as long as the stuff we want to eager load is in `main` we [should be good to go!](https://github.com/NekR/offline-plugin/blob/master/docs/caches.md)

outoftime:
(Actually I think we would want the linters in `additional`

ajgreenb:
We'll also need to replace the [line](https://github.com/popcodeorg/popcode/blob/master/test/index.js#L6) in `test/index.js` that imports `src/validations/linters` if we're doing away with that file. Any preferences on how that should be done? I could just inline the first half of `linters.js` into `test/index.js`...

outoftime:
@ajgreenb hmm, I don’t have a great idea—what we’d really want I think is basically just an across-the-board [eager `webpackMode`](https://webpack.js.org/api/module-methods/) but I’m not sure if it’s possible to do that at the webpack config level…

outoftime:
(Also it’s probably worth confirming that that whole thing is still necessary at all)

ajgreenb:
Sorry for the slow responses; I started a new job yesterday so I may not have a chance to do too much on this until the weekend. In your previous comment, which whole thing should I be confirming necessary?

outoftime:
@ajgreenb ah damn I could have sworn I already responded to this—basically the thing in `test/index` where we synchronously import the linters. In theory Karma should be able to handle async imports just fine, and maybe given all the other changes it can?

outoftime:
@ajgreenb really looking forward to shipping this!!

ajgreenb:
Any suggestions for how to use `retryingFailedImports` on `Slowparse`? In `linters.js`,`Slowparse` is imported like so:

```
import Slowparse from 'slowparse/src';
import SlowparseHTMLParser from 'slowparse/src/HTMLParser';

SlowparseHTMLParser.prototype.omittableCloseTagHtmlElements = [];

export {
  Slowparse,
};
```

I expect we wouldn't want to do

```
const Slowparse = retryingFailedImports(() => import(
  /* webpackChunkName: 'linters' */
  'slowparse/src',
));
const SlowparseHTMLParser = retryingFailedImports(() => import(
  /* webpackChunkName: 'linters' */
  'slowparse/src/HTMLParser',
));
SlowparseHTMLParser.prototype.omittableCloseTagHtmlElements = [];
```

Should we do something similar to `stylelint` where we import from `../../util/minimalStylelint`? I could make a `util/slowparse`. Thoughts?

outoftime:
@ajgreenb yes, I think the `util` approach is a good one!

(Side note: maybe there should be a top-level directory for stuff like that? Basically of the category “import a dependency, mess with it in some way, then re-export it”. I think we also do the same with Bugsnag, maybe other stuff too…)

outoftime:
Bad pitch `src/configuredDependencies`

ajgreenb:
We have some time to think on it; I'll worry about shuffling directory structure after the meat of this PR is done (or in a follow-on PR.)

src/validations/html/slowparse.js
```diff
@@ -124,7 +125,11 @@ class SlowparseValidator extends Validator {
   }
 
   async _getRawErrors() {
-    const {Slowparse} = await System.import('../linters');
+    const Slowparse = await retryingFailedImports(() => import(
+      /* webpackChunkName: 'linters' */
+      '../../util/slowparse',
+    ));
+    console.log(Slowparse);
```
ajgreenb: @outoftime not sure if you're still online, but if you are—is this an issue you've had before? See the commit message above. And try logging `Slowparse` right before the `export default Slowparse;` line to compare.

I feel like I've spent a cumulative month of my life trying to debug these weird module errors from webpack/Babel and I still don't understand what causes them 😅 

I'll debug some more as soon as I get some more time.

ajgreenb:
I just need to dig in a little bit deeper because it seems like it _is_ working properly for other imports. It's only failing with `src/util/slowparse.js`...

outoftime:
@ajgreenb good news! this is [just how dynamic imports of ES2015 modules works in Webpack](https://github.com/webpack/webpack.js.org/pull/1109). The `stylelint` import has the same problem in this branch, but is wrapped in a `try` / `catch` causing it to fail silently.

Anyway, all you need is:

```js
const {default: Slowparse} = await retryingFailedImports(…
```

Kind of annoying that `import()` treats CommonJS modules and ES2015 modules so differently though! I guess maybe unavoidable. Also funny that the spec [does not seem to contemplate this situation at all](https://github.com/tc39/proposal-dynamic-import#proposed-solution).

ajgreenb:
Ah okay, I guess the distinction is that `import()` always imports the _namespace_ rather than the default export. But with CommonJS, you don't have namespaces, just the `module.exports` object? Thanks for tracking that down for me!

outoftime:
Yep I think that pretty much sums it up!

ajgreenb:
Just want to give an update on this! The project _appears_ to work properly (i.e., has parity with popcode.org), but the tests currently do not run (see below). I'm going to set aside some time to try to figure out why on Sunday.

```
yarn test v1.0.2
$ yarn run lint
$ karma start --single-run --no-auto-watch
webpack: wait until bundle finished: 

webpack: Compiled with warnings.
23 11 2017 20:51:44.588:WARN [Chrome 62.0.3202 (Mac OS X 10.12.6)]: Disconnected (1 times), because no message in 10000 ms.
Chrome 62.0.3202 (Mac OS X 10.12.6) ERROR
  Disconnected, because no message in 10000 ms.
Chrome 62.0.3202 (Mac OS X 10.12.6): Executed 0 of 0 DISCONNECTED (12.958 secs / 0 secs)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

outoftime:
@ajgreenb just checked out the branch and took a quick look—if you open up the console in the instance of Chrome that Karma is running, it appears (unsurprisingly) to be a Webpack error when the validation tests are trying to run:

```
# valid flexbox
not ok 101 TypeError: Cannot read property '0' of undefined
  ---
    operator: error
    expected: |-
      undefined
    actual: |-
      [TypeError: Cannot read property '0' of undefined]
    stack: |-
      TypeError: Cannot read property '0' of undefined
          at Function.requireEnsure [as e] (http://localhost:9876/base/test/index.js?0f68f276793dadbfaeb2bdcecbb3fb251935bfc9:89:74)
          at http://localhost:9876/base/test/index.js?0f68f276793dadbfaeb2bdcecbb3fb251935bfc9:212031:46
          at http://localhost:9876/base/test/index.js?0f68f276793dadbfaeb2bdcecbb3fb251935bfc9:214436:12
          at http://localhost:9876/base/test/index.js?0f68f276793dadbfaeb2bdcecbb3fb251935bfc9:169915:24
          at <anonymous>
  ...
```

Ideally I would say we just want all modules loaded eagerly, ideally in one big chunk, in test mode. This is one promising-looking (if somewhat hacky) approach: https://gist.github.com/jcenturion/892c718abce234243a156255f8f52468

This suggests that setting the chunk count limit to 0 might have the desired effect, although it did not fix the error we are seeing, and others seem to have mixed success: https://github.com/webpack/webpack/issues/3637#issuecomment-270821817

outoftime:
@ajgreenb also wanted to flag that right now, if you load a fresh environment and then disable the network, `linters` will fail to load. Glancing at the OfflinePlugin configuration, it seems like it should eagerly load `linters.js`, so I am not clear on what the issue is.

outoftime:
Figured it out! It’s because `InlineChunkManifestHtmlPlugin` was being used in the test environment, where we don’t actually have a manifest chunk. If you move that into the branch that’s only applied in non-test environments, the tests run cleanly!

ajgreenb:
Ah, dude, doing my job for me 😄  fix pushed for the broken tests. Now I'll see if I can figure out what's going on with the linters failing to load when there's not connection.

ajgreenb:
Not sure how to reproduce the error with loading linters in offline mode (not much experience here.) I tried reloading the page and then quickly checking the "Offline" checkbox in the Chrome Developer Tools network tab, but linters was already loaded. What are the exact steps needed to trigger the error?

outoftime:
interesting. What I did was load the environment, click the offline button, and then type into the code to trigger linting. I got an error. Maybe I had an outdated service worker?

ajgreenb:
¯\_(ツ)_/¯

Could you try one more time? Seems to work for me. I disconnected as soon as the UI looked finished and linting appears to work.

screen shot 2017-11-26 at 4 07 21 pm

The line in the console, **[SW]: Cached assets: additional** suggests to me that the right thing is happening.

outoftime:
@ajgreenb I can still reproduce the issue—and in fact, it appears that service worker offline caching has stopped working generally in this branch (no idea why).

To reproduce the original issue:

1. Open a _fresh_ Popcode environment (make sure it‘s not rehydrating a project from local storage—this will trigger an immediate validation)
2. Use the developer tools to switch to offline mode
3. Observe that the Pop throbber persists forever as `linters.js` tries to load with retries

To reproduce the overall non-working service worker issue:

1. Open a Popcode environment
2. Use the developer tools to switch to offline mode
3. Command-L + Enter to perform a cache-friendly refresh
4. Observe that the application fails to meaningfully load. Interestingly, the initial request for the HTML document _does_ seem to get successfully served from the service worker.

I can confirm that neither of these issues happen in the master branch.

Definitely happy to help debug if you get stuck—very excited to release this, so I hope this ends up being a simple fix : )

outoftime:
Here’s something weird: If you try to load the environment in offline mode, most of the resources don’t load, e.g. `vendor.js`.

BUT, if (still in offline mode), you just type `http://localhost:3000/vendor.js` into the browser’s location bar, it loads fine (and the Network tab confirms it’s loading from the service worker).

Weird!

(And now I will stop!)

ajgreenb:
Thanks for digging up those clues. I'll see if I'm missing some edge case or configuration option from `OfflinePlugin`. I don't think the changes introduced in this PR takes us very far off the "happy path", so hopefully the answer won't be too hard to uncover. Planning to take another whack at this tomorrow!

ajgreenb:
Damn, I'm stumped—I just can't seem to reproduce that error! It appears to always work for me. I loaded a fresh Popcode environment, waited for the console to "stabilize", then turned off my computer's WiFi and triggered a validation by typing in the HTML editor. `linters.js` loaded and linting appeared to work correctly.

I was able to reproduce the overall non-working service worker issue. I think there must be a problem with the current OfflinePlugin configuration. I'll tinker with that and see if I can get a better result.

ajgreenb:
Oddly, though, I get the overall non-working service worker on `master` too.

ajgreenb:
Okay, just pushed a change that reverts the configuration changes to OfflinePlugin. On my local setup, it still caches `linters.js` right out of the gate, so maybe we don't need to get fancy with the configuration after all. Can you try again and see if that happens for you?

outoftime:
Hey @ajgreenb—I’m so sorry, after poking around with this a bunch I realized that I had left the “Bypass for Network” setting on for the service worker in Chrome dev tools 😳 

With the service worker operating normally everything in this branch works fine, including when I revert the last commit. Do you mind just reverting that or rebasing it out (having `linters` in additional seems correct to me) and resolving the merge conflicts, and we will OFFICIALLY BE READY TO MERGE THIS???

ajgreenb:
Done and done! Thanks for your patience on this PR @outoftime 🙇 Let me know if you can think of anything else that should be done before merge.

PR #1141, opened 2017-10-19
Added .DS_Store to .gitignore file.

iamnickvolpe:
Small thing I noticed and a way to dip my foot in the water here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 npm-debug.log
 bower_components
 yarn-error.log
+.DS_Store


outoftime:
@iamnickvolpe thanks for this! As a rule, Popcode doesn’t include operating system or editor artifacts in the project `.gitignore` file. I recommend setting up a [global `.gitignore`](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) so you’ll never have to worry about this again. Or, if you prefer, you can also set up [an excludes file that is local to both the project and your computer](https://help.github.com/articles/ignoring-files/#explicit-repository-excludes). Let me know if you have any questions!

Issue #1140, opened 2017-10-18
Header Bar Updates - Clearer Project Workflow


iamnickvolpe:
I have a few suggestions for a clearer project workflow an ultimately a more familiar experience when saving and working with projects. Also some visual design tweaks to ultimately make the header bar feel less overwhelming and more familiar.

![header bar updates - popcode-01](https://user-images.githubusercontent.com/3432105/31731835-531ca94a-b404-11e7-834a-668566e02837.png)

Flow and Layout Changes
- Add a distinct "title" that can be edited instead of using "Page title."
- This includes an icon to let users know it can be edited and the "inline-edit" experience.
- Add a "saved" indicator once the project has been named and is being worked on.
- Separate current project actions from new project actions (left and right side).
- Make "New Project" disabled in initial state and add a popover to let users know why it's disabled.
- Move "zoom button" into hamburger menu. (not a popular action)

Visual Design Changes
- Add "line" separators between logical groupings of actions
- Add "button" outline for buttons
- Lighten up the text and consider not using bold on everything.
- Remove the "popcode" text in favor of logo to make room for actions.
- I personally like the "chevron-style" dropdown icon. Just a personal taste thing.

PR #1142, opened 2017-10-19
Header Bar Style Updates

iamnickvolpe:
Various updates as part of [this issue](https://github.com/popcodeorg/popcode/issues/1140). It varies slightly from the mockups but required less rework.

### Style updates to add clarity to the header bar.
- Separated header bar buttons into 2 logical groups on the left and right side.
- Removed wordmark to make room for actions.
- Added borders to header bar buttons to make them feel more actionable.
- Added hover states and active states to header bar buttons to increase clarity.
- Decreased the font-size of header bar buttons to increase clarity.
- Made the neutral notification blue to add visual separation between it and the header bar.

### Style updates to make the the error transition less overwhelming.
- Removed "yellow" thinking state as loader makes this unnecessary 
- Added super subtle color transition between red and green to make the experience less abrupt.

### Logged Out State
screenshot 2017-10-19 00 23 47

### Logged In State
screenshot 2017-10-19 00 24 02

### Hover State
screenshot 2017-10-19 00 24 55

### Active State
screenshot 2017-10-19 00 25 26

### Neutral Notification
screenshot 2017-10-19 00 26 11

### Transition Updates (No yellow state)
zcqg3c1gdd

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -118,13 +118,11 @@ body {
   flex: 0 0 auto;
   font-size: 2vh;
   font-weight: bold;
-  align-items: center;
-  padding: 0 1em;
+  align-items: stretch;
   z-index: 2;
-}
-
-.top-bar_yellow {
-  background-color: var(--color-yellow);
+  min-width: 760px;
+  color: rgba(0, 0, 0, 0.8);
+  transition: background-color .2s;
 }
 
 .top-bar_red {
@@ -156,11 +154,17 @@ body {
   height: 2em;
   border-radius: 50%;
   vertical-align: middle;
-  line-height: 1em;
+  margin-right: .5em;
 }
 
-.top-bar__username {
-  margin-left: 0.5em;
+.top-bar__group {
+  display: flex;
+  flex: 0 0 auto;
+  width: 50%;
+}
+
+.top-bar__group--right {
+  justify-content: flex-end;
 }
 
 .top-bar__menu-button {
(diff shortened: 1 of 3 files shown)


src/css/application.css
```diff
@@ -118,13 +118,11 @@ body {
...
-  padding: 0 1em;
+  align-items: stretch;
   z-index: 2;
-}
-
-.top-bar_yellow {
-  background-color: var(--color-yellow);
+  min-width: 760px;
+  color: rgba(0, 0, 0, 0.8);
+  transition: background-color .2s;
```
outoftime: So, I like this, and I am down to get rid of the yellow, but I feel like the color in the middle of the transition is kind of gross-looking. Can we use an animation to make it animate by way of an intermediate color (maybe…YELLOW…?)

src/css/application.css
```diff
@@ -118,13 +118,11 @@ body {
...
   font-weight: bold;
-  align-items: center;
-  padding: 0 1em;
+  align-items: stretch;
   z-index: 2;
-}
-
-.top-bar_yellow {
-  background-color: var(--color-yellow);
+  min-width: 760px;
```
outoftime: What’s the thinking here?

src/css/application.css
```diff
@@ -156,11 +154,17 @@ body {
...
   border-radius: 50%;
   vertical-align: middle;
-  line-height: 1em;
+  margin-right: .5em;
+}
+
+.top-bar__group {
+  display: flex;
+  flex: 0 0 auto;
+  width: 50%;
```
outoftime: Why fix the width at 50%? Seems like we’re losing most of the benefit of having this be a flex item… why not just do a `1 0 auto`?

PR #1151, opened 2017-10-29
Auth channel

outoftime:
Consolidates all login/logout business logic into the `user` saga, where it belongs.

First, uses a Redux Saga channel to represent the login/logout state, which we can then just watch with a saga, which allows us to do the same thing we were doing with `rxjs`, but in a way that feels much more natural in this code base. Allows us to remove the code in the `Workspace` component (of all places) that monitored Firebase login/logout state and dispatched actions accordingly.

Also creates `LOG_IN` AND `LOG_OUT` actions that are consumed by the `user` saga, which makes the appropriate calls to the Firebase client and handles errors where appropriate. Bonus: this behavior is now under test!


outoftime:
@ericliu121187 this is something i’ve been meaning to do for awhile but i was particularly motivated by your PR—I think getting the different behaviors right for explicit login vs. implicit login will be much easier now. just should require some slightly fancy redux-saga footwork (but it’s the exact sort of thing that redux-saga is really good at)

Issue #1002, opened 2017-08-22
User Feedback


outoftime:
Submitter: Courtney LeGates
Email: courtney@scripted.org
Can you please make the HTML/CSS/JavaScript boxes smaller so that there is less of a chance of overlap with the first line of code?

iamnickvolpe:
I personally feel there's no better way of doing it than to move the content down so it never bumps into the label box. Yes, you take a hit on vertical space efficiency, but you gain ultimate clarity. This could be mitigated by making them a little smaller. In which case, they would probably look better over on the left.

What I like to do is check out how other products in the space handle a similar pattern. Here's what I found:

## Codepen
Bars take up their own space. Code never runs into them.
![screenshot 2017-11-13 17 37 34](https://user-images.githubusercontent.com/3432105/32753296-4e161c7e-c89a-11e7-952d-61eb8c51c8ba.png)

## JSFiddle
Code labels are slightly translucent.
![screenshot 2017-11-13 17 37 52](https://user-images.githubusercontent.com/3432105/32753324-65dd3c0c-c89a-11e7-97e5-a17da8a9c9e9.png)

## JSBin
Bars take up their own space. Code never runs into them.
![screenshot 2017-11-13 17 38 16](https://user-images.githubusercontent.com/3432105/32753342-775aca30-c89a-11e7-9618-4f85eabc65a5.png)

## Visual Studio Code
Code label is on its own tab. Takes up its own space.
<img width="497" alt="screenshot 2017-11-13 17 39 28" src="https://user-images.githubusercontent.com/3432105/32753370-8b24bb3e-c89a-11e7-9f67-5d5cd54b96c8.png">

PR #1152, opened 2017-10-30
editor labels hidden on focus

kaylee42:
This is in regards to Issue #1002 

Low-key I'm expecting disagreement with this solution but it was just so simple I couldn't not 😝 CSS-only solution, with small HTML reorg so the selectors would work. When I was messing with it I didn't like the label fully disappearing because it felt unclear how to close, so I left the chevron with a transparent background. 

Had to ignore the linter because the ace editor native classes don't use BEM. Theoretically I know I should be able to use `ignoreSelectors` in the rules but I couldn't get it to work in less time than I wrote this CSS so I let it go, but would appreciate a nudge in the right direction of how to get that working.

### DIFF
--- a/src/components/EditorContainer.jsx
+++ b/src/components/EditorContainer.jsx
@@ -20,6 +20,8 @@ function EditorContainer({children, language, source, style, onHide, onRef}) {
       ref={onRef}
       style={prefixAll(style)}
     >
+      {helpText}
+      {children}
       <div
         className="editors__label editors__label_expanded"
         onClick={onHide}
@@ -28,8 +30,6 @@ function EditorContainer({children, language, source, style, onHide, onRef}) {
         {' '}
         <span className="u__icon">&#xf078;</span>
       </div>
-      {helpText}
-      {children}
     </div>
   );
 }
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -359,6 +359,16 @@ body {
   z-index: 1;
 }
 
+/* postcss-bem-linter: ignore */
+.ace_focus ~ .editors__label {
+  visibility: hidden;
+
+  /* postcss-bem-linter: ignore */
+  & .u__icon {
+    visibility: visible;
+  }
+}
+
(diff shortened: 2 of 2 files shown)


outoftime:
Well!

@kaylee42 do you think, at bottom, the CSS-only solution is the best one? It definitely scores points for brevity…

I think my main concern is that it couples Popcode’s behavior to the DOM structure that the ACE editor uses, which should probably be an internal concern to the editor. In particular, this solution would silently stop working if ACE stopped using the `ace_focus` class for the outermost container element of the editor; I think it would be reasonable to ACE to make a change like that without considering it breaking or documenting it as such in a changelog/semver/etc.

So ultimately it feels like we’re relying on an ersatz “API” from ACE that isn’t part of its public interface.

WDYT?

iamnickvolpe:
Took some screenshots of how similar products handle this pattern. Might help with ideas here. https://github.com/popcodeorg/popcode/issues/1002#issuecomment-344085126

outoftime:
@kaylee42 so what are your thoughts on the above?

iamnickvolpe:
@kaylee42 @outoftime I took a stab at some styling tweaks that _might_ help with this issue.
https://github.com/popcodeorg/popcode/pull/1202

While I still think the **cleanest** method would be to make them block-level elements, It's not necessarily the **best** solution, as it would diminish a lot of precious real estate. IMO, It's one of those "what do you value more?" type of decisions that don't have a perfect solution.

So in the interest of incremental improvements, I was hoping these small CSS tweaks would at least help alleviate the issue without any serious layout changes. Let me know what you think!

outoftime:
@iamnickvolpe agree that your PR will help ease the pain with this a bit, but I still think we need this (or a full-width label) to fully solve the problem

outoftime:
This one has been dormant for awhile so I’m going to go ahead and close it—feel free to pick it up again if you feel moved to @kaylee42 !

PR #1181, opened 2017-11-12
Header Bar Style Updates

iamnickvolpe:
# Header Bar Style Updates
## Summary
This PR contains various styling updates to the header bar and surrounding UI. The goal of this update is to add clarity to the interface. 

- Rearranged buttons based on a general formula where the left side is pertaining to the project being viewed, and the right side is about going further, or doing more "advanced" things.
- Varied the styles of buttons based on their importance and behavior.
- Added hover states to all buttons.
- Modified the active states of buttons.
- Modified the styles of menus.
- Modified the style of snapshot notification.

### Initial State
screenshot 2017-11-12 02 11 11

### Logged in state
screenshot 2017-11-12 02 11 27

### Button Hover State
screenshot 2017-11-12 02 11 46

### Button Active State
screenshot 2017-11-12 02 12 08

### Libraries Menu Design
screenshot 2017-11-12 02 12 24

### Projects Menu Design
screenshot 2017-11-12 02 12 35

### Screenshots Notification Design
screenshot 2017-11-12 02 13 09


iamnickvolpe:
@outoftime This new PR addresses your comments about the previous one (the padding around the logo and the line-height of the header bar itself). 

One thing I inadvertently did though was do away with the "sizing hack." (https://github.com/popcodeorg/popcode/pull/1077) I think that hack was interesting but isn't really that great for accessibility and sort of unpredictable, making it difficult to reign in the new styling.

2 suggestions there: 1, making the "zoom" button a dropdown menu, allowing for more zoom levels, and 2, a "presentation mode" where the header bar is just hidden from view.

outoftime:
@iamnickvolpe agree that the zoom-busting behavior of the header is hacky and a bit unpredictable, but it’s also addressing a pretty important use case, so I don’t think we can do away with it without a viable replacement. Musing on the two options you present:

> making the "zoom" button a dropdown menu, allowing for more zoom levels

I think we could do this more intuitively by having separate “zoom in” and “zoom out” buttons next to each other in the header, but we’d also need to change their behavior to _also_ zoom the iframe (a quick spike in the dev tools suggests that this would work fine using `transform`). Perhaps that would be the desired behavior for zooming in general?

> a "presentation mode" where the header bar is just hidden from view

This is an intriguing idea. Would “presentation mode” also increase the zoom? How would one get the header back / exit presentation mode?

iamnickvolpe:
Yeah, totally agree about needing a viable replacement. To be honest, It wasn't until I was fixing my tests until I realized, "Wait, I wonder if the sizing was set like that for a reason." And sure enough, I found that old PR. :)

I'll tell you what. Let me adjust this to at least retain the current functionality, and then we'll work from there. It might not be quite as nice, but I can get something passable enough to get this in.

Then, I'd love to try messing with these two other potential features, because I think both could be pretty valuable regardless.

100% agree with the zoom buttons. A little plus and minus would be awesome. What about something like this?

screenshot 2017-11-12 18 51 23

As for "Presentation Mode" I was thinking that it wouldn't have to be anything more complicated than simply hiding the header bar. I feel like this wouldn't be a "saved" option and it would always show on load.

screenshot 2017-11-12 18 47 56

iamnickvolpe:
Er, I think that carat is backwards, but you get the gist.

iamnickvolpe:
OK, everything's in ems now. Even the shadows! So absolutely nothing changes on zoom. @outoftime let me know what you think of this one.

src/css/application.css
```diff
@@ -113,13 +113,15 @@ body {
 /** @define top-bar */
 
 .top-bar {
+  align-items: center;
   background-color: var(--color-green);
+  color: rgba(0, 0, 0, 0.75);
   display: flex;
   flex: 0 0 auto;
-  font-size: 2vh;
+  font-family: sans-serif;
```
outoftime: What’s the thinking here? I would expect this to be one of the project fonts, or nothing at all?

src/css/application.css
```diff
@@ -131,67 +133,87 @@ body {
...
 
 .top-bar__snapshot_in-progress {
-  color: var(--color-light-gray);
+  opacity: 0.5;
   cursor: progress;
 }
 
 .top-bar__avatar {
-  height: 2em;
-  border-radius: 50%;
```
outoftime: Hmm, honestly I think the circle looks quite a bit nicer…

src/css/application.css
```diff
@@ -131,67 +133,87 @@ body {
...
+  box-shadow: inset 0 0 0 0.08em rgba(0, 0, 0, 0.1);
 }
 
 .top-bar__menu-button_active {
-  background-color: white;
-  border-color: var(--color-low-contrast-gray);
+  box-shadow: inset 0 0.08em 0.2em 0 rgba(0, 0, 0, 0.25);
+}
+
+.top-bar__menu-button_projects .top-bar__menu {
```
outoftime: No nesting in BEM!!

src/css/application.css
```diff
@@ -131,67 +133,87 @@ body {
...
 
 .top-bar__menu {
-  border: 1px solid var(--color-low-contrast-gray);
-  box-shadow: 2px 2px 2px var(--color-low-contrast-gray);
-  left: -1px;
+  box-shadow: 0 0 20px rgba(0, 0, 0, 0.25);
+  right: 0;
   position: absolute;
-  top: 100%;
+  top: 3.4em;
```
outoftime: Where does this number come from? Does 100% not work?

src/css/application.css
```diff
@@ -527,7 +547,8 @@ body {
 }
 
 .notification-list__notification_notice {
-  background-color: var(--color-green);
+  background-color: #00b8ff;
```
outoftime: Same as in other PR, please put all colors in variables!

src/css/application.css
```diff
@@ -527,7 +547,8 @@ body {
...
+  color: #fff;
```
outoftime: This can just be `white` (no need for a custom variable)

src/css/application.css
iamnickvolpe: It was part of my original mockup. In general, it's a little wider and easier to read.

21slczmmuh

src/css/application.css
iamnickvolpe: I agree the circle is trendy and looks good, but 2 reasons I changed it. 1, Github uses squares and 2, it matches the other buttons and general style. Happy to go back to circle if you really feel strongly about it though.

src/css/application.css
iamnickvolpe: Should it be `menu-button--projects`? How would you do it?

src/css/application.css
iamnickvolpe: I wanted the menus to line up with the bar instead of the button itself. It felt a little crowded and out of place to me.

o5zp8up6h6

iamnickvolpe:
OK @outoftime Try this out.

I made the updates you requested. A couple of things to net out on though:

- I decided to try a .2em border-radius on all the buttons, menus, and the avatar. Let me know what you think. (see Slack)

- And while I also like circle avatars in general, I think I would prefer to match Github, especially being that their default avatars are pixel-y characters that our circles tend to cut off. 

screenshot 2017-11-13 23 56 18

- I still think sans-serif is a more readable font to use for interface elements, especially with bold text.

- Make sense what I'm saying about the menus? Posted a screenshot.

src/components/TopBar/createMenu.jsx
```diff
@@ -69,7 +70,9 @@ export default function createMenu({
       const menu = isOpen ?
         (
           <div
-            className="top-bar__menu"
+            className={classnames('top-bar__menu',
+              menuClass,
+            )}
```
outoftime: Not a huge deal but I feel like this could all go on one line?

src/css/application.css
```diff
@@ -113,13 +115,14 @@ body {
...
   background-color: var(--color-green);
+  color: rgba(0, 0, 0, 0.75);
   display: flex;
   flex: 0 0 auto;
-  font-size: 2vh;
+  font-size: 1.7vh;
   font-weight: bold;
-  align-items: center;
-  padding: 0 1em;
+  padding: 1vh;
```
outoftime: Shouldn’t this be `1vw 1vh`?

src/css/application.css
```diff
@@ -131,73 +134,107 @@ body {
   background-color: var(--color-red);
 }
 
-.top-bar__logo-container > svg {
-  height: 3em;
+.top-bar__logo-container {
+  margin: 0 0.4em 0 0.2em;
+  min-width: 1.6em;
+  transform: scale(1.8);
+  width: 1.6em;
```
outoftime: Does this need both a `width` and a `min-width`?

src/css/application.css
```diff
@@ -131,73 +134,107 @@ body {
   background-color: var(--color-red);
 }
 
-.top-bar__logo-container > svg {
-  height: 3em;
+.top-bar__logo-container {
+  margin: 0 0.4em 0 0.2em;
+  min-width: 1.6em;
+  transform: scale(1.8);
```
outoftime: Bringing in `scale` seems like a weird move here, particularly given it’s an SVG and we can size it however we want!

Seems like the main hurdle here is that we want Pop to overflow the top/bottom padding that the rest of the menu items are subject to. A couple of alternative ideas:

1. Make Pop the size we want it using the `width` property, then position it relative and add a negative `top` to offset the padding
2. Instead of giving padding to the overall top bar, just give it an explicit `height`; the use of `align-items: center` will still keep everything looking nice. And we can again just give Pop the size we want

I’m strongly partial to Option 2, because it solves another problem, which I will mention below!

src/css/application.css
```diff
@@ -131,73 +134,107 @@ body {
...
+  width: 1.6em;
 }
 
 .top-bar__wordmark-container {
   position: relative;
-  margin-left: -0.25em;
+  margin: 0 1em 0 0.5em;
+  min-width: 6.5em;
+  width: 6.5em;
```
outoftime: Wondering about the `min-width` and `width` together again?

src/css/application.css
outoftime: I’m into the way the rounded corners look.

src/css/application.css
```diff
@@ -131,73 +134,107 @@ body {
...
 .top-bar__snapshot_in-progress {
-  color: var(--color-light-gray);
+  opacity: 0.5;
   cursor: progress;
 }
 
 .top-bar__avatar {
-  height: 2em;
-  border-radius: 50%;
+  border-radius: 0.2em;
```
outoftime: Can we extract this radius into a var, since we’re using it in a bunch of places?

src/css/application.css
```diff
@@ -131,73 +134,107 @@ body {
...
 .top-bar__username {
   margin-left: 0.5em;
 }
 
 .top-bar__menu-button {
-  align-self: stretch;
-  border-color: transparent;
-  border-style: solid;
-  border-width: 0 1px;
+  align-self: center;
```
outoftime: We don’t need this do we? Since the flex container has `align-items: center`?

src/css/application.css
```diff
@@ -131,73 +134,107 @@ body {
...
 
 .top-bar__menu {
-  border: 1px solid var(--color-low-contrast-gray);
-  box-shadow: 2px 2px 2px var(--color-low-contrast-gray);
-  left: -1px;
+  border-radius: 0.2em;
+  box-shadow: 0 0 20px rgba(0, 0, 0, 0.25);
   position: absolute;
-  top: 100%;
+  top: 3.4em;
```
outoftime: Another advantage of using an explicit height for the top bar instead of padding is that we could go back to 100% height here. That’s very appealing because it means we don’t have the knowledge of “what is the outer height of the top bar” spread over several places in the stylesheet.

src/css/application.css
```diff
@@ -131,73 +134,107 @@ body {
...
 
+.top-bar__menu_library-picker {
+  left: 0;
+}
+
+.top-bar__menu_project-picker,
+.top-bar__menu_current-user,
+.top-bar__menu_hamburger {
+  right: 0;
+}
```
outoftime: I’m a big proponent of just creating blocks/elements that straightforwardly describe the style that’s being applied, rather than naming everything semantically. You’ll find some semantic block/element names in Popcode but mostly for top-level layout concerns that don’t really have any scope for reuse.

Non-semantic block/element names are nice to work with because then the view templates can just straightforwardly describe how they’d like to look.

So: how about `top-bar__menu_left` and `top-bar__menu_right`?

src/components/TopBar/createMenu.jsx
iamnickvolpe: Done.

src/css/application.css
iamnickvolpe: Actually it works just as well in ems. Changed to ems.

src/css/application.css
iamnickvolpe: Option 2 is wayyy better than what I was doing. Done.

src/css/application.css
iamnickvolpe: Done. Handled differently.

src/css/application.css
iamnickvolpe: Done. Handled differently.

src/css/application.css
iamnickvolpe: Done.

src/css/application.css
iamnickvolpe: Done.

src/css/application.css
iamnickvolpe: Still doesn't quite work at 100% if you want the menus to sit below the bar, not directly below the button. I'm not partial to the way this looks.

screenshot 2017-11-16 20 33 09

src/css/application.css
iamnickvolpe: Done.

iamnickvolpe:
- Merged latest
- Addressed reviews
- Added subtle shadow to top bar.

Latest design:
screenshot 2017-11-16 20 53 03

src/css/application.css
```diff
@@ -68,6 +68,9 @@
   --color-yellow: #ffd100;
   --color-red: #ff5e4f;
   --color-purple: #f0f;
+  --color-low-contrast-blue: #baecff;
+  --color-blue: #00b8ff;
+  --rounded-corners: 0.2em;
```
outoftime: How about `--size-rounded-corners` to maintain the naming convention established by `--color-` prefixed variables above?

src/css/application.css
```diff
@@ -133,72 +139,96 @@ body {
...
 
 .top-bar__menu {
-  border: 1px solid var(--color-low-contrast-gray);
-  box-shadow: 2px 2px 2px var(--color-low-contrast-gray);
-  left: -1px;
+  border-radius: var(--rounded-corners);
+  box-shadow: 0 0 20px rgba(0, 0, 0, 0.25);
   position: absolute;
-  top: 100%;
+  top: 3.4em;
```
outoftime: So it makes sense why this can’t be 100%—I had forgotten that it is a child of the button and not the `TopBar` directly— but I still don’t love duplicating the knowledge of how tall overall the top bar is going to be. Can we extract that information into variables and/or use a `calc` statement to make more explicit how we get to `3.4em` here?

iamnickvolpe:
@outoftime Made the changes you requested. I think my math is right there. It **does** seem to work. Let me know what you think.

PR #1184, opened 2017-11-12
README notes on how to kill processes

coryetzkorn:
After stopping my local server, port 3000 still appeared to be blocked. I use the same port for a bunch of other apps. Had a really tough time tracking down how to free up the port. This did the trick!


outoftime:
@coryetzkorn this is an interesting one… so the problem isn’t that the port itself was taken up, but rather that Popcode had installed a service worker that caused your browser to blindly load Popcode from the browser cache when you accessed `http://localhost:3000`.

I believe this should only happen if you haven’t yet started a different server process on that port—Popcode’s service worker is supposed to just use the local cache as a backup if the resource is unavailable over the network—but it’s still a confusing situation.

Ideally I’d like to make it a little bit smoother than having to actually dig into Chrome’s service worker cache and clear it out manually. Unfortunately I can’t say I have any _great_ ideas. A few not-so-great ideas:

* Change Popcode’s default port to something less common
* Don’t run a service worker in development by default, enable it with an environment variable
* Figure out some way to “expire” the service worker (probably just in development) after some short amount of time
* ?

What do you think?

coryetzkorn:
@outoftime makes sense!

Changing the port number would be an easy fix, but obviously won't resolve the underlying problem.

Maybe there's a way to setup a gulp command that removes the service worker? Where is the service worker currently being initialized?

Simply noting that a service worker is being used in the README would be an ok solution too.

outoftime:
@coryetzkorn the service worker runs in the browser, so I don’t think it’s feasible to clear it from the server side / command line.

Seems worth doing the port change to at least “fix” the issue for most practical purposes?

coryetzkorn:
@outoftime sure... i'll just replace 3000 with something like 9000. I can't seem to locate where the post is set?

outoftime:
@coryetzkorn right now we’re just using the default but I think you’d configure it [here](https://github.com/popcodeorg/popcode/blob/master/gulpfile.js#L132-L144)

outoftime:
This one’s been dormant for awhile so I’m going to go ahead and close it, but feel free to pick it back up whenever @coryetzkorn !

PR #1186, opened 2017-11-13
Error message for lists (ul, ol, dl) with text children #1178

### LINKED ISSUE #1178: Should be an error message if you write <ul> or <ol> without nested <li>
This would fix a misconception that I have seen students have because of the permissiveness of HTML. Example:

```
<ul>First item</ul>
<ul>Second item</ul>
```

I expect an error message in this situation saying "All lists must contain list items wrapped with `<li>`"

wylieconlon:
Closes #1178 

This is a common student mistake I see, which is that students forget to write `<li>` inside their lists, and therefore don't actually get what they expect. I've added an error message for this use case, because for the purposes of teaching this is the only valid syntax we should expect.

Example of the error message:
screenshot 2017-11-12 20 58 47

### DIFF
--- a/locales/en/translation.json
+++ b/locales/en/translation.json
@@ -80,7 +80,7 @@
         "width": "Don't use the `width` attribute. Instead, use the CSS `width` property"
       },
       "lower-case-attribute-name": "Use lower case for attribute names",
-      "invalid-list-children": "The `<{{-tag}}>` tag must contain `<{{-children}}>`",
+      "text-elements-as-list-children": "The text `{{-textContent}}` inside the `<{{-tag}}>` tag must be surrounded by `<{{-children}}>` tags",
       "invalid-tag-location": "The `<{{-tag}}>` tag can’t go inside the `<{{-parent}}>` tag.",
       "invalid-tag-parent": "The `<{{-tag}}>` tag needs to be inside the `{{-parent}}`",
       "attribute-quotes": "You need to put the value of the `{{-attribute}}` attribute in quotation marks\nTry: `{{-attribute}}=\"myvalue\"`",
--- a/locales/en/translation.json
+++ b/locales/en/translation.json
@@ -80,6 +80,7 @@
         "width": "Don't use the `width` attribute. Instead, use the CSS `width` property"
       },
       "lower-case-attribute-name": "Use lower case for attribute names",
+      "invalid-list-children": "The `<{{-tag}}>` tag must contain `<{{-children}}>`",
       "invalid-tag-location": "The `<{{-tag}}>` tag can’t go inside the `<{{-parent}}>` tag.",
       "invalid-tag-parent": "The `<{{-tag}}>` tag needs to be inside the `{{-parent}}`",
       "attribute-quotes": "You need to put the value of the `{{-attribute}}` attribute in quotation marks\nTry: `{{-attribute}}=\"myvalue\"`",
--- a/src/validations/html/htmlInspector.js
+++ b/src/validations/html/htmlInspector.js
@@ -1,5 +1,6 @@
 import last from 'lodash/last';
 import isNull from 'lodash/isNull';
+import trim from 'lodash/trim';
 import {localizedArrayToSentence} from '../../util/arrayToSentence';
 import importLinters from '../importLinters';
 import Validator from '../Validator';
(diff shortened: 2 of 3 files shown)


src/validations/html/htmlInspector.js
```diff
@@ -38,8 +39,37 @@ const errorMap = {
...
+      },
+    };
+  },
 };
 
+function invalidListChildrenValidator(listener, reporter) {
+  listener.on('element', (elementName, domElement) => {
+    if (domElement.childNodes.length &&
+        (elementName === 'ul' || elementName === 'ol')) {
+      // Iterate over direct descendents of lists
```
outoftime: I generally like to avoid inline comments like this. The general principle is that, if the line-by-line working of the code requires further explanation, that suggests an opportunity to improve self-documentation by giving variables, functions, and classes more descriptive names (and possibly factoring out additional functions and variables whose main purpose is to self-describe). In this case, I think your code is perfectly clear without the comments here and a few lines below.

src/validations/html/htmlInspector.js
```diff
@@ -38,8 +39,37 @@ const errorMap = {
...
+      for (let node = domElement.firstChild; node; node = node.nextSibling) {
```
outoftime: Is there an advantage to this style of iteration rather than the more straightforward `for (const node of domElement.childNodes)`?

src/validations/html/htmlInspector.js
```diff
@@ -38,8 +39,37 @@ const errorMap = {
...
+        children: 'li',
+      },
+    };
+  },
 };
 
+function invalidListChildrenValidator(listener, reporter) {
+  listener.on('element', (elementName, domElement) => {
+    if (domElement.childNodes.length &&
+        (elementName === 'ul' || elementName === 'ol')) {
```
outoftime: Project style for multiline if statements would be:

```js
if (
  domElement.childNodes.length &&
    (elementName === 'ul' || elementName === 'ol')
) {
```

outoftime: (I _really_ wish eslint had better support for enforcing consistency with that sort of thing)

src/validations/html/htmlInspector.js
```diff
@@ -52,10 +82,14 @@ class HtmlInspectorValidator extends Validator {
     }
 
     const {HTMLInspector} = await importLinters();
+
+    HTMLInspector.rules.add('validate-list-children',
+      invalidListChildrenValidator);
```
outoftime: Similarly, project style for multi-line function calls:

```js
HTMLInspector.rules.add(
  'validate-list-children',
  invalidListChildrenValidator,
);
```

src/validations/html/htmlInspector.js
```diff
@@ -38,8 +39,37 @@ const errorMap = {
...
+      reason: 'invalid-list-children',
+      payload: {
+        tag,
+        children: 'li',
+      },
+    };
+  },
 };
 
+function invalidListChildrenValidator(listener, reporter) {
```
outoftime: I would name this something a little more specific, e.g. `noListsWithTextChildrenValidator`

src/validations/html/htmlInspector.js
outoftime: In other news, is it worth including `<dl>`s in here as well?

outoftime:
@wylieconlon would love to get this merged—think you’ll have a chance to go through the feedback soon?

src/validations/html/htmlInspector.js
wylieconlon: Done

src/validations/html/htmlInspector.js
wylieconlon: It's weird that ESLint seems to support that style of spacing with functions, but not conditionals or loops.

I can modify the code to support an error about required children for the `<dl>`, since `<dl>` contains `<dt>` or `<dd>`- what do you think?

Were you thinking of the `<menu>` tag, which can also contain `<li>` elements?

src/validations/html/htmlInspector.js
wylieconlon: Done

locales/en/translation.json
```diff
@@ -80,6 +80,7 @@
         "width": "Don't use the `width` attribute. Instead, use the CSS `width` property"
       },
       "lower-case-attribute-name": "Use lower case for attribute names",
+      "invalid-list-children": "The `<{{-tag}}>` tag must contain `<{{-children}}>`",
```
wylieconlon: @outoftime I think this error message is a bit confusing if you have written some valid `<li>` tags, but then wrote some text outside of any tag. I saw a student do this last week for example. Suggestions on wording? Maybe something like

`All text inside the <{{-tag}}> must be wrapped by <{{-children}}>`

outoftime: Definitely agree with your line of thinking, not sure how recognizable the word “wrapped” will be in this context though. Maybe “surrounded” instead?

src/validations/html/htmlInspector.js
outoftime: I hadn’t given any thought to the `<menu>` tag—a quick glance at MDN suggests that the rules for that one are considerably more complicated, but in the general case you can put more or less whatever you want in there, so I wouldn’t worry about it too much.

As for `<dl>`, seems like a nice-to-have, but my main observation was just that `<dl>`, like `<ul>` and `<ol>`, doesn’t permit text node children so might be an appropriate addition (even though the children it *does* permit are different)

outoftime:
@wylieconlon just responded to your notes, pretty much good to merge as-is unless you want to make any last changes based on those discussions. let me know!

locales/en/translation.json
wylieconlon: I've made another update to the wording, in the latest it will be `The text asdf inside the <ul> tag must be surrounded by <li> tags`

outoftime:
@wylieconlon OK, good to go except for a lint error that’s failing the build!

wylieconlon:
@outoftime This is ready to merge! Happy Thanksgiving!

PR #1187, opened 2017-11-13
Change color or labels and preview bar

iamnickvolpe:
This PR subtly changes the color of the code labels so that it's clearer to differentiate them between the background color of the "current line" in the code. Also updated the "preview bar" to match for for general consistency.

color-update

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -341,7 +341,7 @@ body {
 }
 
 .editors__label {
-  background-color: var(--color-low-contrast-gray);
+  background-color: #e2e2e2;
   color: var(--color-dark-gray);
   cursor: pointer;
   font-family: 'Roboto';
@@ -443,7 +443,7 @@ body {
   font-weight: bold;
   line-height: 1.5em;
   vertical-align: middle;
-  background-color: var(--color-low-contrast-gray);
+  background-color: #e2e2e2;
   z-index: 2;
 }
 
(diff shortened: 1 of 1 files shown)


src/css/application.css
```diff
@@ -341,7 +341,7 @@ body {
 }
 
 .editors__label {
-  background-color: var(--color-low-contrast-gray);
+  background-color: #e2e2e2;
```
outoftime: Project style is to keep all CSS colors in a variable—you won’t find color literals elsewhere in CSS property definitions (I hope!).

Can we extract this into a variable?

src/css/application.css
```diff
@@ -341,7 +342,7 @@ body {
 }
 
 .editors__label {
-  background-color: var(--color-low-contrast-gray);
+  background-color: var(--color-medium-contrast-gray);
```
outoftime: @iamnickvolpe sorry to go back and forth so much on a two-line pull request! But I don’t think “medium contrast gray” is doing the descriptive work we need here. The idea behind “low contrast gray” is specifically to call out “you shouldn’t use this in a situation where it _needs_ to contrast with white for all users”. Otherwise I would propose putting a new gray color on the general spectrum of “light gray”, “gray”, “dark gray”. Where does this new gray fall on that spectrum? Can we give any hints about when it’s appropriate to use this gray instead of the four other ones available?

Issue #312, opened 2016-10-22
JavaScript console


outoftime:
Add a built-in JavaScript console for the Popcode environment. JavaScript statements should be evaluated in the preview frame environment.

Obviously, we should give the same friendly feedback for erroneous console statements as we do for erroneous statements in the project code itself.

One question: what to do when the environment refreshes (e.g. because code is edited)? Keep the existing statements visible, but add some visual indicator that they’re out-of-date? (Maybe everything just gets grayed out?)

**One important thing:** A huge problem with JS Bin’s design is that many students conflate the console and the actual page, since they are very similar in design and hierarchy. We should make the console look very much distinct from the page preview; if anything, it should be visually coupled with the editors, rather than the output.

[react-console-component](https://www.npmjs.com/package/react-console-component) looks very promising as an out-of-box solution for the terminal interface itself. [jq-console](https://github.com/replit/jq-console) from our friends over at Repl.it also may be worth looking at, more for inspiration than as a library to integrate (since we don’t use jQuery).

Worth thinking about whether the console should be fully integrated into the React-Redux lifecycle or if we’re OK with the console interface itself being somewhat self-contained, with one-way data flow into the rest of the application (sort of like a very fancy version of a form field).

### Release roadmap

- [x] Bare-bones console (experimental mode only): type an expression into the console; it’s evaluated in preview environment and result is printed back to console (probably just in JSON at this point)
- [x] Handle pressing enter in the middle of the line
- [x] Properly capture `console.log` in student code and echo it back to console
- [x] Polish UI
- [x] Deal with scrolling properly when entries overflow available space: console should stay the same size, with content overflowing the top boundary and the last entry always fixed to the bottom. internal scrolling should of course be possible
- [x] Robust [pretty-printing](https://github.com/cvadillo/js-object-pretty-print) of result values (including e.g. function objects)
- [x] Handle up/down arrows
- [x] Clear button / handle control-L
- [x] Runtime errors in console statements should be mapped to user-friendly error messages like they are if the error comes from project code
- [x] Give good feedback when the preview reloads (e.g. because the student has made a change to the project source)—previous console interaction should have a clear “disabled” look to make it clear that they no longer apply to the current page
- [x] When closed, console tab should display an indicator/badge of unread messages (from `console.log`)
- [x] Runtime errors caused by project code should appear in the console (perhaps this should completely replace the current UI for displaying runtime errors?)
[...]

PR #1201, opened 2017-11-19
MVP JavaScript Console

outoftime:
Introduces a bare-bones JavaScript console, not ready for general release but enough to start playing with in experimental mode (`https://popcode.org?experimental`) and iterating on.

Specifically, support the following workflow:

* Type an expression into the console, press enter
* Expression is evaluated in the current preview page environment
* Result of evaluation is printed in the console

The is the first step in the project described by #312; see that issue description for a full list of steps (and to get a better sense of what this PR is *not* yet trying to do).

Here is the console in action:

console


jwang1919:
Great stuff @outoftime!

Can't comment much about the code because I honestly don't understand it well enough to form an expert opinion on it, but I can comment on the functionality.

At it's current incarnation, I think it is almost good enough to be usable in class. I think we can start trying to integrate it into some lessons after a clear button/command gets implemented.

Also found some strange bugs when attempting to type in an object.
Typing `{` and enter, causes Popcode to freeze
Typing `{}` and enter, doesn't return any feedback
Typing `}` and enter, returns expected result.

src/util/generatePreview.js
```diff
@@ -80,6 +118,7 @@ export class PreviewGenerator {
     this._addCss();
     if (options.propagateErrorsToParent) {
       this._addErrorHandling();
+      this._addMessageHandling();
```
outoftime: Should be its own option

src/components/Preview.jsx
```diff
@@ -15,10 +19,17 @@ export default function Preview({
...
   }
 
-  const projectFrames = compiledProjects.map(({source, timestamp}) => (
+  const projectFrames = compiledProjects.map(({source, timestamp}, key) => (
     <PreviewFrame
+      consoleEntries={
+        key === compiledProjects.keySeq().last() ?
+          consoleEntries :
+          new OrderedMap()
+      }
```
outoftime: Is there a less clunky way to do this?

src/components/ConsoleEntry.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {ConsoleEntry as ConsoleEntryRecord} from '../records';
+
+export default function ConsoleEntry({
+  entry: {expression, status, value, error},
+}) {
+  let output;
+
+  if (status === 'evaluated' && value !== undefined) {
```
outoftime: Do we need `status`?

src/components/ConsoleEntry.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {ConsoleEntry as ConsoleEntryRecord} from '../records';
+
+export default function ConsoleEntry({
+  entry: {expression, status, value, error},
+}) {
+  let output;
```
outoftime: `ConsoleOutput` component

src/components/Console.jsx
```diff
@@ -0,0 +1,100 @@
...
+  return session;
+}
+
+export default class Console extends Component {
+  constructor() {
+    super();
+    bindAll(this, '_handleInput');
+  }
+
+  _handleInput(containerElement) {
```
outoftime: Decompose into a `<ConsoleInput>` component

Issue #1200, opened 2017-11-19
Use semantic HTML5 elements where applicable


outoftime:
Layout is currently done pretty much exclusively using `<div>`s—let’s use `<header>`, `<main>`, etc. where appropriate.

outoftime:
Closed by #1211

PR #1211, opened 2017-11-20
Html5 semantics

CodeWritingCow:
Hello, I rename two `<div>`s as `<header>` and `<main>`. Issue #1200

### DIFF
--- a/src/components/Workspace.jsx
+++ b/src/components/Workspace.jsx
@@ -250,7 +250,7 @@ class Workspace extends React.Component {
     }
 
     return (
-      <main className="environment">
+      <div className="environment">
         <EditorsColumn
           currentProject={currentProject}
           editorsFlex={editorsFlex}
@@ -275,7 +275,7 @@ class Workspace extends React.Component {
           />
         </DraggableCore>
         {this._renderOutput()}
-      </main>
+      </div>
     );
   }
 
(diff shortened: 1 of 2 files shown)


src/components/Workspace.jsx
```diff
@@ -250,7 +250,7 @@ class Workspace extends React.Component {
     }
 
     return (
-      <div className="environment">
+      <main className="environment">
```
outoftime: I think the `<div>` with `className="layout"` below may be the most appropriate `<main>` element? LMK your thoughts

CodeWritingCow: Hi @outoftime: If we make `<div className="layout">` the `<main>`, the `<header>` would end up inside `<main>`.  

Since Popcode is a single-page site, this is fine. But if the site expands into multiple pages, it may be better to separate `<header>` from `<main>`. Each page would probably share a similar `<header>` but have different `<main>` contents. 

Just something to consider in the future. For now, I can change the `<div className="layout">` to `<main>` as you suggested! LMK what you think.

outoftime: Hey—you’re totally right that `.layout` isn’t a good `<main>` element since it would contain the `<header>`.

Maybe `.layout__columns`, though? Then the `<header>` and `<main>` would be siblings, which seems appropriate?

LMK what you think!

CodeWritingCow: You're right! `.layout__columns` works much better as `<main>`. I'll make the changes now.

PR #1218, opened 2017-11-25
Google Classroom Share Button

pwjablonski:
First Pass at adding Google Classroom Share Button to Hamburger Menu.

I used [Google Classroom Share Button](https://developers.google.com/classroom/guides/sharebutton) as a starting point. The button out of the box doesn't look that appealing and is hard to manipulate. Instead, I open a new window using the redirect URL "https://classroom.google.com/u/0/share?url=" with the appended snapshot link. From there Google Classroom API handles the rest. Works both for teacher and student accounts to assign or submit projects.

A couple of points:
1. Had some trouble figuring out how to handle and notifications effectively
2. I have not written tests for this feature


src/sagas/ui.js
```diff
@@ -37,6 +39,27 @@ function* githubExport(
   }
 }
 
+function* classroomExport(
+  successAction,
+  failureAction,
+  notDisplayedAction,
+  displayedAction) {
```
outoftime: If I’m not mistaken, the idea behind having the `githubExport` take all these parameters is so that it can agnostically handle both gist and repo exports. `classroomExport`, as written, is single-purpose, so there’s no need to parameterize the actions like this.

That said, I think it’s worth taking a step back and looking at the three “export” flows:

* Gist
* GitHub Repo
* Classroom

These three workflows are basically all identical, with the only variance being in what exactly goes into generating the export URL (which is of course a pretty complicated question involving various internal logic and calls to external APIs, but still, just one question).

Right now the flow is simply duplicated three times in the code, with duplicate actions, reducer logic, etc. I am generally a fan of a “rule of three” when it comes to code duplication: once you have three distinct code paths basically all doing the same thing, it’s time to refactor. This would be the point where the rule of three is triggered (I think snapshots are sufficiently different from the others to not be eligible).

So, I would propose restructuring the redux state/flow around a generic concept of an “export”, with action payloads including an export type, either `gist`, `repo`, or `classroom`. If I’m not mistaken, the only parts of the code that will need to be sensitive to the export type are:

* The `clients` saga
* The notification component (and probably only in that the translation key used is derived from the export type)

What do you think? Obviously that does add significant extra work to getting this feature released, but ultimately I think this is the right time to do the refactor (happy to justify that belief over Slack) and I think you’ll find that the total footprint of this PR won’t change substantially (there will just be more changed LOC and fewer new ones). Let me know!

src/clients/googleClassroom.js
```diff
@@ -0,0 +1,9 @@
+const BASE_URL = 'https://classroom.google.com/u/0/share?url=';
+
+export function createShareToClassroomUrl(snapshotKey) {
+  const uri = document.createElement('a');
+  uri.setAttribute('href', '/');
```
outoftime: We shouldn’t use attributes here—I believe you just want `uri.href = '/'`

src/clients/googleClassroom.js
```diff
@@ -0,0 +1,9 @@
...
+  uri.search = `snapshot=${snapshotKey}`;
+  const classroomShareUrl = BASE_URL + uri.href;
```
outoftime: This won’t properly escape the URL being used in the query parameter—better to remove the `uri=` from the `BASE_URL`, and then do `classroomShareUrl = BASE_URL + qs.stringify({url: uri.href})` ([`qs` reference here](https://www.npmjs.com/package/qs); it’s already a dependency)

src/sagas/ui.js
pwjablonski: Hey @outoftime Yes this makes sense. I think it sounds like the right time to do it and ultimately more organized in the long run. Looking forward to implementing this restructuring and getting feedback.

pwjablonski:
Hey @outoftime time took my first try at restructuring the flow to a single export function. I used "exportProject" instead of "export" as not to confuse with the keyword export. 

Couple things I had questions / notes on:

1. I updated the the github-export.html to project-export.html instead of using separate files for each export type. Thoughts?
2. I created a separate projectExportError notification component. Is this necessary/ is there a better way to implement?
3. I used the value "classroom" to identify the exportType for sharing to google classrooms. Is this descriptive enough?
4. In the saga/client.js the `projectExportError()` function previously took `e` as an argument. When I encountered a problem passing both `e` and `exportType` to the function. I choose to only pass `exportType` instead since `e` did not seem to be used later on.
5. I did not touch any tests yet / I haven't regularly used a testing framework so there is a bit learning curve there for me. I would like to learn though!

outoftime:
Inline responses:

> "exportProject" instead of "export" as not to confuse with the keyword export

Very reasonable!

> I updated the the `github-export.html` to `project-export.html` instead of using separate files for each export type. Thoughts?

Good call!

> I created a separate projectExportError notification component. Is this necessary/ is there a better way to implement?

Can we just use the generic notification component?

> I used the value "classroom" to identify the `exportType` for sharing to google classrooms. Is this descriptive enough?

Works for me!

> In the saga/client.js the `projectExportError()` function previously took `e` as an argument. When I encountered a problem passing both `e` and `exportType` to the function. I choose to only pass `exportType` instead since `e` did not seem to be used later on.

Hmm, this is surprising. Did you pass an single object with `error` and `exportType` properties, or just two arguments? What was the problem you encountered?

> I did not touch any tests yet / I haven't regularly used a testing framework so there is a bit learning curve there for me. I would like to learn though!

Cool! Hopefully you’ll find them pretty approachable—generally the reducer and saga tests are reasonably expressive, but do let me know if you get stuck or need a pointer in the right direction.

src/clients/googleClassroom.js
```diff
@@ -0,0 +1,11 @@
+import qs from 'qs';
+
+const BASE_URL = 'https://classroom.google.com/u/0/share?';
+
+export function createShareToClassroomUrl(snapshotKey) {
+  const uri = document.createElement('a');
+  uri.href = '/';
+  uri.search = `snapshot=${snapshotKey}`;
+  const classroomShareUrl = BASE_URL + qs.stringify({url: uri.href});
+  return classroomShareUrl;
```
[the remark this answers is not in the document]
pwjablonski: Yeah - took out extra line.

src/components/TopBar/HamburgerMenu.jsx
```diff
@@ -13,26 +14,33 @@ const HamburgerMenu = createMenu({
 
   renderItems({
     isExperimental,
-    isGistExportInProgress,
+    isProjectExportInProgress,
```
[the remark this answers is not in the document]
pwjablonski: I created three separate selectors. One for `isGistExportInProgress`, `isClassroomExportInProgress` and `isRepoExportInProgress`. Is this what you had in mind?

src/components/TopBar/HamburgerMenu.jsx
```diff
@@ -13,26 +14,33 @@ const HamburgerMenu = createMenu({
...
     isUserAuthenticated,
-    onExportGist,
-    onExportRepo,
+    onExportProject,
```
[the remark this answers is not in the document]
pwjablonski: Okay - I went ahead a changed it. Interested in hearing more about the arguments for the different approaches.

src/reducers/clients.js
```diff
@@ -2,7 +2,7 @@ import Immutable from 'immutable';
 
 const defaultState = new Immutable.Map({
   firebase: new Immutable.Map({exportingSnapshot: false}),
-  gists: new Immutable.Map({lastExport: null}),
+  projects: new Immutable.Map({lastExport: null}),
```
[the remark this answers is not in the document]
pwjablonski: This one I was less sure about the implementation. A couple questions?

1. What should the defaultState look like? 

2. In your above example what does value for lastExport `15119982141321` represent? Is it the current project id? It doesn't seem as though this id is used anywhere in the preexisting export flow.

pwjablonski:
@outoftime Made some of the changes requested. Not quite sure for `isGistExportInProgress` if I creating three separate selectors is what you had in mind. Also looking for feedback on the client reducer and how to structure the default state. 

Re the problem that I encounter in the saga/client.js the projectExportError() function. When I passed both '`e` and `exportType` as arguments to `projectExportError()`. ex. `projectExportError(e, exportType)` The payload only contained the error and not the exportType. However, if I passed in a string as an argument instead of the error,  ex. `exportType('test', exportType)` the payload was in the correct format `{e: 'test', exportType: 'gist'}`

src/reducers/clients.js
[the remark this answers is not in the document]
outoftime: @pwjablonski answers inline!

> What should the defaultState look like?

I think _probably_ an empty `Immutable.Map`—generally as a rule we would want to explicitly initialize properties to null if we’re dealing with an object-like structure (a group of discrete heterogeneous properties, usually modeled in Immutable with a `Record`) vs. this which is more of a keyed collection (all collection values are the same kind of thing; the keys do not have a predefined structure except inasmuch as there are three kinds of exports the application actually does).

> In your above example what does value for lastExport 15119982141321 represent? Is it the current project id? It doesn't seem as though this id is used anywhere in the preexisting export flow.

Oof sorry that was just a goof on my part, ignore that… what you’ve got looks good : )

pwjablonski:
@outoftime made some more updates. The tests work, however, they are do not currently test each individual export type. Would be love to get feedback on how they should be structured.

src/sagas/clients.js
```diff
@@ -24,34 +25,31 @@ export function* createSnapshot() {
...
-export function* exportRepo() {
-  const state = yield select();
-  const project = getCurrentProject(state);
-  const user = state.get('user').toJS();
   try {
-    const {html_url} = yield call(createRepoFromProject, project, user);
-    yield put(repoExported(html_url));
+    if (exportType === 'gist') {
+      const {html_url: url} = yield call(createGistFromProject, project, user);
+      yield put(projectExported(url, exportType));
```
outoftime: This line is the same across all three logical branches—might as well just write it once after the `if`/`else`?

src/sagas/ui.js
```diff
@@ -37,35 +35,28 @@ function* githubExport(
...
+export function* exportProject() {
+  yield* projectExport(
+    'PROJECT_EXPORTED',
+    'PROJECT_EXPORT_ERROR',
+    projectExportNotDisplayed,
+    projectExportDisplayed,
   );
 }
 
+
```
outoftime: No need for the extra line of whitespace here

src/selectors/index.js
```diff
@@ -19,6 +19,8 @@ import isCurrentProjectSyntacticallyValid
   from './isCurrentProjectSyntacticallyValid';
 import isExperimental from './isExperimental';
 import isGistExportInProgress from './isGistExportInProgress';
+import isRepoExportInProgress from './isRepoExportInProgress';
+import isClassroomExportInProgress from './isClassroomExportInProgress';
```
outoftime: I could go either way on having three separate selectors but I think it’s basically good. Selectors that take arguments are a little more complicated to deal with anyway.

test/unit/reducers/clients.js
```diff
@@ -7,15 +7,15 @@ import {
...
+  exportProject,
+  projectExported,
+  projectExportError,
+  projectExportDisplayed,
+  projectExportNotDisplayed,
 } from '../../../src/actions/clients';
 import {clients as states} from '../../helpers/referenceStates';
 
-const error = new Error();
+// const error = new Error();
```
outoftime: Looks like we don’t need this anymore?

test/unit/sagas/clients.js
```diff
@@ -44,13 +44,14 @@ test('createSnapshot()', (t) => {
   });
 });
 
-test('exportGist()', (t) => {
+test('exportProject()', (t) => {
   const url = 'https://gist.github.com/abc123';
+  const exportType = 'gist';
```
outoftime: So for the most part, since generally all export types are treated the same, it’s totally fine/good to only test one export type. However in the case of the saga, the saga’s behavior is in fact quite different depending on the export type, so we should write tests for all three scenarios here.

src/sagas/clients.js
pwjablonski: Yeah I agree - I tried it at the end and there are linting errors
`error  'url' is assigned a value but never used`
not sure why this is

src/sagas/ui.js
pwjablonski: Yup!

src/selectors/index.js
pwjablonski: Yeah was going back and forth here. That sounds good to me!

test/unit/reducers/clients.js
pwjablonski: Deleted

test/unit/sagas/clients.js
pwjablonski: Yeah this makes sense. Will update the tests!

src/sagas/clients.js
outoftime: Ah, because [`const` and `let` are block-local](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let)! But you should be able to fix that by declaring `let url;` before the `if` statement.

pwjablonski: Okay cool! How would I do that with the object destructing? The below throws an error at `{html_url: url}` 

``` 
 let url;

  try {
    if (exportType === 'gist') {
      {html_url: url} = yield call(createGistFromProject, project, user);
    } else if (exportType === 'repo') {
      {html_url: url} = yield call(createRepoFromProject, project, user);
    } else if (exportType === 'classroom') {
      const snapshotKey = yield call(createProjectSnapshot, project);
      url = yield call(createShareToClassroomUrl, snapshotKey);
    }
    yield put(projectExported(url, exportType));
  } 
```

The below throws gives back the `url` as undefined.

``` 
 let url;

  try {
    if (exportType === 'gist') {
      let {html_url: url} = yield call(createGistFromProject, project, user);
    } else if (exportType === 'repo') {
      let {html_url: url} = yield call(createRepoFromProject, project, user);
    } else if (exportType === 'classroom') {
      const snapshotKey = yield call(createProjectSnapshot, project);
      let url = yield call(createShareToClassroomUrl, snapshotKey);
    }
    yield put(projectExported(url, exportType));
  } 
```

outoftime: Yeah it’s gross but you have to do:

```js
({html_url: url} = yield call(createGistFromProject, project, user));
```

As in just literally [put parentheses around the whole statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) 😿

pwjablonski:
```
  let url;

  try {
    if (exportType === 'gist') {
      const {html_url: html_url} = yield call(createGistFromProject, project, user);
      url = html_url;
    } else if (exportType === 'repo') {
      const {html_url: html_url} = yield call(createRepoFromProject, project, user);
      url = html_url;
    } else if (exportType === 'classroom') {
      const snapshotKey = yield call(createProjectSnapshot, project);
      url = yield call(createShareToClassroomUrl, snapshotKey);
    }
    yield put(projectExported(url, exportType));
}
```

pwjablonski: I didn't see above! Cool!

pwjablonski:
@outoftime made the changes as suggested. For the export tests, they seem to be a bit repetitive as they are currently implemented ( `with successful export` and `with error` are the same in each test).

locales/en/translation.json
```diff
@@ -38,7 +39,9 @@
     "click-to-copy": "Click here to copy it",
     "snapshot-export-error": "Something went wrong trying to create that snapshot. Please try again.",
     "snapshot-import-error": "Something went wrong trying to load that snapshot. Please try again.",
-    "snapshot-not-found": "That snapshot doesn’t seem to exist. Check the link and try again."
+    "snapshot-not-found": "That snapshot doesn’t seem to exist. Check the link and try again.",
+    "classroom-export-complete": "Your project is ready to share to google classrooms",
```
outoftime: Let’s use “Google Classroom” (capitalized, singular) for the copy

locales/en/translation.json
```diff
@@ -38,7 +39,9 @@
...
+    "classroom-export-error": "Something went wrong trying to share to google classroom. Please try again."
```
outoftime: Ibid. capitalization

outoftime:
BTW I definitely think we can live with the repetition that’s currently in the saga tests.

locales/en/translation.json
pwjablonski: Can do!

PR #1245, opened 2017-12-07
First pass at sub-bar redesign.

iamnickvolpe:
This is the first part of a story to add some more functionality to editor labels, which includes a notification about poorly-formatted code and the ability to fix it.

screenshot 2017-12-07 02 45 44

### DIFF
--- a/src/components/Preview.jsx
+++ b/src/components/Preview.jsx
@@ -28,16 +28,20 @@ export default function Preview({
 
   return (
     <div className="preview output__item">
-      <div className="preview__title-bar">
-        <span
-          className="preview__button preview__button_pop-out u__icon"
-          onClick={onPopOutProject}
-        >&#xf08e;</span>
-        {title}
-        <span
-          className="preview__button preview__button_reset u__icon"
-          onClick={onRefreshClick}
-        >&#xf021;</span>
+      <div className="preview__title-bar sub-bar">
+        <div className="sub-bar__left">
+          {title}
+        </div>
+        <div className="sub-bar__right">
+          <span
+            className="sub-bar__button"
+            onClick={onPopOutProject}
+          >Full screen</span>
+          <span
+            className="sub-bar__button"
+            onClick={onRefreshClick}
+          >Refresh</span>
+        </div>
       </div>
       {projectFrames}
     </div>
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -71,6 +71,7 @@
 :root {
   --color-low-contrast-gray: #eee;
   --color-chrome: #e2e2e2;
+  --color-border: #bbb;
   --color-light-gray: #aaa;
   --color-gray: #666;
   --color-dark-gray: #444;
@@ -82,6 +83,7 @@
   --color-blue: #00b8ff;
   --size-rounded-corners: 0.2em;
   --size-top-bar: 4em;
+  --size-sub-bar: 2em;
   --size-top-bar-menu-button: 2.8em;
   --box-shadow: 0 0 0.1em 0 rgba(0, 0, 0, 0.5);
   --font-size-menu: 1.7vh;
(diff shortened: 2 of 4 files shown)


src/components/Preview.jsx
```diff
@@ -28,16 +28,20 @@ export default function Preview({
...
-        >&#xf021;</span>
+      <div className="preview__title-bar sub-bar">
+        <div className="sub-bar__left">
+          {title}
+        </div>
+        <div className="sub-bar__right">
+          <span
+            className="sub-bar__button"
+            onClick={onPopOutProject}
+          >Full screen</span>
```
outoftime: I find the icons more usable here (the eye finds what it’s looking for more quickly if it doesn’t have to read the words). And I have never noticed students finding them unintuitive. Have you?

src/css/application.css
```diff
@@ -71,6 +71,7 @@
 :root {
   --color-low-contrast-gray: #eee;
   --color-chrome: #e2e2e2;
+  --color-border: #bbb;
```
outoftime: _do we really need another gray tone?_

src/css/application.css
```diff
@@ -292,6 +295,57 @@ body {
   flex: 1 0 0;
 }
 
+/** @define sub-bars */
+
+.sub-bar {
+  box-shadow: var(--box-shadow);
+  font-size: var(--font-size-menu);
```
outoftime: I think we need a new name for this font size variable!

src/css/application.css
```diff
@@ -292,6 +295,57 @@ body {
...
+
+.sub-bar__right {
+  text-align: right;
+}
+
+.sub-bar__button {
+  border-left: 0.1em solid var(--color-border);
+  padding: 0 1em;
+  height: var(--size-sub-bar);
+  line-height: var(--size-sub-bar);
```
outoftime: Can we not use percentages here?

src/css/application.css
```diff
@@ -566,11 +583,16 @@ body {
 /** @define notification-list */
 
 .notification-list__notification {
-  margin: 0.2rem 0;
+  font-size: var(--font-size-menu);
```
outoftime: This seems unrelated to the other stuff in here?

outoftime (changes requested):
This may be a sub-bar redesign, but it’s certainly not a sub-par redesign……………

Anyway left a few comments, nothing major!

src/components/Preview.jsx
iamnickvolpe: Just looking for ways to increase overall clarity and discoverability. Anecdotally I see a lot of students don't know that those icons are even actions at all, let alone what they do. Would you be averse to icon + text?

src/css/application.css
iamnickvolpe: I'd say yes. I tried the existing ones but they looked ugly and too pronounced.

src/css/application.css
iamnickvolpe: Perhaps. It's becoming a general goto "base" font-size.

src/css/application.css
iamnickvolpe: Somewhat. Kinda rolling it into general "chrome/menu/etc." design tweaks to reign in the sizing and spacing of everything and make things look more standard. Would you be totally averse to keeping it in here?

iamnickvolpe:
Made some comments. Will shoot over changes soon. But any thoughts on how to fix those errors? I don't see how that's not valid BEM syntax.

outoftime:
@iamnickvolpe ah sorry I forgot to address that originally—you just have a typo in the `@define` statement, writing `sub-bars` rather than `sub-bar`

outoftime:
@iamnickvolpe any interest in picking this back up? would love to see these designs give a little bit of a “fresh new look” for the new school year!

Issue #1261, opened 2017-12-09
Make it visually clear that previous console expressions are defunct after preview refreshes


outoftime:
If I:

1. Type an expression into the console (particularly one with side effects)
2. Modify the project code (or press the refresh button)

It should be visually clear that the expressions in my console history are no longer in effect. Probably something pretty simple like graying out the previous entries (though we should be mindful of accessibility when choosing a gray tone)

outoftime:
FIxed by #1268

PR #1268, opened 2017-12-10
Make console entries inactive on refresh

pwjablonski:
Hey @outoftime took first steps toward #1261. Currently works to grey all previous entries but only on preview refresh. 

A couple of questions
1. What are actions for which we will want to grey the defunct expressions? 
2. Is this the right grey to use?


src/reducers/console.js
```diff
@@ -26,7 +26,11 @@ export default function console(stateIn, {type, payload, meta}) {
     case 'EVALUATE_CONSOLE_ENTRY':
       return state.set(
         meta.key,
-        new ConsoleEntry({expression: payload}),
+        new ConsoleEntry({expression: payload, status: 'active'}),
+      );
+    case 'DEACTIVATE_CONSOLE_ENTRIES':
```
outoftime: Instead of explicitly deactivating console entries, I think we can calculate whether a given entry should be deactivated based on two pieces of information:

* The key of the current compiled project
* The key of the compiled project that evaluated the entry

With these two pieces of information, we can just built logic right into the presentational components that says: “a console entry is disabled if it was evaluated by a compiled project version that is not the current (tail) compiled project”.

This would mean introducing no new actions at all, and we would no longer need to worry about the question of “what are actions for which we will want to grey the defunct expressions?”

We would of course need to update the `CONSOLE_VALUE_PRODUCED` and `CONSOLE_ERROR_PRODUCED` actions to include the compiled project key, and the reducer to include that information in the entry record. This would have the added advantage of giving us an explicit way to know which entries have been evaluated.

src/css/application.css
```diff
@@ -471,19 +471,29 @@ body {
...
 }
 
-.console__entry,
+.console__entry-active,
 .console__input {
   margin-bottom: 0.5em;
 }
 
-.console__entry {
+.console__entry-active {
```
outoftime: In BEM, we should always have a base block or element (in this case `console__entry`) alongside any necessary modifiers (in this case `console__entry_active`). This would prove practically useful as well as there would no longer be a need to duplicate the 4 pixel margin, for instance.

I would also lean toward just having a single `_inactive` modifier, with the base element definition having the active look.

Also this particular flavor of BEM uses a single underscore, rather than a dash, for modifiers.

src/css/application.css
```diff
@@ -471,19 +471,29 @@ body {
...
 
-.console__error {
+.console__error-active {
   color: var(--color-red);
 }
 
+.console__entry-inactive {
+  margin-bottom: 0.5em;
+  margin-left: 4px;
+  color: var(--color-low-contrast-gray);
```
outoftime: I think low-contrast gray is the one we definitely don’t want to use, since the contrast will too low for accessibility.

No strong opinions otherwise, let’s use whatever gray looks best and also passes e.g. [this](https://webaim.org/resources/contrastchecker/)

src/actions/console.js
```diff
@@ -4,20 +4,16 @@ import uuid from 'uuid/v4';
 
 export const consoleValueProduced = createAction(
   'CONSOLE_VALUE_PRODUCED',
-  (key, value) => ({key, value}),
+  (key, value, projectKey) => ({key, value, projectKey}),
```
pwjablonski: Should rename this

src/components/Console.jsx
```diff
@@ -17,11 +19,18 @@ export default function Console({
...
+
   const console = (
     <div className="console__repl output__item">
       {history.map((entry, key) => (
         // eslint-disable-next-line react/no-array-index-key
-        <ConsoleEntry entry={entry} key={key} />
+        <ConsoleEntry
+          compiledProjectKey={compiledProjectKey}
+          entry={entry}
+          key={key}
```
pwjablonski: This gave an error  `Do not use Array index in keys  react/no-array-index-key` though not when all on a single line

src/css/application.css
```diff
@@ -471,27 +471,23 @@ body {
...
   margin-left: 4px;
 }
 
-.console__error-active {
+.console__error {
   color: var(--color-red);
 }
 
-.console__entry-inactive {
+.console__entry_inactive * {
```
pwjablonski: I am trying to gray the whole entry including the red error message. This seems like not the right way to do it. Should I pass a prop down to console output and be more specific?

src/actions/console.js
outoftime: Maybe `currentCompiledProjectKey`?

src/components/Console.jsx
outoftime: You can use an `eslint-disable-next-line` for that—unfortunately since JS is dynamically typed there’s no way to statically tell the difference between an array iteration and a keyed collection iteration, so eslint produces some false positives.

src/css/application.css
```diff
@@ -484,6 +484,12 @@ body {
   color: var(--color-red);
 }
 
+.console__entry_inactive * {
```
outoftime: Your instinct to add a prop to `ConsoleOutput` and then use more specific selectors is the right one—unfortunately one of the more annoying patterns required by BEM but overall worth it imo!

src/records/ConsoleEntry.js
```diff
@@ -4,4 +4,5 @@ export default Record({
   expression: '',
   value: null,
   error: null,
+  projectKey: '',
```
outoftime: I would probably call this `evaluatedByCompiledProjectKey` and set its default to `null`

src/records/ConsoleError.js
```diff
@@ -3,4 +3,5 @@ import {Record} from 'immutable';
 export default Record({
   name: '',
   message: '',
+  projectKey: '',
```
outoftime: Do we need this?

src/containers/Preview.js
```diff
@@ -27,12 +27,12 @@ function mapStateToProps(state) {
 
 function mapDispatchToProps(dispatch) {
   return {
-    onConsoleError(key, name, message) {
-      dispatch(consoleErrorProduced(key, name, message));
+    onConsoleError(key, name, message, projectKey) {
```
outoftime: Throughout this I would encourage you to use the language of `compiledProjectKey` rather than `projectKey`—they are keys for different entities!

outoftime: (Is there a better/less confusing name for the `CompiledProject` entity? Maybe `CompiledPage`?)

src/components/ConsoleOutput.jsx
```diff
@@ -3,7 +3,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {ConsoleEntry} from '../records';
 
-export default function ConsoleOutput({entry: {value, error}}) {
+export default function ConsoleOutput({entry}) {
+  const {value, error} = entry;
```
outoftime: Do we need `entry` in scope?

src/components/Console.jsx
```diff
@@ -17,11 +19,18 @@ export default function Console({
     return null;
   }
 
+  const mostRecentCompiledProject = compiledProjects.last();
+  const compiledProjectKey = get(mostRecentCompiledProject, 'timestamp', 0);
```
outoftime: Let’s push this down into a selector—the `Console` component shouldn’t need to know about the entire collection of compiled projects.

pwjablonski:
@outoftime I did not see the above comments. Please disregard last push.

pwjablonski:
@outoftime made updates on above changes. Does this naming look right for variables?

compiledProjectKey => this is passed into actions from the preview
evaluatedBycompiledProjectKey => once reduced in the entry record
currentcompiledProjectKey =>  this is selected from the list of compiled projects. it indicates the compiled project key(timestamp) of the current project and is updated once there is a new compiled project.

I also think its a bit unclear how it is call timestamp in the compiled project record, but we are calling it a compiledProjectKey here. Should this be updated to be more consistent?

I also think the conditional rendering in the console entry seems sloppy.  Thoughts on how to make that better.

src/components/Preview.jsx
```diff
@@ -20,6 +20,7 @@ export default function Preview({
 
   const projectFrames = compiledProjects.map(({source, timestamp}, key) => (
     <PreviewFrame
+      compiledProjectKey={timestamp}
```
outoftime: I agree that the inconsistency of “timestamp” vs. “key” is grating. In fact it does not matter that it’s a timestamp—that is just of a detail of how it is generated. Let’s rename it to `compiledProjectKey` in the record itself.

src/components/PreviewFrame.jsx
```diff
@@ -141,6 +150,7 @@ class PreviewFrame extends React.Component {
 }
 
 PreviewFrame.propTypes = {
+  compiledProjectKey: PropTypes.number.isRequired,
```
outoftime: Don’t feel super-strongly about this, but I think we should either:

* Pass the entire `CompiledProject` record rather than doing `src` and `compiledProjectKey` piecemeal, or
* Not pass `compiledProjectKey` in here at all, and instead have the `<Preview>` be responsible for adding that information into the callback

src/components/ConsoleEntry.jsx
```diff
@@ -3,17 +3,28 @@ import PropTypes from 'prop-types';
...
+  const {expression, evaluatedByCompiledProjectKey} = entry;
+  const isActive = currentCompiledProjectKey === evaluatedByCompiledProjectKey;
 
   return (
     <div className="console__entry">
-      <div className="console__expression">{expression}</div>
-      <ConsoleOutput entry={entry} />
+      <div
+        className={
+          isActive ? 'console__expression' : 'console__expression_inactive'
```
outoftime: The `classnames` module gives you a really nice way to do this:

```
className={
  classnames('console__expression', {console__expression_inactive: !isActive}
}
```

Note that this also always gives it the `console__expression` class—a block or element should always have the base class, with modifier classes alongside it if needed.

src/components/ConsoleEntry.jsx
```diff
@@ -3,17 +3,28 @@ import PropTypes from 'prop-types';
...
+        }
+      >
+        {expression}
+      </div>
+      <ConsoleOutput
+        currentCompiledProjectKey={currentCompiledProjectKey}
```
outoftime: I would probably just pass `isActive` as a prop, no need to make the `<ConsoleOutput>` do the same logic again

src/components/ConsoleOutput.jsx
```diff
@@ -3,14 +3,21 @@ import React from 'react';
...
 
-export default function ConsoleOutput({entry: {value, error}}) {
+export default function ConsoleOutput({entry, currentCompiledProjectKey}) {
+  const {value, error, evaluatedByCompiledProjectKey} = entry;
+  const isActive = currentCompiledProjectKey === evaluatedByCompiledProjectKey;
+
   if (!isNil(value)) {
-    return <div className="console__value">=&gt; {value} </div>;
+    return (
+      <div className={isActive ? 'console__value' : 'console__value_inactive'}>
```
outoftime: Ibid. `classnames` and always having the base element class

src/components/ConsoleOutput.jsx
```diff
@@ -3,14 +3,21 @@ import React from 'react';
...
+        =&gt; {value}
+      </div>
+    );
   }
 
   if (!isNil(error)) {
     return (
-      <div className="console__error">
+      <div className={isActive ? 'console__error' : 'console__error_inactive'}>
```
outoftime: Ibid. `classnames` and always having the base element class

src/components/Console.jsx
```diff
@@ -20,8 +21,12 @@ export default function Console({
   const console = (
     <div className="console__repl output__item">
       {history.map((entry, key) => (
-        // eslint-disable-next-line react/no-array-index-key
-        <ConsoleEntry entry={entry} key={key} />
+        <ConsoleEntry
+          currentCompiledProjectKey={currentCompiledProjectKey}
```
outoftime: I don’t feel strongly about it but I think there’s probably an argument for just doing the logic here and passing an `isActive` prop into `<ConsoleEntry>`

src/selectors/getCurrentCompiledProjectKey.js
```diff
@@ -0,0 +1,9 @@
+import get from 'lodash/get';
+
+export default function getCurrentCompiledProjectKey(state) {
+  return get(
+    state.get('compiledProjects').last(),
+    'timestamp',
+    0,
```
outoftime: Do we want a default of 0 here? I think `undefined` is probably more correct

pwjablonski: No we do not :)

src/reducers/console.js
```diff
@@ -8,20 +8,33 @@ export default function console(stateIn, {type, payload, meta}) {
...
     state = initialState;
   }
-
+  //          input.set('projectKey', payload.projectKey);
   switch (type) {
     case 'CONSOLE_VALUE_PRODUCED':
       return state.update(
         payload.key,
         input => input.set('value', payload.value),
+      ).update(
```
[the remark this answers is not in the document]
pwjablonski: Will delete

src/reducers/console.js
```diff
@@ -8,20 +8,33 @@ export default function console(stateIn, {type, payload, meta}) {
   if (state === undefined) {
     state = initialState;
   }
-
+  //          input.set('projectKey', payload.projectKey);
```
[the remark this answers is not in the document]
pwjablonski: Yeah that's a mistake

src/css/application.css
```diff
@@ -484,6 +484,18 @@ body {
   color: var(--color-red);
 }
 
+.console__expression_inactive {
```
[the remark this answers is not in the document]
pwjablonski: Makes sense

src/components/PreviewFrame.jsx
pwjablonski: I see how I would go about the first option but not quite sure how to implement it the second way

src/components/Preview.jsx
pwjablonski: At what point should this happen?

If this is updated in the record should it also be updated in the action, reducer and compiledProjects saga as well

ie.
`const compiledProjectKey = Date.now();` instead of `const timestamp = Date.now();`

src/components/ConsoleEntry.jsx
pwjablonski: That makes sense

src/components/Console.jsx
pwjablonski: Yeah I was thinking that. It seems cleaner but wasn't sure if it was important that the logic was done within the element for which the evaluation was being used

src/components/PreviewFrame.jsx
outoftime: @pwjablonski so basically like this:

```jsx
  const projectFrames = compiledProjects.map(({source, timestamp: compiledProjectKey}, key) => (
    <PreviewFrame
      consoleEntries={consoleEntries}
      isActive={key === compiledProjects.keySeq().last()}
      key={timestamp}
      src={source}
      onConsoleError={(key, name, message) =>
        onConsoleError(key, name, message, compiledProjectKey)
      }
      onConsoleValue={(key, value) =>
        onConsoleValue(key, value, compiledProjectKey)
      }
      onRuntimeError={onRuntimeError}
    />
  ));
```

src/components/Preview.jsx
outoftime: Yeah I think we should do it as a separate PR, not much advantage in blocking this one.

src/components/Console.jsx
outoftime: There’s definitely no objective answer to that question—it’s just a matter of design:

* The way you’ve got it now makes the `<ConsoleEntry>` more self-contained; the calling component does not need to know anything about active vs. inactive entries. Useful if `<ConsoleEntry>` might be used in a bunch of different contexts, but always have the same logic for active vs. inactive
* The change I’m proposing makes the `<ConsoleEntry>` more flexible—it doesn’t need to know _why_ it might be inactive, it just does what it’s told. Useful if `<ConsoleEntry>` might be used in different contexts with _different_ logic for what makes it active vs. inactive

In this case neither of the practical scenarios seems likely, which is why I don’t feel particularly strongly about it : )

src/components/PreviewFrame.jsx
pwjablonski: Ahh cool! JsX doesn't seem to like arrow functions in the props though. I think I'll go about it the first way unless you have a preference?

outoftime: Ah, I think that might be an over-eager eslint rule… first way is good with me!

src/components/Console.jsx
```diff
@@ -62,5 +68,6 @@ Console.propTypes = {
 };
 
 Console.defaultProps = {
+  currentCompiledProjectKey: undefined,
```
outoftime: `null` and `undefined` are kind of a weird JavaScript idiosyncrasy and I don’t have a great mental model for when to use which, but generally in Popcode I try to avoid ever giving something an explicit value of `undefined`, preferring `null` for cases like this.

src/selectors/getCurrentCompiledProjectKey.js
```diff
@@ -0,0 +1,9 @@
+import get from 'lodash/get';
+
+export default function getCurrentCompiledProjectKey(state) {
+  return get(
+    state.get('compiledProjects').last(),
+    'compiledProjectKey',
+    undefined,
```
outoftime: You can just omit the last argument here and get the same behavior from `get()`

PR #1269, opened 2017-12-11
Validates source on project creation, closes #1257

### LINKED ISSUE #1257: Console expressions not evaluated in freshly loaded environment
If you load up Popcode in a fresh environment and, without making any changes to code, just type a console expression, nothing happens.

This is because we don’t actually render any preview until code is changed. So, there is no preview frame to evaluate the expression.

We should make sure to create an initial compiled project on bootstrap (probably just need to add this to the final case in the `applicationBootstrap()` saga defined under `sagas/projects`)

alessbell:
This PR closes [issue 1257](https://github.com/popcodeorg/popcode/issues/1257).

It's also my first Popcode PR 🎉 Validating the source on project creation made sense to me but feedback is welcome as I learn more about the code base :)

### DIFF
--- a/src/sagas/compiledProjects.js
+++ b/src/sagas/compiledProjects.js
@@ -4,6 +4,7 @@ import {
   put,
   select,
   throttle,
+  takeEvery,
 } from 'redux-saga/effects';
 
 import every from 'lodash/every';
@@ -32,5 +33,8 @@ export function* validatedSource() {
 }
 
 export default function* () {
-  yield all([throttle(100, 'VALIDATED_SOURCE', validatedSource)]);
+  yield all([
+    throttle(100, 'VALIDATED_SOURCE', validatedSource),
+    takeEvery('PROJECT_CREATED', validatedSource),
+  ]);
 }


src/sagas/projects.js
```diff
@@ -44,6 +47,7 @@ export function* applicationLoaded(action) {
 
 export function* createProject() {
   yield put(projectCreated(generateProjectKey()));
+  yield put(validatedSource('javascript', []));
```
outoftime: So this approach definitely works, but kind of as a second-order effect, because the `compiledProjects` saga [consumes this action and initiates compilation in response to it](https://github.com/alessbell/popcode/blob/dc49520b55c57487166fdeac44aaa1449b41240a/src/sagas/compiledProjects.js#L35).

My preference is to keep action dispatches as close as possible to an unvarnished record of what actually happened (somebody clicked a button, some async operation finished, etc.) In this case we are saying we’ve finished validating a source, but in fact we haven’t (this intuition is strengthened, I think, by the fact that we have to arbitrarily pick a language to say we have validated).

My proposal would be: instead of adding a new dispatch into the flow, simply make the system react more robustly to the actions that are already being dispatched. In particular, I think the `compiledProjects` saga should consume the `PROJECT_CREATED` action and compile the project in response.

LMK what you think!

alessbell: Yeah, makes sense! This solution indeed felt slightly wrong with the language hard-coded. 

The thinking was that, given that the source is a brand new Popcode, it can safely be assumed valid. I was after that second-order effect :) Much clearer to simply compile the project, updated.

src/sagas/compiledProjects.js
```diff
@@ -32,5 +33,8 @@ export function* validatedSource() {
 }
 
 export default function* () {
-  yield all([throttle(100, 'VALIDATED_SOURCE', validatedSource)]);
+  yield all([
+    throttle(100, 'VALIDATED_SOURCE', validatedSource),
+    takeEvery('APPLICATION_LOADED', validatedSource),
```
outoftime: This is fairly inside baseball, but I think we actually want to consume `PROJECT_CREATED` here—`APPLICATION_LOADED` will trigger different behavior in the projects saga depending on the initial application state (snapshot/gist ID in the URL, project rehydrated from local storage, etc.), and in some circumstances the saga will kick off a full validation. The case when it doesn’t do a full validation is exactly the case when it dispatches `PROJECT_CREATED`.

As a bonus, that’ll makes the console work right after you click “New Project”, which it currently does not—a closely related bug!.

alessbell: Oops! Right yes makes sense, tho didn't realize that re: "New Project"

Issue #1267, opened 2017-12-10
Invariant Violation in /


outoftime:
## Error in Popcode

**Invariant Violation** in **/**
Uncaught Invariant Violation: Minified React error #31; visit http://facebook.github.io/react/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7B0%2C%20length%2C%20prevObject%7D&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

[View on Bugsnag](https://app.bugsnag.com/popcode/popcode/errors/5a2c85534bef8c001be55c69?event_id=5a2c85534bef8c001be55c68&i=gh&m=ci)

## Stacktrace

    webpack:///node_modules/react-dom/cjs/react-dom.production.min.js:14 - E

[View full stacktrace](https://app.bugsnag.com/popcode/popcode/errors/5a2c85534bef8c001be55c69?event_id=5a2c85534bef8c001be55c68&i=gh&m=ci)

*Created automatically via Bugsnag*

outoftime:
Closed by #1273

Issue #1252, opened 2017-12-09
Console should pretty-print values


outoftime:
The JavaScript console should pretty-print values:

* Values should be syntax highlighted
* We should print all values in a reasonable way, including functions, complex objects (e.g. jQuery collection), etc.
* As a baseline, the printing behavior for a given object should be at least as good as it is in the current built-in Chrome console

I suspect [https://github.com/cvadillo/js-object-pretty-print](https://github.com/cvadillo/js-object-pretty-print) will be of use—I’m thinking we will probably want to actually load this into the preview environment and just pass an already-pretty-printed string through the `postMessage`.

alessbell:
Started looking into some tools that might help us here and came across [this post by LogRocket](https://blog.logrocket.com/recreating-the-chrome-console-in-react-123d6c4aa930) which led me to [`react-inspector`](https://github.com/xyc/react-inspector).

Looks promising, going to play around with it as soon as I can. I think the hardest part might actually be handling jQuery objects...

outoftime:
@alessbell react-inspector looks awesome! one potential hurdle—we won’t have access to React inside the preview frame, but some values (such as function objects) can’t be sent over the `postMessage` interface (as far as I know). any idea how we might bridge the divide?

outoftime:
So for a bare-bones implementation of pretty-printing I went with [`object-inspect`](https://github.com/substack/object-inspect) which has good default behavior in most cases. However, I think we might want to move to [`javascript-stringify`](https://github.com/blakeembrey/javascript-stringify), mainly because it allows us to define a custom visitor function; I think we’ll need to write custom behavior to really get this right.

We might just be able to skip that step and go straight to `react-inspector`, though—could use something like [`serialize-javascript`](https://github.com/yahoo/serialize-javascript) to send values across the frame boundary without losing functions/DOM/etc.

PR #1273, opened 2017-12-12
Stringifies console value in preview frame, closes #1254

### LINKED ISSUE #1254: Application crashes if you evaluate jQuery in the console
If you type, for instance `$("body")` into the console with jQuery enabled, the entire environment crashes.

I believe this is because we’re currently passing the value returned by console expressions as a JSON-serialized JavaScript value, and we keep it that way all the way through actually interpolating it into the JSX output. You can’t interpolate an object into JSX.

I think the right move here would just be to stringify the value in the preview frame, probably just using `JSON.stringify` to start. That would set us up to do the better version with #1252.

alessbell:
Closes #1254.

Followed the advice on the issue: prevent the app from crashing first by stringifying the value and worry about pretty printing in #1252. Still thought a bit about how we'll accomplish that and have some links I'll share on the latter issue...

screen shot 2017-12-12 at 5 51 07 pm

### DIFF
--- a/src/components/PreviewFrame.jsx
+++ b/src/components/PreviewFrame.jsx
@@ -124,7 +124,7 @@ class PreviewFrame extends React.Component {
   }
 
   _handleConsoleValueMessage({key, value}) {
-    this.props.onConsoleValue(key, value);
+    this.props.onConsoleValue(key, JSON.stringify(value));
   }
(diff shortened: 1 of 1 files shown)


outoftime:
Closes #1267

Issue #1249, opened 2017-12-09
Capture console.log in JavaScript code and propagate it to the console


outoftime:
If the regular JavaScript code contains a call to `console.log`, the message should appear in the Popcode JavaScript console. This means we will need to either replace the `console` object entirely in the preview environment, or overwrite the `log` method. For completeness, we might as well handle `debug`, `info`, `warn`, `error` as well.

We should make sure _not_ to break any other built-in console methods (e.g. `console.time()`).

I think it makes sense for calls to `console.log` and friends to still appear in the browser’s JavaScript console—so when we override the built-in functionality, we should hold references to the original methods and call them as well.

alexpelan:
Looking at this: https://developer.mozilla.org/en-US/docs/Web/API/console#Outputting_text_to_the_console

I think for initial scope, I'm going to ignore both 'Using string substitutions' and 'Styling console output' These will both gracefully degrade in the Popcode console, just displaying literaly `%s` and `%c`

outoftime:
Seems entirely reasonable!

PR #1281, opened 2017-12-16
Evaluate console.(log, debug, etc) and send to the Popcode console

alexpelan:
In the preview frame, we capture console.log and its cousins and send the output to the Popcode console, before deferring to the browser implementation.

For now we treat all of these functions exactly the same - console.log is the only one used in the ScriptEd curriculum as far as I know.

Fixes: #1249

### DIFF
--- a/src/components/PreviewFrame.jsx
+++ b/src/components/PreviewFrame.jsx
@@ -30,7 +30,7 @@ class PreviewFrame extends React.Component {
 
     if (this._channel && isActive) {
       for (const [key, {expression}] of newProps.consoleEntries) {
-        if (!previousConsoleEntries.has(key)) {
+        if (!previousConsoleEntries.has(key) && expression) {
           this._evaluateConsoleExpression(key, expression);
         }
       }
@@ -110,6 +110,10 @@ class PreviewFrame extends React.Component {
     });
   }
 
+  _handleConsoleLog(output) {
+    this.props.onConsoleLog(output);
+  }
+
   _attachToFrame(frame) {
     this._frame = frame;
 
@@ -131,6 +135,9 @@ class PreviewFrame extends React.Component {
 
     this._channel.bind('error', (_trans, params) => {
       this._handleErrorMessage(params);
+    });
+    this._channel.bind('log', (_trans, params) => {
+      this._handleConsoleLog(params.output);
     });
   }
 
@@ -155,6 +162,7 @@ PreviewFrame.propTypes = {
   consoleEntries: ImmutablePropTypes.iterable.isRequired,
   isActive: PropTypes.bool.isRequired,
   onConsoleError: PropTypes.func.isRequired,
+  onConsoleLog: PropTypes.func.isRequired,
   onConsoleValue: PropTypes.func.isRequired,
   onRuntimeError: PropTypes.func.isRequired,
 };
--- a/src/components/PreviewFrame.jsx
+++ b/src/components/PreviewFrame.jsx
@@ -111,7 +111,8 @@ class PreviewFrame extends React.Component {
   }
 
   _handleConsoleLog(output) {
-    this.props.onConsoleLog(output);
+    const {compiledProjectKey} = this.props.compiledProject;
+    this.props.onConsoleLog(output, compiledProjectKey);
   }
 
   _attachToFrame(frame) {
--- a/src/components/PreviewFrame.jsx
+++ b/src/components/PreviewFrame.jsx
@@ -110,7 +110,8 @@ class PreviewFrame extends React.Component {
     });
   }
 
-  _handleConsoleLog(output) {
+  _handleConsoleLog(consoleArgs) {
+    const output = consoleArgs.map(arg => JSON.stringify(arg)).join(' ');
     const {compiledProjectKey} = this.props.compiledProject;
     this.props.onConsoleLog(output, compiledProjectKey);
   }
@@ -138,7 +139,7 @@ class PreviewFrame extends React.Component {
       this._handleErrorMessage(params);
     });
     this._channel.bind('log', (_trans, params) => {
-      this._handleConsoleLog(params.output);
+      this._handleConsoleLog(params.args);
     });
   }
 
--- a/src/previewSupport/handleConsoleLogs.js
+++ b/src/previewSupport/handleConsoleLogs.js
@@ -1,22 +1,8 @@
-import isString from 'lodash/isString';
 import channel from './channel';
-// eslint-disable-next-line no-eval
-const globalEval = window.eval;
-
-function reducer(accumulator, currentValue) {
-  let evaluatedValue;
-  if (isString(currentValue)) {
-    evaluatedValue = currentValue;
-  } else {
-    evaluatedValue = globalEval(currentValue);
-  }
-
-  return `${accumulator} ${JSON.stringify(evaluatedValue)}`;
-}
 
 function notifyChannel(args) {
   if (args.length > 0) {
-    const output = args.reduce(reducer);
+    const output = args.map(arg => JSON.stringify(arg)).join(' ');
     channel.notify({
       method: 'log',
       params: {output},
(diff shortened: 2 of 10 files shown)


src/components/PreviewFrame.jsx
```diff
@@ -29,7 +29,7 @@ class PreviewFrame extends React.Component {
 
     if (this._channel && isActive) {
       for (const [key, {expression}] of newProps.consoleEntries) {
-        if (!previousConsoleEntries.has(key)) {
+        if (!previousConsoleEntries.has(key) && expression) {
```
alexpelan: I used the existing Record for `ConsoleEntry` and `ConsoleEntry` component, with an empty expression and a check here to not evaluate expressions that are falsy. 

Wanted your opinion - this seemed a little simpler, but I could also see us using different Record and different component for console log entries - we'd have to add a 'type' to the console state to know which one to render, but wouldn't be too bad.

src/previewSupport/handleConsoleLogs.js
```diff
@@ -0,0 +1,68 @@
...
+      method: 'log',
+      params: {output},
+    });
+  }
+}
+
+export default function handleConsoleLogs() {
+  /* eslint-disable no-console */
+  const browserConsoleLog = console.log.bind(console);
+  const browserConsoleDebug = console.debug.bind(console);
```
outoftime: Let’s make sure `console.debug` exists in the oldest browser versions we support ([according to caniuse the other ones do](https://caniuse.com/#search=console))

src/previewSupport/handleConsoleLogs.js
```diff
@@ -0,0 +1,68 @@
...
+  const browserConsoleInfo = console.info.bind(console);
+  const browserConsoleWarn = console.warn.bind(console);
+  const browserConsoleError = console.error.bind(console);
+  /* eslint-enable no-console */
+
+  Object.defineProperties(console, {
```
outoftime: I would be inclined to do this in a loop

src/previewSupport/handleConsoleLogs.js
```diff
@@ -0,0 +1,68 @@
...
+  } else {
+    evaluatedValue = globalEval(currentValue);
+  }
+
+  return `${accumulator} ${JSON.stringify(evaluatedValue)}`;
+}
+
+function notifyChannel(args) {
+  if (args.length > 0) {
+    const output = args.reduce(reducer);
```
outoftime: Isn’t this essentially a `map` + `join`? That seems like a simpler way to write it than a `reduce`.

src/reducers/console.js
```diff
@@ -28,6 +28,11 @@ export default function console(stateIn, {type, payload, meta}) {
         meta.key,
         new ConsoleEntry({expression: payload}),
       );
+    case 'CONSOLE_LOG_PRODUCED':
+      return state.set(
+        payload.key,
+        new ConsoleEntry({expression: '', value: payload.value}),
```
outoftime: No need to pass an explicit `expression` here, can just use the record’s default (which as mentioned elsewhere I think should change to `null`)

src/components/PreviewFrame.jsx
alexpelan: ConsoleEntry now has a default null expression and ConsoleEntrys created by logs just use that default.

src/previewSupport/handleConsoleLogs.js
```diff
@@ -0,0 +1,68 @@
...
+import channel from './channel';
+// eslint-disable-next-line no-eval
+const globalEval = window.eval;
+
+function reducer(accumulator, currentValue) {
+  let evaluatedValue;
+  if (isString(currentValue)) {
+    evaluatedValue = currentValue;
+  } else {
+    evaluatedValue = globalEval(currentValue);
```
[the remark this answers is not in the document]
alexpelan: If I don't do the string check, then eval tries to evaluate whatever string passed to it as code. eval("whatever"); in your console says 'Uncaught ReferenceError: whatever is not defined.'

We need to eval stuff because what if I do console.log(myFunction(6)) or, simpler, console.log(5 * 6).

src/previewSupport/handleConsoleLogs.js
alexpelan: done

src/previewSupport/handleConsoleLogs.js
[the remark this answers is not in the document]
outoftime: @alexpelan but if you do `console.log(5 * 6)` then our `log` implementation will receive the value `30`, won’t it?

And as you rightly point out, if we were to do `console.log("5 * 6")`, we would want it to print `"5 * 6"`, not `30`.

So I am still not clear on why we want `eval` here at all?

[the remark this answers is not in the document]
alexpelan: hmmmmmmmmm I don't know how I ended up with this impression but I swear at one point our implementation was not receiving 30 and I had to put the eval in to get it to work right. You are correct, we can literally just do a join(" ") on the arguments it appears.

src/previewSupport/handleConsoleLogs.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+}
+
+export default function handleConsoleLogs() {
+  /* eslint-disable no-console */
+  const browserConsoleLog = console.log.bind(console);
+  const browserConsoleDebug = console.debug.bind(console);
+  const browserConsoleInfo = console.info.bind(console);
+  const browserConsoleWarn = console.warn.bind(console);
+  const browserConsoleError = console.error.bind(console);
```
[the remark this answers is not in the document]
alexpelan: I couldn't figure this out? Didn't seem to be a way to set `this` with lodash wrap? I could use lodash's bind?

src/previewSupport/handleConsoleLogs.js
alexpelan: DRYed

alexpelan:
@outoftime take another look?

src/previewSupport/handleConsoleLogs.js
```diff
@@ -0,0 +1,32 @@
+import channel from './channel';
+
+function notifyChannel(args) {
+  if (args.length > 0) {
+    const output = args.map(arg => JSON.stringify(arg)).join(' ');
```
outoftime: As discussed IRL, let’s send the values as an array and then stringify on the other end.

src/previewSupport/handleConsoleLogs.js
```diff
@@ -0,0 +1,32 @@
...
+    channel.notify({
+      method: 'log',
+      params: {output},
+    });
+  }
+}
+
+export default function handleConsoleLogs() {
+  const consoleFunctions = ['log', 'info', 'warn', 'error'];
```
outoftime: This seems best defined at the top module level?

src/previewSupport/handleConsoleLogs.js
```diff
@@ -0,0 +1,32 @@
...
+
+  // eslint-disable-next-line no-console
+  if (console.debug) {
+    consoleFunctions.push('debug');
+  }
```
outoftime: So I actually think we do want to allow `console.debug` even if the browser doesn’t define it—should just do a noop for the `browserFunction`. WDYT?

src/previewSupport/handleConsoleLogs.js
```diff
@@ -0,0 +1,32 @@
...
+
+  consoleFunctions.forEach((functionName) => {
+    // eslint-disable-next-line no-console
+    const browserFunction = console[functionName].bind(console);
+    // eslint-disable-next-line prefer-reflect
+    Object.defineProperty(console, functionName, {
```
outoftime: Did `Reflect` cause a problem here? We might need to pull the Babel polyfill into the preview support script…

alexpelan: I saw Reflect wasn't supported back to Chrome 22 and the other preview frame code also disabled prefer-reflect - do you want me to make this Babel polyfill change in this PR?

alexpelan:
@outoftime  I just imported 'babel-polyfill' at the top of the preview module, I don't actually have access to a browser without `Reflect` so wasn't sure if it actually worked.

PR #1302, opened 2017-12-19
add clear button to console

pwjablonski:
Had some fun on my train ride out to AOIT. Below does a first step towards clearing console. It's in the Backlog but it was so appetizing I couldn't resist. 

Questions...
1. Is it important to keep track of console history? ie should we remove the records completely or just move the entries up out of view ala terminal clear. 
2. Not sure if stopping the propagation is the correct way to block the click through.
3. Is circle with a line through the right type of button to have. I was mimicking the chrome console but not sure if this is the best for students. 

Notes
1. I'm waiting on the console.log() implementation to figure out console.clear() and command L.

### DIFF
--- a/src/components/Console.jsx
+++ b/src/components/Console.jsx
@@ -49,13 +49,19 @@ export default function Console({
         className="label console__label"
         onClick={partial(onToggleVisible, currentProjectKey)}
       >
-        Console
-        <span className="console__chevron u__icon">{chevron}</span>
+        <div>
+          Console
+          <span className="console__chevron u__icon">{chevron}</span>
+        </div>
         <span
           className="console__button console__button_clear u__icon"
-          onClick={onClearConsoleEntries}
+          // eslint-disable-next-line react/jsx-no-bind
+          onClick={(e) => {
+            e.stopPropagation();
+            onClearConsoleEntries();
+          }}
         >
-          {'\uf05e'}
+          &#xf05e;
         </span>
       </div>
       {isOpen ? console : null}
--- a/src/components/Console.jsx
+++ b/src/components/Console.jsx
@@ -13,6 +13,7 @@ export default function Console({
   isEnabled,
   isOpen,
   isTextSizeLarge,
+  onClearConsoleEntries,
   onInput,
   onToggleVisible,
 }) {
@@ -50,6 +51,12 @@ export default function Console({
       >
         Console
         <span className="console__chevron u__icon">{chevron}</span>
+        <span
+          className="console__button console__button_clear u__icon"
+          onClick={onClearConsoleEntries}
+        >
+          {'\uf05e'}
+        </span>
       </div>
       {isOpen ? console : null}
     </div>
@@ -63,6 +70,7 @@ Console.propTypes = {
   isEnabled: PropTypes.bool.isRequired,
   isOpen: PropTypes.bool.isRequired,
   isTextSizeLarge: PropTypes.bool,
+  onClearConsoleEntries: PropTypes.func.isRequired,
   onInput: PropTypes.func.isRequired,
   onToggleVisible: PropTypes.func.isRequired,
 };
--- a/src/components/Console.jsx
+++ b/src/components/Console.jsx
@@ -53,16 +53,15 @@ export default function Console({
           Console
           <span className="console__chevron u__icon">{chevron}</span>
         </div>
-        <span
+        <div
           className="console__button console__button_clear u__icon"
-          // eslint-disable-next-line react/jsx-no-bind
           onClick={(e) => {
             e.stopPropagation();
             onClearConsoleEntries();
           }}
         >
           &#xf05e;
-        </span>
+        </div>
       </div>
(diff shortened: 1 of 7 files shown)


src/containers/Console.js
```diff
@@ -23,6 +27,11 @@ function mapStateToProps(state) {
 
 function mapDispatchToProps(dispatch) {
   return {
+    onClearConsoleEntries(e) {
+      e.stopPropagation();
```
outoftime: So `stopPropagation()` is the right place to do, but this isn’t an ideal place to do it—containers shouldn’t know anything about the UI that’s made by the presentational components they wrap. Here we are leaking the knowledge of what actual UI would trigger the `clearConsoleEntries()` callback.

So, we should do the `stopPropagation` inside the presentational component. A couple of options:

1. Use an arrow function for the `onClick`, which stops propagation and calls back to `onClearConsoleEntries`. You’d also have to remove the `jsx-no-bind` eslint rule, which I think is basically fine, since it’s based on a misunderstanding of performance characteristics anyway.
2. Change `<Console>` into a class component and define a `_handleClickClearButton` method which is then given as the `onClick`. This has the advantage of maintaining our purity with respect to `jsx-no-bind`, but little else.

src/css/application.css
```diff
@@ -483,6 +483,10 @@ body {
   background-color: var(--color-low-contrast-gray);
 }
 
+.console__button_clear {
+  float: right;
```
outoftime: nooooooooo anything but `float`!

(practical suggestion: either flexbox or just absolutely position this)

outoftime:
> Is it important to keep track of console history? ie should we remove the records completely or just move the entries up out of view ala terminal clear.

I think removing them completely is right unless/until we have a UI that allows you to retrieve them, which I definitely don’t think we need out the gate. I also have never met a browser console that lets you get back to entries you’ve cleared.

> Is circle with a line through the right type of button to have. I was mimicking the chrome console but not sure if this is the best for students.

It seems reasonable, I’m certainly open to other suggestions, but probably not worth agonizing over at this point—if it’s unclear, someone will suggest something better or at least flag that it’s not intuitive to students.

> I'm waiting on the console.log() implementation to figure out console.clear() and command L.

Definitely makes sense on `console.clear()`—I don’t think control-L has any particular overlap with that work, but also definitely not a must-have for the first release of this.

src/components/Console.jsx
```diff
@@ -50,6 +51,12 @@ export default function Console({
       >
         Console
         <span className="console__chevron u__icon">{chevron}</span>
+        <span
+          className="console__button console__button_clear u__icon"
+          onClick={onClearConsoleEntries}
+        >
+          {'\uf05e'}
```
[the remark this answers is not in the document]
pwjablonski: Ahh that makes sense!

src/containers/Console.js
pwjablonski: Will go with option 1 :)

src/css/application.css
pwjablonski: Hahaha sounds good! Saw it used in the preview button (pop out and refresh) so thought I'd give it a try. Will update!

pwjablonski:
@outoftime added flexbox. I feel like the div seems a little off without a class name. Should we give a class or does it not matter.

src/components/Console.jsx
```diff
@@ -49,13 +49,19 @@ export default function Console({
...
-        Console
-        <span className="console__chevron u__icon">{chevron}</span>
+        <div>
+          Console
+          <span className="console__chevron u__icon">{chevron}</span>
+        </div>
         <span
           className="console__button console__button_clear u__icon"
-          onClick={onClearConsoleEntries}
+          // eslint-disable-next-line react/jsx-no-bind
```
outoftime: Let’s just remove this from `.eslintrc` altogether—`eslint-disable` is good in the case where a generally good rule should be turned off for a special case or false positive. But I basically just don’t think we should have this rule because arrow functions in components aren’t actually bad.

src/components/Console.jsx
```diff
@@ -49,13 +49,19 @@ export default function Console({
         className="label console__label"
         onClick={partial(onToggleVisible, currentProjectKey)}
       >
-        Console
-        <span className="console__chevron u__icon">{chevron}</span>
+        <div>
+          Console
+          <span className="console__chevron u__icon">{chevron}</span>
+        </div>
         <span
```
outoftime: I would make this a `div` as well since it’s a flex item (I think practically it doesn’t matter, `<div>` just seems more semantic)

src/components/Console.jsx
```diff
@@ -49,13 +49,19 @@ export default function Console({
         className="label console__label"
         onClick={partial(onToggleVisible, currentProjectKey)}
       >
-        Console
-        <span className="console__chevron u__icon">{chevron}</span>
+        <div>
```
outoftime: In response to your question, I think it’s fine to not give this a class name if we don’t need to style it!

PR #1304, opened 2017-12-21
Click in console focuses input

### LINKED ISSUE #1272: Clicking anywhere in the console should focus the input
Right now, you have to click precisely on the console input for it to be focused. Clicking anywhere within the console area should focus the input.

This is conceptually related to the mechanism by which clicking an error message in the preview focuses the erroneous line in the editor, so we should probably generalize the data flow that handles that.

pwjablonski:
@outoftime Here's my first pass at focusing the console when console clicked. Used the existing `focusLine` logic from error focusing functionality. 

Notes
1. I haven't disabled the eslint on arrow functions in props in this PR. Also since it is a class component should I do it the "right" way and create a separate function to handle this?

2. On `focusLine` takes a row and column. I am passing in zero's as fillers. The cursor gets moved to the begin of the line and then,  use `navigateLineEnd()` to place the cursor at the end of the line. Seems like this is taking an extra step. Open to ideas on doing this more directly.

Closes #1272


src/components/ConsoleInput.jsx
```diff
@@ -14,12 +15,33 @@ export default class ConsoleInput extends Component {
...
+    if (nextProps.isTextSizeLarge !== this.props.isTextSizeLarge) {
       requestAnimationFrame(() => {
         inheritFontStylesFromParentElement(this._editor);
       });
     }
+    this._focusRequestedLine(nextProps.requestedFocusedLine);
+  }
+
+  _focusRequestedLine(requestedFocusedLine) {
+    if (get(requestedFocusedLine, 'language') !== 'console') {
```
outoftime: Let’s generalize the structure of `requestedFocusedLine`—instead of `language` let’s maybe `component`, and just do `editor.html`, `editor.css`, `console`, etc. Reasonable?

src/components/ConsoleInput.jsx
```diff
@@ -14,12 +15,33 @@ export default class ConsoleInput extends Component {
     bindAll(this, '_ref');
   }
 
-  componentWillReceiveProps({isTextSizeLarge}) {
-    if (isTextSizeLarge !== this.props.isTextSizeLarge) {
+  componentDidMount() {
+    this._focusRequestedLine(this.props.requestedFocusedLine);
+  }
+
+  componentWillReceiveProps(nextProps) {
```
outoftime: I think destructuring `nextProps` would still be nicer here?

src/components/ConsoleInput.jsx
```diff
@@ -51,15 +73,25 @@ export default class ConsoleInput extends Component {
   }
 
   render() {
-    return <div className="console__input" ref={this._ref} />;
+    return (
+      <div
+        className="console__input"
+        ref={this._ref}
+        // eslint-disable-next-line react/jsx-no-bind
+        onClick={e => e.stopPropagation()}
```
outoftime: I would just use `react-prevent-clickthrough` here (arguably we should have used it in the other one too but more of a slam dunk when the _only_ thing we’re doing is preventing clickthrough)

src/components/ConsoleInput.jsx
```diff
@@ -14,12 +15,33 @@ export default class ConsoleInput extends Component {
...
+
+  _focusRequestedLine(requestedFocusedLine) {
+    if (get(requestedFocusedLine, 'language') !== 'console') {
+      return;
+    }
+
+    this._editor.moveCursorTo(
+      requestedFocusedLine.line,
+      requestedFocusedLine.column,
+    );
```
outoftime: Honestly I think we should just ignore `line` and `column` altogether and just call `_editor.navigateFileEnd()`

src/components/ConsoleInput.jsx
```diff
@@ -14,12 +15,33 @@ export default class ConsoleInput extends Component {
     bindAll(this, '_ref');
   }
 
-  componentWillReceiveProps({isTextSizeLarge}) {
-    if (isTextSizeLarge !== this.props.isTextSizeLarge) {
+  componentDidMount() {
+    this._focusRequestedLine(this.props.requestedFocusedLine);
```
outoftime: Seems more strictly correct to do this in the `_ref()` method, if there is an editor element?

src/components/ConsoleInput.jsx
```diff
@@ -51,15 +73,25 @@ export default class ConsoleInput extends Component {
   }
 
   render() {
-    return <div className="console__input" ref={this._ref} />;
+    return (
+      <div
+        className="console__input"
+        ref={this._ref}
+        // eslint-disable-next-line react/jsx-no-bind
```
outoftime: No need for this if you merge `master`!

src/components/Editor.jsx
```diff
@@ -74,7 +74,7 @@ class Editor extends React.Component {
 
     this._scrollToLine(requestedFocusedLine.line);
 
-    this._editor.clearSelection();
+    // this._editor.clearSelection();
```
outoftime: What’s this about?

src/containers/Console.js
```diff
@@ -30,6 +37,16 @@ function mapDispatchToProps(dispatch) {
...
+
+    onConsoleClicked() {
+      dispatch(focusLine('console', 0, 0));
+    },
+
+    onRequestedLineFocused() {
+      dispatch(editorFocusedRequestedLine());
+    },
+
+
```
outoftime: Extra whitespace

src/selectors/getRequestedFocusedLine.js
```diff
@@ -0,0 +1,8 @@
+export default function(state) {
```
outoftime: We can give the function a name here so let’s do it.

src/components/ErrorList.jsx
```diff
@@ -16,7 +16,7 @@ function ErrorList({errors, onErrorClick, language}) {
       key={[error.reason, error.row]}
       onClick={partial(
         onErrorClick,
-        language,
+        `editor.${language}`,
```
outoftime: I would actually keep this the way it was, since the `ErrorList` component shouldn’t need to know the details of the store—from its standpoint, it’s just giving some information about an error being clicked. We can decorate it in the `Workspace#_handleErrorClick` method, which has direct knowledge of the `focusLine` action and what format it expects arguments in.

pwjablonski: Makes sense. I also just noticed the `Workspace#_handleErrorClick` method looks like it is never used. The `onErrorClick` in the `ErrorReport` Container seems to take care of this.

outoftime: Nice—feel free to remove the unused method as well!

src/sagas/ui.js
```diff
@@ -50,10 +53,15 @@ export function* exportProject() {
   );
 }
 
+export function* clearConsoleEntries() {
+  yield put(focusLine('console', 0, 0));
```
outoftime: In general we want to structure our Redux code so that it is reactive rather than imperative—actions should describe _what happened_, then the reducers and sagas should know what to do in response to the action. One simple rule of thumb is that we should never dispatch two different actions from the same context. As a corollary, we shouldn’t have a saga whose only job is to consume one action and dispatch another.

Here we are dispatching an action with an imperative intent: “focus this line”. But there is already an action being dispatched which says what happened: the `CLEAR_CONSOLE_ENTRIES` action. The `ui` reducer can thus update its state (specifically `requestedFocusedLine`) accordingly, with no need for a second action dispatch.

Since this particular bit of functionality is being tackled in a separate PR already, I think the best bet would be to just revert the commit that introduced this logic and we’ll merge this PR as it was previously.

pwjablonski:
@outoftime ah apologies! Jumped the gun on this one. Thanks for the explanation. Good to revert to last commit and let the other PR handle it!

pwjablonski:
@outoftime rolled back changes previous commit

PR #1309, opened 2017-12-23
Hides the <Console/> if the project has errors

### LINKED ISSUE #1270: Console should not be visible when there are validation errors
When the student errors appear in the preview pane, the console is moved to the top of the screen. It should disappear altogether.

ajgreenb:
Implementation copied from the <Preview/> component and its
associated container.

Fixes #1270


ajgreenb:
Are there component tests for testing this case? I see we test reducers, sagas, and validations, but I couldn't find any component tests.

PR #1318, opened 2017-12-29
Fix console input focus on input clearing.

### LINKED ISSUE #1317: After clearing console, input should be focused
Right now, after clearing the console, the input is unfocused. It should be focused.

Should be pretty easy once we have #1304 in place.

harry1064:
fixes #1317

### DIFF
--- a/src/containers/Console.js
+++ b/src/containers/Console.js
@@ -35,6 +35,7 @@ function mapDispatchToProps(dispatch) {
   return {
     onClearConsoleEntries() {
       dispatch(clearConsoleEntries());
+      dispatch(focusLine('console', 0, 0));
     },
 
     onInput(input) {
--- a/src/containers/Console.js
+++ b/src/containers/Console.js
@@ -33,7 +33,6 @@ function mapDispatchToProps(dispatch) {
   return {
     onClearConsoleEntries() {
       dispatch(clearConsoleEntries());
-      dispatch(focusLine('console', 0, 0));
     },
 
     onInput(input) {
--- a/src/reducers/ui.js
+++ b/src/reducers/ui.js
@@ -78,6 +78,15 @@ export default function ui(stateIn, action) {
           set('column', action.payload.column),
       );
 
+    case 'CLEAR_CONSOLE_ENTRIES':
+      return state.setIn(
+        ['editors', 'requestedFocusedLine'],
+        new Immutable.Map().
+          set('component', 'console').
+          set('line', 0).
+          set('column', 0),
+      );
+
     case 'EDITOR_FOCUSED_REQUESTED_LINE':
       return state.setIn(['editors', 'requestedFocusedLine'], null);
 


src/components/ConsoleInput.jsx
```diff
@@ -20,6 +20,9 @@ export default class ConsoleInput extends Component {
         inheritFontStylesFromParentElement(this._editor);
       });
     }
+    if (!this._editor.isFocused()) {
+      this._editor.focus();
+    }
```
outoftime: I don’t think this is quite the logic we want—in this case the console will focus any time props are pushed out (which might not even involve a change to the props it’s getting). A concrete example of bad behavior:

* In the JavaScript project code, add a `console.log()` line
* Open the console
* Make a trivial modification to the JavaScript code
* Observe that the console input is now focused, which is not expected/desirable behavior

I believe the best bet will be to add logic that updates the `ui.requestedFocusedLine` Redux state when the `CLEAR_CONSOLE_ENTRIES` action is consumed—however this will need to wait on #1304 which generalizes `requestedFocusedLine` to work with the console. That PR is approved, I just need to get a clean build with the branch up to date from `master`.

harry1064: Thanks for the feedback. Its transfering the focus to console on change in javascript code. Will have a look at #1304

harry1064:
have pushed another solution. Sorry for the inconsistent new commit. But have a look at this new solution if this is helpful.

harry1064:
@outoftime You have valid points. Will look into new changes from #1304 .

harry1064:
@outoftime you were right. `requestedFocusedLine` works like charm.

src/containers/Console.js
```diff
@@ -35,6 +35,7 @@ function mapDispatchToProps(dispatch) {
   return {
     onClearConsoleEntries() {
       dispatch(clearConsoleEntries());
+      dispatch(focusLine('console', 0, 0));
```
outoftime: This is a totally reasonable instinct, but in the Popcode codebase we tend to think about Redux actions more like _events_ and less like _methods_. Which is to say, an action should convey **what happened**, not instruct **what to do**. Actions that are formulated in the imperative tense typically convey a direct user intent.

Given the above, it’s usually if not always an antipattern to call two actions synchronously in a row. This would imply that two different things happened at the exact same time, which is unlikely.

Instead, we should use a single action to convey what happened—in this case `clearConsoleEntries()` is appropriate, because that is the direct user intent. Then each reducer can decide how the state it’s concerned with should change based on that action. In this case, the `ui` reducer would update the `requestedFocusedLine` property appropriately when consuming the `CLEAR_CONSOLE_ENTRIES` action.

This definitely isn’t the only way to think about the way actions are structured in a Redux application, but having tried out the more “imperative” approach of actions more directly describing the changes to state that they want to effect, I’ve found that the fully reactive approach is easier to reason about and makes the code more nimble as it develops over time.

harry1064:
@outoftime as per your comment, i have add case `CLEAR_CONSOLE_ENTRIES` in ui reducer which i believe not breaking the pattern now.

PR #1335, opened 2018-01-04
Fix react key warning for ErrorItem.

### LINKED ISSUE #1226: React key warning in development mode
When developing, if you open a fresh Popcode environment and type `<h1` into the HTML editor, the console displays this React warning:

```
Warning: Encountered two children with the same key, `unclosed-tag,8`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in ul (created by ErrorList)
    in div (created by ErrorList)
    in ErrorList (created by ErrorReport)
    in div (created by ErrorReport)
    in ErrorReport (created by Connect(ErrorReport))
    in Connect(ErrorReport) (created by Output)
    in div (created by Output)
    in div (created by Output)
    in Output (created by Workspace)
    in div (created by Workspace)
    in div (created by Workspace)
    in main (created by Workspace)
    in div (created by Workspace)
    in Workspace (created by Connect(Workspace))
    in Connect(Workspace) (created by Application)
    in Provider (created by Application)
    in Application
```

You don't see this in production since the warnings get stripped in the production version of React. Just wanted to note this—feel free to close if you already know about it or aren't worried about it.

harry1064:
fixes #1226

### DIFF
--- a/src/components/ErrorList.jsx
+++ b/src/components/ErrorList.jsx
@@ -10,10 +10,10 @@ function ErrorList({errors, onErrorClick, language}) {
     return false;
   }
 
-  const errorItems = map(errors.items, error => (
+  const errorItems = map(errors.items, (error, index) => (
     <ErrorItem
       {...error}
-      key={[error.reason, error.row]}
+      key={[error.reason, error.row, index]}
       onClick={partial(
         onErrorClick,
         language,


src/components/ErrorList.jsx
```diff
@@ -10,10 +10,10 @@ function ErrorList({errors, onErrorClick, language}) {
     return false;
   }
 
-  const errorItems = map(errors.items, error => (
+  const errorItems = map(errors.items, (error, index) => (
     <ErrorItem
       {...error}
-      key={[error.reason, error.row]}
+      key={[error.reason, error.row, index]}
```
outoftime: If we’re going to use the `index` in the `key`, I don’t think there’s any reason to still use the `reason` and `row`, since the `index` alone provides a unique (if not stable) identifier for each error.

However [using an array index in a key is generally discouraged](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318). The underlying problem here is that the `reason` is not sufficiently unique—in particular, there are two different `unclosed-tag` errors, for two different tags, but on the same line.

If instead of `error.reason` we used `error.raw` (a string containing the full error message in Markdown), I think we’d be in good shape while avoiding the problems with using an index?

outoftime:
This one has been dormant for awhile so I’m going to go ahead and close it—feel free to pick it back up whenever @harry1064 !

PR #1336, opened 2018-01-06
Add title to Google Classroom Export

pwjablonski:
@outoftime Thought this would be a quick and nice addition the Classroom Export Flow.

Not sure if I am generating the title in the right place.  It was the easiest to implement rather than updating/creating a selector.


src/sagas/clients.js
```diff
@@ -38,7 +39,8 @@ export function* exportProject({payload: {exportType}}) {
       ({html_url: url} = yield call(createRepoFromProject, project, user));
     } else if (exportType === 'classroom') {
       const snapshotKey = yield call(createProjectSnapshot, project);
-      url = yield call(createShareToClassroomUrl, snapshotKey);
+      const projectTitle = generateTextPreview(project);
```
outoftime: In general in sagas we should prefer `yield call()` to directly calling other modules in the application, to make unit testing easier. (Not a totally hard-and-fast rule—happy to elaborate if it’s not clear where the line is)

PR #1337, opened 2018-01-09
Print undefined for certain entries; fixes #1333

### LINKED ISSUE #1333: Distinguish between non-expression statements and expressions with undefined values
Entering the following in the console:

```js
var i = 1
```

Should yield no output, because it is not an expression.

Entering the following, however:

```js
window.nothingHere
```

Should output `undefined`, because it is an expression with an undefined value.

Both code fragments return `undefined` when passed to `eval`, so we will need to do something smarter to distinguish between them (I’m guessing `esprima` might be able to help us tell the difference).

tenyoung795:
Specifically, whitelists expression statements whose expression is assignment, an identifier, a literal (i.e. `undefined`), or a member expression (e.g. `obj.foo`, `arr[0]`).

Testing caveats: for some reason, the console answers never appear when I run via `yarn dev`. The answers appear from `yarn preview`, but I have to open the server in incognito mode for the `preview.js` to refresh. For `preview.js` to refresh in my normal browser session, I had to `rm -rf dist static/compiled bower_components node_modules`; not sure which directories actually cleaned up my browser session's `preview.js`.

Therefore, this PR may be incomplete; let me know how to improve testability!

### DIFF
--- a/src/components/PreviewFrame.jsx
+++ b/src/components/PreviewFrame.jsx
@@ -6,6 +6,7 @@ import Bowser from 'bowser';
 import bindAll from 'lodash/bindAll';
 import {t} from 'i18next';
 import normalizeError from '../util/normalizeError';
+import retryingFailedImports from '../util/retryingFailedImports';
 import {sourceDelimiter} from '../util/compileProject';
 import {CompiledProject as CompiledProjectRecord} from '../records';
 
@@ -41,7 +42,13 @@ class PreviewFrame extends React.Component {
     return false;
   }
 
-  _evaluateConsoleExpression(key, expression) {
+  async _evaluateConsoleExpression(key, expression) {
+    const {hasExpressionStatement} = await retryingFailedImports(
+      () => import(
+        /* webpackChunkName: 'mainAsync' */
+        '../util/javascript',
+      ),
+    );
     // eslint-disable-next-line prefer-reflect
     this._channel.call({
       method: 'evaluateExpression',
@@ -49,7 +56,7 @@ class PreviewFrame extends React.Component {
       success: (printedResult) => {
         this.props.onConsoleValue(
           key,
-          printedResult,
+          hasExpressionStatement(expression) ? printedResult : null,
           this.props.compiledProject.compiledProjectKey,
         );
       },
--- a/src/previewSupport/prettyPrint.js
+++ b/src/previewSupport/prettyPrint.js
@@ -1,9 +1,5 @@
 import inspect from 'object-inspect';
-import isUndefined from 'lodash/isUndefined';
 
 export default function prettyPrint(value) {
-  if (isUndefined(value)) {
-    return value;
-  }
   return inspect(value, {quoteStyle: 'double'});
 }
--- a/src/util/javascript.js
+++ b/src/util/javascript.js
@@ -1,0 +1,16 @@
+import {parse} from 'esprima';
+import find from 'lodash/find';
+
+function tryParse(javascript) {
+  try {
+    return parse(javascript);
+  } catch (e) {
+    return null;
+  }
+}
+
+export function hasExpressionStatement(javascript) {
+  const maybeParsed = tryParse(javascript);
+  return Boolean(maybeParsed && find(maybeParsed.body,
+    ({type}) => type === 'ExpressionStatement'));
+}
--- a/src/validations/css/css.js
+++ b/src/validations/css/css.js
@@ -22,7 +22,7 @@ class CssValidator extends Validator {
 
   async _getRawErrors() {
     const css = await retryingFailedImports(() => import(
-      /* webpackChunkName: 'linters' */
+      /* webpackChunkName: 'mainAsync' */
       'css',
     ));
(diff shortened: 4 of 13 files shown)


tenyoung795:
> Hey! This is awesome! Definitely love the implementation here, I think you found some cases I hadn’t sussed out when initially researching this. Nice work!

Thanks so much! :smile: 

> * From playing around, it looks like it’s the importing of esprima that is breaking the preview environment. It’s not obvious why, although it’s not the first time using esprima with webpack has been a pain in the ass (I haven’t been able to upgrade to esprima 4.0 because it just randomly doesn’t work for obscure reasons)
> * That said, even before that I was going to mention that maybe we want to do this logic in the main application environment rather than the preview window.

Ohh, I had found that just importing esprima broke the preview environment under `yarn dev`... on that reason alone, I'd *gladly* move code to the main environment. :+1:

But before that:

> From a semantic standpoint, I would probably frame this as “does this expression produce a meaningful value” rather than “should print undefined” (i.e. if it’s not an expression that produces a meaningful value, we don’t care what the value is, we should not print it)

Ah, that reminded me of the question of function calls ([what esprima calls a CallExpression](http://esprima.readthedocs.io/en/3.1/syntax-tree-format.html#call-and-new-expressions))... so given functions like

```js
function nothing() {}
function identity(x) {
  return x;
}
function return_only_if_true(x) {
  if (x) {
    return x;
  }
}
function return_early_if_true(x) {
  if (x) {
    return;
  }
  return x;
}
```

Would `nothing()` be meaningful? Assuming something like `identity(1)` is meaningful, what about `identity(undefined)`? Or if `return_only_if_true(true)` is meaningful, what about `return_only_if_true(false)`? `return_early_if_true(true)`?

I initially went to assume that `undefined` is not meaningful for a function call, which would result in false negatives (e.g. assuming `identity(undefined)` is meaningful).

However, I'd also consider to print `undefined` for any function call is meaningful, even if the function did not explicitly return `undefined`, since that could help teach students the reality that implicit return, and explicit return without a value, both return `undefined`. OTOH, the `console.log` calls would spam the console with `undefined`.

What do you think?

outoftime:
@tenyoung795 great question! I think we should just look at the code entered in the console itself and draw a line between expressions and statements. So, a function call would always qualify as an expression even if the function implicitly returns `undefined` (which as you point out probably is the right behavior from a learning standpoint anyway). Iʼm not too worried about `console.log()` as I wouldn't typically expect students to be calling that from the console itself. Seem reasonable?

tenyoung795:
Yep, sounds good! I simplified the logic to check just whether the JavaScript has an expression statement.

To move the esprima use out of the preview frame, I switched the `handleExpression` channel method parameter from a string to `{ expression, shouldPrint }`, and `handleExpression` only prints the "expression" result if `shouldPrint` is true. (The "expression" terminology would sound a bit weird now... Let me know if it'd be worthwhile to rename "expressions" as "entries" or the like!)

Also forgot to mention: since this PR also changes `prettyPrint` not to special-case `undefined`, console logging would previously print nothing for `undefined`, but console logging now prints `undefined`. For example, `console.log("hello", undefined, "world")` now prints `"hello" undefined "world"` instead of `"hello"  "world"`. So assuming that's a bug, this PR gets to fix that, too! :smile:

tenyoung795:
@outoftime Ah, you're right that we don't need to touch the preview frame itself; I updated my branch to just have the `PreviewFrame` component itself discard the value.

tenyoung795:
I think so; I updated the PR to do

```js
    const {hasExpressionStatement} = await retryingFailedImports(
       () => import(
         /* webpackChunkName: 'linters' */
         '../util/javascript',
       ),
     );
```

Disclaimer: I have close to no idea how the Webpack bundles work; my understanding is that:

- The "main" bundle is the set of modules that block starting the app
- The "linters" bundle is the set of modules loaded lazily for the app's main frame
- The "previewLibraries" bundle is the set of modules loaded for the app's preview frame

outoftime:
@tenyoung thatʼs a good read of the current schema although the name of the `linters` bundle comes from its original role of providing the linting libraries used by the validation system. I think the schema you propose is probably a better one–I would be biased toward having fewer, broader bundles rather than more granular ones. This bundle should also ultimately load stuff like the Firebase database client and GitHub SDK that aren’t needed at boot time (this is of course outside the scope of this PR). Maybe we should rename `linters` something like `applicationAsync`? (Definitely open to other suggestions).

outoftime:
Or I guess `mainAsync` to match `main`?

tenyoung795:
`mainAsync` SGTM! (I've no strong opinion)

I added an extra [commit](https://github.com/popcodeorg/popcode/pull/1337/commits/083edc33db4aceab2271a804db4030f747c19425) to rename `linters` to `mainAsync`; PTAL!

outoftime:
@tenyoung795 can you merge or rebase master to get the [fix for Chrome on Travis](https://github.com/popcodeorg/popcode/pull/1342) so the build can run properly? Thanks!

tenyoung795:
Yep, and hooray, the checks finally passed! :smiley:

PR #1345, opened 2018-01-26
Includes p5.js as a default library. 

demc:
Creates a new library option for p5.js (Processing).

Since Processing relies on importing any functions in the global namespace (by design), those functions and constants have been added to the library's "predefined" list for jshint.

Try snapshot ?snapshot=61ba517e-c300-4154-ac84-dbab2845a23c


yarn.lock
```diff
@@ -6476,6 +6480,12 @@ node-static@^0.7.9:
     mime "^1.2.9"
     optimist ">=0.3.4"
 
+node@^9.4.0:
+  version "9.4.0"
+  resolved "https://registry.yarnpkg.com/node/-/node-9.4.0.tgz#607459ecd79af0f91e74318124bec0537644cdee"
+  dependencies:
+    node-bin-setup "^1.0.0"
+
```
outoftime: I don’t think this is related to the change in question?

pwjablonski:
So cool @demc love p5!

demc:
Good to go? Can this go into production quickly? Thanks!

outoftime:
@demc good to go! should be in production in about ~20 minutes, modulo various levels of caching

PR #1349, opened 2018-02-01
Remove yarnlock artifacts

outoftime:
Removes `yarn.lock` artifacts from #1345


outoftime:
Opening separate PR under my own account

Issue #1359, opened 2018-02-22
Improve syntax highlighting color constrast


kaylee42:
Inspired loosely by the idea behind [colorblind-wombat-dark-syntax](https://atom.io/themes/colorblind-wombat-dark-syntax), improve the color contrast of syntax highlighting so that it is more accessible to students with various forms of colorblindness. As noted in the above linked article, most syntax highlighters need more 7 colors, so this will likely not be perfect, but at a minimum we can improve the contrast for syntax highlighting.

More research is probably needed to figure out good colors for the white background popcode uses, but here's a starting point on [color palettes for colorblindness](http://mkweb.bcgsc.ca/colorblind/)

PR #1385, opened 2018-03-05
Syntax highlighting updates

kaylee42:
In reference to #1359: 
This implements a new syntax highlighting with an eye towards it being more easily readable for colorblind users (and also me real talk I have a hard time with current syntax highlighting). 

### Before

image

### After

image

I'd like other folks to play around with this and see how you're feeling about color choices. To implement this, I started with a different theme (sqlserver - you can see the original [here](http://ajaxorg.github.io/ace-builds/kitchen-sink.html)), which I then adjusted so that the colors would be dark enough not to be color contrast failures, plus I added highlighting for numerics. 

When I was working on this I used 3 main tools: 
1) The [Spectrum Chrome extension](https://chrome.google.com/webstore/detail/spectrum/ofclemegkcmilinpcimpjkfhjfgmhieb?hl=en), which allows you to toggle through how the page would look for people with various types of colorblindness. There are a lot in the list but I focused on deuteranopia, protanopia, and tritanopia, as those were what I'd seen listed on other resources.
2) This website on [color palettes](http://mkweb.bcgsc.ca/colorblind/) for colorblindess, specifically the [12-color palette for colorblindness](http://mkweb.bcgsc.ca/colorblind/img/colorblindness.palettes.simple.png)
3) This [contrast ratio checker](http://leaverou.github.io/contrast-ratio/), which I used to ensure that the syntax highlighting was still readable (most colors pass AAA, all pass AA)

I know the linter isn't passing right now! I will come back and fix I'm just sleepy haha.

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -545,27 +545,27 @@ body {
 }
 
 /* postcss-bem-linter: ignore */
-.ace-sqlserver .ace_keyword.ace_operator,
-.ace-sqlserver .ace_lparen,
-.ace-sqlserver .ace_rparen,
-.ace-sqlserver .ace_punctuation {
-  color: #4f4f4f !important;
+.editors__editor.ace-sqlserver .ace_keyword.ace_operator,
+.editors__editor.ace-sqlserver .ace_lparen,
+.editors__editor.ace-sqlserver .ace_rparen,
+.editors__editor.ace-sqlserver .ace_punctuation {
+  color: #4f4f4f;
 }
 
-.ace-sqlserver .ace_support.ace_function {
-  color: #9e21c4 !important;
+.editors__editor.ace-sqlserver .ace_support.ace_function {
+  color: #9e21c4;
 }
 
-.ace-sqlserver .ace_storage {
-  color: #1724fb !important;
+.editors__editor.ace-sqlserver .ace_storage {
+  color: #1724fb;
 }
 
-.ace-sqlserver .ace_string {
-  color: #c40b17 !important;
+.editors__editor.ace-sqlserver .ace_string {
+  color: #c40b17;
 }
 
-.ace-sqlserver .ace_numeric {
-  color: #17942e !important;
+.editors__editor.ace-sqlserver .ace_numeric {
+  color: #17942e;
 }
 
 /** @define output */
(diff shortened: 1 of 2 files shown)


outoftime:
@kaylee42 awesome!! i updated the description with before/after screenshots so folks can take a quick look without having to check out the branch and spin up a dev environment. maybe worth hitting up the ScriptEd Popcode channel to get reax?

Overall this is a clear improvement, without feeling visually overwhelming at all. A couple issues I notice:

* In CSS, tag selectors and property names don’t seem to get any color—they do in the current production scheme
* In JavaScript, certain words are getting highlighted (`random` and `click` in the example above) for reasons other than their syntactic role (e.g. if I change it to `rando` it’s no longer highlighted). I can see how that might be nice in some scenarios but in the novice case I’m afraid that would be more confusing than helpful. The current production scheme does actually do this as well but the color is so subtle I’d never noticed it; if it’s possible to just remove this rule (or make the color the same as the baseline) that seems preferable

outoftime:
Also, did you have a corpus of sample code you were using while designing this? If so can you maybe post a gist so I/others can check out a more complete survey of available syntax than what I screenshotted above?

outoftime:
@kaylee42 any chance you’ve got some time to pick this one up and bring it to the finish line? I think it’s pretty close! would be great to have it ready for the new school year : )

outoftime:
@kaylee42 so playing with this some more, I think we probably should roll our own theme from scratch based on the 12-color palette you linked (those resources are very helpful btw!). Starting with an existing theme—particularly one that, judging by the name, is presumably targeted at SQL rather than any of the languages Popcode uses—sets up kind of a whack-a-mole situation where unexpected colors keep popping up. While playing around with the theme in this PR using more comprehensive code examples, I ran across several more situations where the highlighting isn’t what we would want (e.g. JS function parameters are a bright orange that can’t be good for low-contrast).

Looking at the [built-in ACE themes](https://github.com/ajaxorg/ace/tree/master/lib/ace/theme), it seems pretty straightforward to roll our own—each appears to be about 100 lines of CSS.

Since I haven’t heard back to you on this, I’m going to go ahead and close this PR, but I’ll be sure to add all relevant information to the corresponding issue. Of course, feel free to re-open if you want to keep working on it!!

PR #1387, opened 2018-03-09
Editable project instructions

### LINKED ISSUE #1311: Editable project instructions
Make it possible to add/edit project instructions from within Popcode. This would probably be via an option in the hamburger menu (maybe in an “Instructor” submenu?)

https://github.com/andrerpena/react-mde might be useful.

ajgreenb:
A basic first pass at making the Popcode instructions panel editable from within the app.

## To Do

* [x] Add a **Save** button that updates the project's `instructions` state.
* [x] Add tests! 🙂
* [x] Add "Markdown supported" blurb at the bottom of `InstructionsEditor`.
* [x] See if it's possible to make the `cursor: default` for the instructions bar when in editing mode.
* [ ] Polish up the styling.

Closes #1311


outoftime:
Nice! I think it makes sense for @jwang1919 to do the first review pass on this?

locales/en/translation.json
```diff
@@ -4,6 +4,7 @@
     "export-gist": "Export Gist",
     "export-repo": "Export Repo",
     "share-to-classroom": "Share To Classroom",
+    "add-or-edit-instructions": "Add/Edit Instructions",
```
jwang1919: I prefer having "Add Instructions" and "Edit Instructions" as two different strings. And have these strings show up whether or not instructions are present in the Popcode project.

locales/en/translation.json
```diff
@@ -60,7 +61,10 @@
         "css": "CSS",
         "javascript": "JS"
       },
-      "output": "Output"
+      "output": "Output",
+      "instructions": {
+        "cancel": "Cancel"
```
jwang1919: Not quite sure why this string is separate from "`add-or-edit-instructions"`.
Couldn't we do...
```
"instructions": {
    "add": "Add Instructions",
    "edit": "Edit Instructions",
    "cancel": "Cancel"
}
```

jwang1919: Purposely omitted `"Save"` because you haven't accomplished that yet, but same deal as with the other strings.

src/css/application.css
```diff
@@ -342,6 +342,31 @@ body {
...
+    height: 100%;
+    margin: 0;
+    outline: 0;
+    word-wrap: break-word;
+    overflow-wrap: break-word;
+    white-space: normal;
+  }
+}
+
+.instructions-editor-menu {
```
jwang1919: Suggest to follow BEM naming rules like the rest of the selectors.
`.instructions-editor-menu` to `.instructions-editor__menu`

src/components/InstructionsEditor.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {t} from 'i18next';
+
+export default function InstructionsEditor({instructions, onCancelEditing}) {
+  return (
+    <div className="instructions-editor">
+      <div className="instructions-editor-menu">
```
jwang1919: Suggest to follow BEM naming rules like the rest of the selectors.
`"instructions-editor-menu"` to `"instructions-editor__menu"`

src/components/TopBar/HamburgerMenu.jsx
```diff
@@ -37,6 +39,17 @@ const HamburgerMenu = createMenu({
         </MenuItem>,
       );
 
+      if (isExperimental) {
+        items.push(
+          <MenuItem
+            key="addOrEditInstructions"
+            onClick={isEditingInstructions ? noop : onStartEditingInstructions}
```
jwang1919: When `isEditingInstructions` is true, I think we should change the Menu Item in someway so it's clear that this menu item is active.

My ideas:
1. Disable the Edit Menu Item
2. Change the text to "Stop Editing Instructions" (this will cancel and close the instructions)
3. Change the text to "Save Instructions" (this will save and close the instructions).

locales/en/translation.json
ajgreenb: I separated them because `add-or-edit-instructions` (soon to be split into two separate strings) is in the **`top-bar`**, while `cancel` and `save` are in the `Instructions` **component**. But I don't feel strongly about that separation, so I'm happy to group them all together if you think that would be better.

ajgreenb:
@jwang1919 I think I've addressed all your feedback (on one point I added a comment, but GitHub has collapsed it.)

I decided to opt for idea 1 when `isEditingInstructions` is `true` (disabling the menu item.) My reasoning is that

1. It's not obvious whether _saving_ or _canceling_ should be the behavior of the menu item when `isEditingInstructions` is `true`.
2. Saving and canceling are already afforded by the buttons in the `InstructionsEditor` and I've heard it's good to only have one way to do something when that's possible.

Let me know what you think!

ajgreenb:
Also, I added the ability to save the new instructions using a `contenteditable` `pre`. It works fine, but React is a little grudging:

```
Warning: A component is `contentEditable` and contains `children` managed by React. It is now your responsibility to guarantee that none of those nodes are unexpectedly modified or duplicated. This is probably not intentional.
```

How do y'all feel about this?

ajgreenb:
Added a couple tests—let me know if there are any more you'd like to see!

outoftime:
@ajgreenb what’s the reasoning behind the `contenteditable` rather than just using a `<textarea>`?

ajgreenb:
I think it gets the job done more simply than using a `textarea`. With a `textarea`, I'd have to add more state management for state that is "ephemeral" in a sense. From my understanding, the goal is to to allow users to **save** and **cancel**, so I'd need to create separate state elements for both `instructions` and `currentInstructionsBuffer` (a better name, hopefully.) When the user saves, `currentInstructionsBuffer` gets copied to `instructions` and when the user cancels, it gets discarded.

Using `contenteditable`, that behavior can be contained within the `InstructionsEditor` component, which feels less leaky to me.

outoftime:
@ajgreenb so it looks like what you’ve got here is basically an [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html) implementation using a `contenteditable` element—transient edit state lives purely in the DOM, and then is pulled out of a ref when it’s time to save changes. What you’re describing doing with a `<textarea>` is a [controlled component](https://reactjs.org/docs/forms.html) implementation, which is indeed the preferred approach to forms in React.

However, there’s no reason you couldn’t have an uncontrolled `<textarea>`—in fact, I think that would be even simpler, since you could avoid messing with `state` at all, and just set the textarea’s `defaultValue` to the `instructions` prop.

WDYT? Definitely possible there is some subtlety I am missing here…

ajgreenb:
Ah, yeah of course. It’s been a while since I worked with React so I had forgotten about uncontrolled components. I’ll make that change; thanks for the suggestion!

pwjablonski:
@ajgreenb Checked this out - this is so awesome! Not sure if you addressed this already or have it on the backlog- I found myself wanting to have the "edit instructions" button appear in the instructions pane rather than having to go back to the hamburger menu. That way I could make changes to the markdown => save => edit to fix changes more quickly. I think "add instructions" should still live in the hamburger menu though. Thoughts?

ajgreenb:
That sounds like a great idea! I’ll try implementing it as soon as I have some time. Thanks for the suggestion!

pwjablonski:
Cool! The upside to the current implementation is that the edit instructions button is less accessible to students under the hamburger menu.  But as an instructor much prefer it in the instructions pane.

outoftime:
I think a fairly subtle (maybe only appearing on hover?) edit icon in the instructions would be the move here—not super eye-catching if you’re not looking for it, but easily accessible if you’re looking doing a quick edit-save-edit-save workflow. Anyway, seems like we could follow up with that.

@ajgreenb @jwang1919 please let me know when you’re ready for me to do final review on this—would love to get it released! Very excited to never export a gist just to add an `INSTRUCTIONS.md` again : D

ajgreenb:
Okay, in the interest of getting this out, let's leave the edit icon in the instructions for a small follow-up PR. This is pretty close to ready for final review—hopefully I'll have some time this weekend to put the final touches on it.

ajgreenb:
@outoftime I think this is ready for feedback! The styles could use some love, but they don't need to block a review.

src/css/application.css
```diff
@@ -83,6 +83,7 @@
   --size-rounded-corners: 0.2em;
   --size-top-bar: 4em;
   --size-top-bar-menu-button: 2.8em;
+  --size-instructions-menu-padding: calc(28px + 1rem);
```
outoftime: Doesn’t look like this is used?

src/css/application.css
```diff
@@ -285,6 +286,12 @@ body {
   font-weight: bold;
 }
 
+.top-bar__menu-item_disabled {
```
outoftime: I think we should remove the hover from this as well?

src/css/application.css
```diff
@@ -342,6 +349,52 @@ body {
...
+  background-color: var(--color-low-contrast-gray);
+}
+
+.instructions-editor__input {
+  padding: 1rem 1rem 0;
+  flex-grow: 1;
+  overflow-y: scroll;
+
+  /* postcss-bem-linter: ignore */
+  & textarea {
```
outoftime: Any reason to break out of BEM here? We control the `<textarea>` so we could just give it an element class…

src/css/application.css
```diff
@@ -342,6 +349,52 @@ body {
...
+    height: 100%;
```
outoftime: This doesn’t seem to be having the desired effect in Chrome 65—the editor is just a couple lines high. I find absolute positioning with all four sides pinned works wonders in this kind of situation ; )

src/css/application.css
```diff
@@ -342,6 +349,52 @@ body {
...
+
+.instructions-editor {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  background-color: var(--color-low-contrast-gray);
+}
+
+.instructions-editor__input {
+  padding: 1rem 1rem 0;
```
outoftime: I think we could do with less padding, and maybe none at all?

src/css/application.css
```diff
@@ -342,6 +349,52 @@ body {
...
+  flex-grow: 1;
+  overflow-y: scroll;
+
+  /* postcss-bem-linter: ignore */
+  & textarea {
+    height: 100%;
+    width: 100%;
+    padding: 0.5rem;
+    box-sizing: border-box;
+    font-family: monospace;
```
outoftime: Let’s use Roboto to match the code editors?

src/css/application.css
```diff
@@ -342,6 +349,52 @@ body {
...
+.instructions-editor__menu {
+  padding: 0.5rem;
+  text-align: right;
+  background-color: var(--color-light-gray);
+}
+
+.instructions-editor__footer {
+  margin: 1rem;
+  padding-top: 1rem;
+  border-top: 1px solid var(--color-light-gray);
```
outoftime: I think this text could stand to be a bit smaller.

src/components/InstructionsEditor.jsx
```diff
@@ -0,0 +1,61 @@
...
+          </button>
+        </div>
+        <div className="instructions-editor__input">
+          <textarea
+            defaultValue={this.props.instructions}
+            ref={this._ref}
+          />
+        </div>
+        <div className="instructions-editor__footer">
+          Styling with Markdown is supported
```
outoftime: Let’s make this a link to a suitable Markdown guide?

src/components/Workspace.jsx
```diff
@@ -184,14 +185,20 @@ class Workspace extends React.Component {
   }
 
   _handleClickInstructionsBar() {
-    this.props.dispatch(toggleComponent(
-      get(this.props, ['currentProject', 'projectKey']),
-      'instructions',
-    ));
+    if (!this.props.isEditingInstructions) {
```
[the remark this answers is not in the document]
outoftime: Oh lol this is one of your pending checklist items!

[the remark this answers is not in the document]
ajgreenb: ✅

ajgreenb:
@outoftime I think I addressed all your feedback! Let me know what you think. I'm not in love with styling of the *Save* and *Cancel* buttons (lack thereof)—should I spend some time trying to make them look more like the buttons in the top bar?

PR #1401, opened 2018-03-24
Follow-up polish for Instructions editor UX

### LINKED ISSUE #1391: Editable Instructions: Use Hack font
Instead of Roboto use Hack font when instructions are editable.

### LINKED ISSUE #1392: Editable Instructions: Auto-focus when editing instructions
Focus on instructions panel when Add/Edit Instructions is clicked.

### LINKED ISSUE #1393: Editable Instructions: Add placeholder to textarea
Add some placeholder the editable instructions panel. Undecided on what the final print should be, but I think `Add Instructions Here` is a good start.

ajgreenb:
1. Focus on the instructions editor when it mounts (if there are no instructions yet.)
2. Add a placeholder for the instructions editor `textarea`.
3. Use the code editors' monospace font for the instructions editor.
4. Rename the `isEnabled` `MenuItem` `prop` to `isActive`.

One thing I didn't notice until today—because of the way I implemented the action/reducer logic for the instructions editor, edits to the instructions don't update the `project.updatedAt` field. This means that if you load a fresh project, spend a bunch of time editing the instructions _without touching anything else_, and then accidentally refresh the page, you'll lose your work. Should the instructions editor be part of the "grace period" flow? I think changing that would be trivial.

Also, should I implement the **Edit** button directly on the `InstructionsEditor` that only appears when you hover on it?

Fixes #1391 
Fixes #1392 
Fixes #1393

### DIFF
--- a/src/actions/projects.js
+++ b/src/actions/projects.js
@@ -21,6 +21,7 @@ export const updateProjectSource = createAction(
 export const updateProjectInstructions = createAction(
   'UPDATE_PROJECT_INSTRUCTIONS',
   (projectKey, newValue) => ({projectKey, newValue}),
+  (_projectKey, _newValue, timestamp = Date.now()) => ({timestamp}),
 );
 
 export const toggleLibrary = createAction(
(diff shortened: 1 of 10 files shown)


outoftime:
@ajgreenb good catch / yes on the grace period tip. Also do the instructions get persisted to the project in Firebase when edited?

outoftime:
@ajgreenb that’s an irritating but expected behavior of `redux-saga-test-plan` running in Tape tests—Tape only counts a test when an assertion is called, while `redux-saga-test-plan` doesn’t use assertions—just throws an error if the saga behavior doesn’t match what the test expects.

Issue #1358, opened 2018-02-22
Preserve experimental flag after initial load


alexpelan:
We currently clear the gist= or snapshot= query params after importing the gist/snapshot because we don't want them to hang around (for example: if the student refreshes after updating a gist with the gist in the query string, we don't want to reimport it and overwrite their work). However, we do want the experimental flag to hang around - mostly useful for dev for now.

PR #1404, opened 2018-03-31
Preserve experimental flag (and only experimental) after refresh

alexpelan:
Factor our functions dealing with the url query parameters to a
helper module

Fixes: #1358

### DIFF
--- a/src/util/queryParams.js
+++ b/src/util/queryParams.js
@@ -1,0 +1,30 @@
+import qs from 'qs';
+
+export function getQueryParameters(queryString) {
+  let gistId = null;
+  let snapshotKey = null;
+  let isExperimental = false;
+  if (queryString) {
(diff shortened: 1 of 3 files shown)


src/util/queryParams.js
```diff
@@ -0,0 +1,30 @@
...
+    isExperimental = Object.keys(query).includes('experimental');
+  }
+  return {
+    gistId,
+    snapshotKey,
+    isExperimental,
+  };
+}
+
+export function setQueryParameters(isExperimental) {
```
outoftime: Even though `isExperimental` is the only possible parameter, I think it would feel better for this to take a params object, since `setQueryParameters(true)` would be a pretty inscrutable function call.

alexpelan:
@outoftime should be good now

Issue #314, opened 2016-10-22
Publish to shareable URL


outoftime:
Allow coders to “publish” their work to a stable, shareable URL.

A few approaches come to mind:

1. Export a gist that contains a single file with the fully-cooked page. Run the gist through [RawGit](https://rawgit.com/). Very simple! Not clear how this would work with [multi-page sites](https://github.com/popcodeorg/popcode/issues/786) though.
2. Push a project straight to a full-fledged GitHub repo, configured to publish to `gh-pages`. Definitely more work, but very nice and robust!
3. Set up a publicly-accessible S3 bucket, upload files directly to that (this is probably a bad idea)
4. Similarly, use something like [BitBalloon](https://www.bitballoon.com/features)

I think I’m leaning Option 2 at this point.

outoftime:
Another interesting option: use `gh-pages`, as discussed [here](https://github.com/popcodeorg/popcode/issues/296)

pwjablonski:
I'm interested in messing around with the rawgit solution.

PR #1424, opened 2018-04-16
Alexpelan/update repo

alexpelan:
When a repo is exported, we store the repo name returned by the github API. This allows us to determine if there is an existing repo, reflect this in the UI, and update the repo when the 'update repo' option is selected.

I did not change any of the functionality related to the existing export repo functionality. That means notably that the following things are not addressed:

- [ ] We still automatically generate a repo name using the page's title and a timestamp for uniqueness. We may want to add UI to capture desired repo name prior to release
- [ ] The link that is returned via the banner if the repo export window is closed is a link to the github repo itself, not the shareable .github.io address - that may be more desirable?

I wanted to keep the diff reasonably sized anyway. As always, Looking forward to yr feedback.

Another step towards #314


outoftime:
🎉🎊👯‍♀️🎊🎉🎊👯‍♀️🎊🎉

alexpelan:
🤔Looks like the fix I put in to stop those timestamp diffs from being wrong doesn't work or at least doesn't work right on travis. won't be able to resolve tonight, but shouldn't be a blocker for code review.

src/components/TopBar/HamburgerMenu.jsx
```diff
@@ -12,15 +12,18 @@ const HamburgerMenu = createMenu({
   name: 'hamburger',
 
   renderItems({
+    hasExportedRepo,
     hasInstructions,
     isEditingInstructions,
     isExperimental,
     isGistExportInProgress,
     isRepoExportInProgress,
+    isRepoUpdateInProgress,
```
outoftime: I don’t feel terribly strongly about this, but I think `isRepoExportInProgress` could reasonably capture both situations here, since only one could possibly be true at any given time. So, it might be nice to save ourselves a prop.

(I do think it makes sense to keep the underlying representation in the store the same—I would just suggest combined semantics at the selector level)

src/components/TopBar/HamburgerMenu.jsx
```diff
@@ -59,14 +62,25 @@ const HamburgerMenu = createMenu({
...
-              onClick={isRepoExportInProgress ? noop : onExportRepo}
-            >
-              {t('top-bar.export-repo')}
-            </MenuItem>,
-          );
+          if (hasExportedRepo) {
+            items.push(
+              <MenuItem
+                key="updateRepo"
+                onClick={isRepoUpdateInProgress ? noop : onUpdateRepo}
```
outoftime: Seems like we could use `isDisabled` here? Will get some styling for free and also it will handle the `noop` on click.

src/components/TopBar/HamburgerMenu.jsx
```diff
@@ -59,14 +62,25 @@ const HamburgerMenu = createMenu({
...
+              >
+                {t('top-bar.update-repo')}
+              </MenuItem>,
+            );
+          } else {
+            items.push(
+              <MenuItem
+                key="exportRepo"
+                onClick={isRepoExportInProgress ? noop : onExportRepo}
```
outoftime: Ibid. `isDisabled`

src/actions/clients.js
```diff
@@ -17,7 +17,9 @@ export const exportProject = createAction(
 );
 export const projectExported = createAction(
   'PROJECT_EXPORTED',
-  (url, exportType) => ({url, exportType}),
+  (url, exportType, projectKey, name) => ({url, exportType, projectKey, name}),
```
outoftime: So adding `projectKey` here is entirely reasonable (although if we’re sticking with positional arguments it seems like it should probably be first?)

However the `name` thing seems pretty specific to repo export—e.g. a gist, snapshot, etc. doesn’t meaningfully have a name.

A couple of alternative options:

1. Just store the URL, and parse it for the information we need when updating. Not _incredibly_ satisfying to have to build in knowledge of GitHub repo URL structure, but it’s not like it’s going to change.
2. Have this action take a `exportData` parameter that can contain an object with any export-type-specific information that might be used elsewhere.

src/clients/github.js
```diff
@@ -45,19 +45,45 @@ export async function createRepoFromProject(project, user) {
...
+  return {
+    url: data.html_url,
+    name: REPO_NAME,
+  };
+}
 
-  await createPreviewFile(github, project, userName, REPO_NAME);
+export async function updateRepoFromProject(project, user) {
+  const github = clientForUser(user);
+  const REPO_NAME = project.repoName;
```
outoftime: How come this is upper-snake-case?

src/records/Project.js
```diff
@@ -15,6 +15,7 @@ export default class Project extends Record({
   hiddenUIComponents: new Set(['console']),
   updatedAt: null,
   instructions: '',
+  repoName: null,
```
outoftime: Maybe premature generalization, but this would feel better to me inside a collection or record of some sort whose role is to store “external locations that this project has been exported to”. Might also dovetail with @pwjablonski’s work on Google Classroom integration, since that’s another case where a project needs to remember a link to an external resource?

src/selectors/getCurrentProjectExportedRepoName.js
```diff
@@ -0,0 +1,9 @@
+import {createSelector} from 'reselect';
+import get from 'lodash/get';
+import getCurrentProject from './getCurrentProject';
+
+export default createSelector(
+  [getCurrentProject],
+  currentProject =>
+    currentProject ? get(currentProject, ['repoName'], '') : '',
```
outoftime: Seems like this selector should return `null` or `undefined` rather than an empty string if there is no repo in play?

src/selectors/isProjectExportInProgress.js
```diff
@@ -0,0 +1,5 @@
+export default function isProjectExportInProgress(state, exportKey) {
```
outoftime: 👍

src/sagas/clients.js
```diff
@@ -30,19 +31,21 @@ export function* exportProject({payload: {exportType}}) {
...
-  let url;
+  let url, name;
 
   try {
     if (exportType === 'gist') {
       ({html_url: url} = yield call(createGistFromProject, project, user));
     } else if (exportType === 'repo') {
-      ({html_url: url} = yield call(createRepoFromProject, project, user));
+      ({url, name} = yield call(createRepoFromProject, project, user));
+    } else if (exportType === 'repo-update') {
```
outoftime: Would it be crazy to not treat this as a separate export type? In principle the saga here could decide which client method to call based on the presence of an existing repo name/URL in the project, or the client could just provide a `createOrUpdateRepoFromProject` which could decide what to do on the same basis. The message used for success is (sensibly) identical already, and the error messages could be merged with only minor modifications. WDYT?

src/clients/github.js
```diff
@@ -159,14 +185,15 @@ async function createSourceFileInRepo(
...
-    ),
+  await performWithRetryNetworkErrors(
+    () => github.getRepo(userName, repoName).
+      writeFile(
+        branchName,
+        fileName,
+        source,
+        COMMIT_MESSAGE,
+        {},
+      ),
```
outoftime: Probably not relevant for this PR, but I did just come across [a way to create/update multiple files via the GitHub API with a fixed number of API requests](https://gist.github.com/StephanHoyer/91d8175507fcae8fb31a).

As it happens, that number of API requests is probably not meaningfully smaller than the number we make in the current implementation, so probably mostly of academic interest unless we find ourselves increasing the number of files we write to GitHub…

src/components/TopBar/HamburgerMenu.jsx
alexpelan: consolidated `isRepoUpdateInProgress` into `isRepoExportInProgress`

src/components/TopBar/HamburgerMenu.jsx
alexpelan: sgtm - also updated gist and classroom export so we're consistent across the burger

src/actions/clients.js
alexpelan: went with `exportData`

src/clients/github.js
alexpelan: ¯\_(ツ)_/¯

(probably it was a constant at some point during the development process). `repoName` it is

src/components/TopBar/HamburgerMenu.jsx
outoftime: if there’s one thing i like in a burger, it’s cheese. if there’s two, it’s cheese and consistency.

src/sagas/clients.js
alexpelan: Got rid of `repo-export` all around - this actually got rid of a lot of duplication at the cost of a conditional in the `createOrUpdateRepoFromProject`

alexpelan:
@outoftime this is ready for another review. 

I'm seeing some test flakiness - the timestamp comparison for the saga tests are sometimes off by 1, not sure if it's a problem with `sinon.useFakeTimers` or what

src/clients/github.js
```diff
@@ -25,37 +25,73 @@ function normalizeTitle(title) {
...
 
-  return data;
+  await createRepoFiles(github, project, userName, repoName);
+
+  const {data} = await performWithRetryNetworkErrors(
+    () => github.getRepo(userName, repoName).getDetails(),
+  );
+
+  return {
+    url: data.html_url,
```
outoftime: Seems like we should include `repoName` here for consistency with the create function?

src/components/TopBar/index.jsx
```diff
@@ -109,6 +111,8 @@ export default function TopBar({
         onExportToClassroom={onExportToClassroom}
         onStartEditingInstructions={
           partial(onStartEditingInstructions, currentProjectKey)}
+        onUpdateRepo={onUpdateRepo}
+
```
outoftime: Extra whitespace

src/records/Project.js
```diff
@@ -8,13 +8,18 @@ const Sources = Record({
   javascript: '',
 });
 
+const ExternalLocations = Record({
+  githubRepoName: null,
+});
```
outoftime: 👍 👍 👍

src/reducers/projects.js
```diff
@@ -209,6 +209,19 @@ export default function reduceProjects(stateIn, action) {
         action.meta.timestamp,
       );
 
+    case 'PROJECT_EXPORTED':
```
outoftime: We probably want to consume this in the `projects` saga and update the data in Firebase in this case too, right?

alexpelan:
@outoftime motivated by not wanting to have to rebase this on the google auth branch, finally got the next set of changes in tonight. If it's good to go, I'm gonna squash this all up into one commit before merge.

src/sagas/projects.js
```diff
@@ -105,11 +106,20 @@ function generateProjectKey() {
   return (date.getTime() * 1000 + date.getMilliseconds()).toString();
 }
 
+export function* projectExported({payload: {exportType}}) {
+  if (exportType === 'repo') {
+    const state = yield select();
+    yield call(saveCurrentProject, state);
+  }
+  yield call(noop);
```
outoftime: Is there a rule that sagas need to yield at least one effect, or somesuch?

alexpelan: it appears there is not. but unrelated, I just ran through a test after making this change and it looks like the reponame isn't getting persisted to firebase right, so I'm going to figure out what's going on before sending this back your way with this change

alexpelan:
@outoftime squashed up into one commit and I think it's ready to launch

pwjablonski:
@alexpelan super cool! This looks awesome! Looks like GitHub has a way to publish without creating a separate branch. https://developer.github.com/v3/repos/pages/. Not sure if this might be useful in the future.

Issue #1427, opened 2018-04-24
Upgrade to latest react version (16.3.2)


meg2208:
We will need this for https://github.com/popcodeorg/popcode/issues/1403 as well as AsyncMode for https://github.com/popcodeorg/popcode/issues/1394 (PR https://github.com/popcodeorg/popcode/pull/1426). 16.3 also adds the new lifecycle methods for Fiber. Quite a doozie!

Issue #1402, opened 2018-03-30
Replace componentWillReceiveProps and other soon-to-be-deprecated lifecycle methods


outoftime:
A future version of React 16 will [deprecate `componentWillReceiveProps()` and a couple of other lifecycle methods](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#side-effects-on-props-change); React 17 will remove them altogether.

We rely on `componentWillReceiveProps()` in combination with a `shouldComponentUpdate()` that always returns `false` to construct components that integrate with external DOM-manipulating libraries, most notably ACE editor. However it appears that the approach we’re taking isn’t the recommended one, in particular since [`shouldComponentUpdate()` may be treated as a hint rather than a rule in a future version of React](https://reactjs.org/docs/react-component.html#shouldcomponentupdate).

Instead, I believe we should just [ensure that the component’s `render()` method always returns the same thing](https://reactjs.org/docs/integrating-with-other-libraries.html), remove `shouldComponentUpdate()`, and use `componentDidUpdate()` to handle prop changes and update the DOM accordingly.

PR #1428, opened 2018-04-24
upgrade to react 16.3

meg2208:
Closes https://github.com/popcodeorg/popcode/issues/1427 and will be useful for a few other issues.

```
➜  popcode git:(react-16.3) ✗ yarn list react                      
yarn list v0.27.5
└─ react@16.3.2
```


outoftime:
@meg2208 should #1402 be included in this?

meg2208:
This version is definitely compatible with those lifecycle methods, so I would say no need, but https://github.com/popcodeorg/popcode/issues/1402 should be considered a pretty high priority since we are going to want to upgrade to 16.4 if we start using `StrictMode`. According to [this](https://medium.com/@baphemot/whats-new-in-react-16-3-d2c9b7b6193b) in this version it will only throw a deprecation warning when you turn on `StrictMode` or `AsyncMode` in 16.3. Looks like https://github.com/popcodeorg/popcode/issues/1402 is definitely a strict requirement for 16.4, but that is not out yet.

meg2208:
@outoftime I would like to merge this asap so I can use AsyncMode for  https://github.com/popcodeorg/popcode/pull/1426 and we wouldn't have to add any dependencies for the async rendering.

PR #1444, opened 2018-05-04
Simple semver dependency upgrades

outoftime:
Upgrades all semver dependencies that don’t have major complications (e.g. stylelint, where we muck with the internals)

### DIFF
--- a/package.json
+++ b/package.json
@@ -285,7 +285,7 @@
     "karma-safari-launcher": "^1.0.0",
     "karma-sinon": "^1.0.5",
     "karma-sourcemap-loader": "^0.3.7",
-    "karma-tap": "^4.1.3",
+    "karma-tap": "4.1.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -308,9 +308,6 @@
     "webpack": "^4.0.1",
     "webpack-dev-middleware": "^3.0.0"
   },
-  "optionalDependencies": {
-    "fsevents": "*"
-  },
```
outoftime: @jwang1919 you originally put this in for Windows compatibility, but I don’t _think_ it should be necessary at this point. Can you see if you can still run the project on latest `master`? Thanks!

jwang1919: Looks okay to remove the `fsevents`, but now running into a Python issue with `node-gyp`. More info here: https://github.com/popcodeorg/popcode/issues/1449

PR #1459, opened 2018-05-08
Regenerate yarn lockfile

outoftime:
The yarn lockfile is path-dependent on the particular order of package installations and upgrades that we do over time, leading to things like unnecessarily duplicate packages, sub-dependencies that are out of date, etc. Apparently the move is to just delete it and re-generate it with `yarn install`. As long as all direct dependencies are already at their semver latest, this doesn’t have any effect on the packages we interact with directly.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -412,17 +412,13 @@ acorn-jsx@^3.0.0:
   dependencies:
     acorn "^3.0.4"
 
-acorn@5.X, acorn@^5.0.3:
-  version "5.5.0"
-  resolved "https://registry.yarnpkg.com/acorn/-/acorn-5.5.0.tgz#1abb587fbf051f94e3de20e6b26ef910b1828298"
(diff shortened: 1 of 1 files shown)


meg2208:
@outoftime in my experience, you should never need to make a PR like this if yarn is used properly. Instead of adding new dependencies by manually updating the `package.json`, they should always be added using `yarn add <dependency>@<version>` and that will update the `package.json` and `yarn.lock` for you. That is how I have done it at work and that is how I did it in other PRs for this project.

Perhaps you already knew that, but just wanted to throw in my two cents!

meg2208:
I see what you mean, so `yarn` doesn't fully support updating transitive dependencies (sub dependencies). As you mentioned `yarn-tools fix-duplicates` command does accomplish this, but it seems like that should be part of yarn natively, but is not (yet).

Interestingly, I have definitely had this problem with other programming languages too, and this is a problem that [Java 9 Modules](https://www.oracle.com/corporate/features/understanding-java-9-modules.html) tried to solve by basically letting you only export transitive dependencies if you explicitly state that you want to do that.

PR #1477, opened 2018-05-19
containerize output

pwjablonski:
This PR containerizes the Output.

@outoftime I think the draggable console PR was getting a bit unruly. I thought it would be better to chunk it out into some separate PRs. Doesn't do much as a stand alone but should make it easier to implement the console divider drag later on.

### DIFF
--- a/src/containers/Output.js
+++ b/src/containers/Output.js
@@ -1,0 +1,17 @@
+import {connect} from 'react-redux';
+import Output from '../components/Output';
+import {
+  isDraggingColumnDivider,
+  getOpenTopBarMenu,
+} from '../selectors';
+
+function mapStateToProps(state) {
+  return {
+    isDraggingColumnDivider: isDraggingColumnDivider(state),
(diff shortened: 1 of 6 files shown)


src/containers/Output.js
```diff
@@ -0,0 +1,17 @@
...
+import Output from '../components/Output';
+import {
+  isDraggingColumnDivider,
+  getOpenTopBarMenu,
+} from '../selectors';
+
+function mapStateToProps(state) {
+  return {
+    isDraggingColumnDivider: isDraggingColumnDivider(state),
+    openTopBarMenu: getOpenTopBarMenu(state),
```
outoftime: I would call this something like `isAnyTopBarMenuOpen` and cast it to a boolean? (I think as written this would fail the PropType check)

pwjablonski:
Sounds cool! I will hold off on the drag-to-resize functionality then.

pwjablonski:
@outoftime should be good to go! Think I'm getting the hang of the rebasing :)

outoftime:
@pwjablonski nice!!

PR #1479, opened 2018-05-19
add export under snapshot

### LINKED ISSUE #1480: Export Functionality Under Create Snapshot Button
We should consolidate all the export functionality under the current **Create Snapshot** button, probably with a detached drop-down arrow that exposes the other possibilities (repo, gist, Google Classroom)

pwjablonski:
From the Export Repo Project Board.

Moved all export functionality to dropdown next to the snapshot button. 

screen shot 2018-05-19 at 7 00 24 pm

Fixes #1480

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -303,6 +303,18 @@ body {
   flex: 1 0 0;
 }
 
+.top-bar__snapshot {
+  margin-right: 0;
+  border-top-right-radius: 0;
+  border-bottom-right-radius: 0;
+}
+
+.top-bar__menu-button_export {
+  margin-left: 0;
+  border-top-left-radius: 0;
+  border-bottom-left-radius: 0;
+}
+
 /** @define instructions */
 
 .instructions {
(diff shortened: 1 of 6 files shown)


src/css/application.css
```diff
@@ -212,11 +212,13 @@ body {
   box-shadow: 0 0.1em 0.2em 0 rgba(0, 0, 0, 0.3);
 }
 
+.top-bar__menu-button_export,
```
outoftime: Any reason not to just give the export button the `top-bar__menu-button_primary` class rather than duplicating these properties in the `_export` class?

Generally better imo to keep the actual class definitions/rules simpler and more transparent, and then apply the classes we need in the HTML.

(To be clear, I would still propose keeping the `_export` modifier as well, to apply the rules around margin and border)

src/css/application.css
```diff
@@ -303,6 +305,18 @@ body {
   flex: 1 0 0;
 }
 
+.top-bar__snapshot {
```
outoftime: Doesn’t look like this is used?

pwjablonski: This takes the margin off the right of the snapshot button

pwjablonski: in `SnapshotButton.jsx`

pwjablonski:
Had to make changes after the rebase to use `isAuthenticatedWithGithub` in exportMenu. May want to take another quick pass to make sure its all good.

Issue #1128, opened 2017-10-15
Make it more obvious that work is saved


outoftime:
Currently it’s not terribly obvious that your work is saved when you’re logged in. Let’s make it clearer. I think in particular we want to lean on the existing **My Projects** button in the top bar, since that’s where you go to get back to saved projects.

I think the basic idea would be to add some sort of temporary visual indicator to **My Projects** whenever the coder updates their project while logged in. A simple idea might be to change the text to “Saved!” with a checkmark icon, maybe with a different text or background color (although we don’t want to get too distracting).

We’d also want to do this on a debounce so as not to be too annoying.

outoftime:
Closed by #1481

PR #1481, opened 2018-05-20
project saved indicator

pwjablonski:
This PR addresses #1128

A couple of points/questions

-  `currentProjectSaved` and `currentProjectSaved` actions executed in UI Saga under `userDoneTyping`. This may not be the right place for this. I figured since `userDoneTyping` takes every `UPDATE_PROJECT_SOURCE` this would suffice for now, however, if there is an error persisting to firebase it seems the project will still show as saved. 

- General Naming Advice
The naming of the actions `currentProjectSaved` and `currentProjectSaveFinished` seems a bit vague/not specific and `projectSaved` property in the redux also doesn't quite seem to reflect what this data represents. Any thoughts?

- Delay
Is `delay` the right function to be using for this. The desired effect is that `Your Projects` briefly turns to `Saved!` ( Currently for 2 seconds but also not sure if that is the desired length of time). 

- Saved!
I think `Saved!` looks pretty good but open to other suggestions.

may-19-2018 22-11-29


pwjablonski:
Makes a ton of sense!

outoftime:
@pwjablonski I see you have made a bunch of progress here—let me know when you’re ready for another look-through!

pwjablonski:
@outoftime Ready for another round of review. I had some trouble creating tests for the projects saga. Currently excludes testing for `saveCurrentProject`.

src/components/TopBar/ProjectPickerButton.jsx
```diff
@@ -1,16 +1,35 @@
...
+      >
+        {t('top-bar.project-saved')}
+      </span>
+    </div>
   );
 }
 
-ProjectPickerButton.propTypes = {};
+ProjectPickerButton.propTypes = {
+  isSaveIndicatorShown: PropTypes.bool.isRequired,
```
outoftime: Weakly held opinion, but `shouldShowSavedIndicator` would read more naturally to me here.

src/css/application.css
```diff
@@ -315,6 +315,20 @@ body {
   border-bottom-left-radius: 0;
 }
 
+.top-bar__project-button-hidden {
```
outoftime: Can just use `u__invisible`?

src/css/application.css
```diff
@@ -315,6 +315,20 @@ body {
...
+  visibility: hidden;
+}
+
+.top-bar__project-saved {
+  position: absolute;
+  visibility: hidden;
+  left: 1em;
+}
+
+.top-bar__project-saved-shown {
```
outoftime: Should be `top-bar__project-saved_shown` (single underscore delimits a modifier)

src/css/application.css
```diff
@@ -315,6 +315,20 @@ body {
...
 }
 
+.top-bar__project-button-hidden {
+  visibility: hidden;
+}
+
+.top-bar__project-saved {
+  position: absolute;
+  visibility: hidden;
+  left: 1em;
```
outoftime: Why not just do `top`/`left`/`bottom`/`right` all at `0` to pin it to the container dimensions, and then center the text?

src/sagas/projects.js
```diff
@@ -81,23 +84,20 @@ export function* importGist({payload: {gistId}}) {
...
 export function* updateProjectSource() {
-  const state = yield select();
-  yield call(saveCurrentProject, state);
+  yield* saveCurrentProject();
 }
 
 export function* userAuthenticated() {
   const state = yield select();
-  yield fork(saveCurrentProject, state);
+  yield* fork(saveCurrentProject);
```
outoftime: As I understand it, `yield*` hands off control of the generator to another generator, but the `fork` effect is not itself a generator, so I don’t think it does anything here. For clarity I believe it should be changed back to a simple `yield`.

src/sagas/projects.js
```diff
@@ -107,11 +107,27 @@ function generateProjectKey() {
...
+      yield call(saveProject, userId, currentProject);
+      yield put(projectSuccessfullySaved());
+    } catch (e) {
+      return false;
+    }
+  }
+  return false;
+}
+
+
```
outoftime: Extra whitespace

src/sagas/projects.js
```diff
@@ -107,11 +107,27 @@ function generateProjectKey() {
...
-    const state = yield select();
-    yield call(saveCurrentProject, state);
+    yield* saveCurrentProject();
   }
 }
 
+function* saveCurrentProject() {
+  const state = yield select();
+  const userId = getCurrentUserId(state);
+  const currentProject = getCurrentProject(state);
```
outoftime: Arguably it would be cleaner to do:

```js
const userId = yield select(getCurrentUserId);
const currentProject = yield select(getCurrentProject);
```

src/sagas/projects.js
```diff
@@ -107,11 +107,27 @@ function generateProjectKey() {
...
+
+  if (userId && currentProject && !isPristineProject(currentProject)) {
+    try {
+      yield call(saveProject, userId, currentProject);
+      yield put(projectSuccessfullySaved());
+    } catch (e) {
+      return false;
```
outoftime: I don’t think these `return false`s are meaningful in the context of a saga? Or are there semantics I’m unaware of?

Assuming they aren’t, there is a good question regarding what _should_ go here in the absence of proper error handling (which as previously discussed would make sense to come in a different PR).

Bad pitch, maybe just remove the `try/catch` altogether and let it fail / generate a Bugsnag error?

src/selectors/isSaveIndicatorShown.js
```diff
@@ -0,0 +1,6 @@
+import {createSelector} from 'reselect';
+
+export default createSelector(
+  state => state.getIn(['ui', 'saveIndicatorShown']),
+  saveIndicatorShown => saveIndicatorShown,
+);
```
outoftime: So there is no requirement that we use `reselect` for every selector—it’s useful when a selector is doing non-trivial computation, since it’s defined as a two-stage process of “cheaply get some source data and see if it’s changed” followed by “expensively do some calculations on that source data”.

But in this case, there’s no calculation at all! It’s just an identity function.

So, more simply:

```js
export default function isSaveIndicatorShown(state) {
  return state.getIn(['ui', 'saveIndicatorShown']);
}
```

test/unit/sagas/projects.js
```diff
@@ -226,8 +228,7 @@ test('updateProjectSource', (assert) => {
     updateProjectSource(scenario.projectKey, 'css', 'p {}'),
   ).
     next().select().
-    next(scenario.state).call(saveCurrentProject, scenario.state).
-    next().isDone();
+    next(scenario.state).isDone();
```
outoftime: So it seems like these tests aren’t really making as meaningful assertions about saga behavior as they were before—I’m guessing because we have now internalized the check for `isPristineProject` which, at a guess, is not the case in the “scenario” that’s been set up here? Assuming that guess is correct, should be pretty straightforward to fix in the scenario file.

I assume this is what you meant by “trouble creating tests for the projects saga”?

pwjablonski:
@outoftime almost there. I'm still not sure if I am approaching the tests correctly - particularly around forking the delegated `saveCurrentProject` saga.

src/css/application.css
outoftime: Looks like you did make that change but this class name is still hanging around?

src/components/TopBar/ProjectPickerButton.jsx
```diff
@@ -1,16 +1,37 @@
...
+        <span className="u__icon top-bar__drop-down-button">
+          &#xf0d7;
+        </span>
       </span>
-    </span>
+      <span
+        className={
+          classnames(
+            'top-bar__project-saved',
+            {'top-bar__project-saved_shown': shouldShowSavedIndicator})
```
outoftime: Could just do `{u__invisible: !shouldShowSavedIndicator}` here too

Issue #1478, opened 2018-05-19
Unable to export/update repo when Instructions panel is opened


jwang1919:
Adding instructions and having the instructions panel open throws a visible "Something went wrong trying to export that repo. Please try again." error.

Interestingly, if the instructions panel is closed, the export/update repo functionality works fine.

Ex: https://popcode.org/?snapshot=79aadac1-2e26-4e1d-8757-0f745fd5ba18

alexpelan:
@jwang1919 can you still reproduce this? I cannot. The issue still persists with the snapshot above, but I changed snapshots to no longer save references to the saved github repos, so I would be interested to see if you could make it happen with a new repo, or whether it was fixed by that change.

jwang1919:
Yep can confirm the bug is fixed now!

PR #1482, opened 2018-05-20
Store exported repos by username 

alexpelan:
Instead of storing a single string for the reponame, store strings keyed by github username to allow multiple people to export the same project (importing snapshot workflows, for example).

One high level comment: this is backwards incompatible - any repos exported since release will be orphaned (update repo will appear as export repo). This does not seem like a big deal to me but I felt it should be called out.

Line-level comments below

### DIFF
--- a/src/clients/github.js
+++ b/src/clients/github.js
@@ -25,8 +25,12 @@ function normalizeTitle(title) {
   return titleWithoutPunctuationAndWhitespace;
 }
 
+function getRepoNameForUser(project, user) {
+  return project.externalLocations.githubRepos[user.githubUsername];
+}
+
 export async function createOrUpdateRepoFromProject(project, user) {
-  const repoAlreadyExists = Boolean(project.externalLocations.githubRepoName);
+  const repoAlreadyExists = Boolean(getRepoNameForUser(project, user));
   if (repoAlreadyExists) {
     return updateRepoFromProject(project, user);
   }
@@ -65,7 +69,7 @@ async function createRepoFromProject(project, user) {
 
 export async function updateRepoFromProject(project, user) {
   const github = await clientForUser(user);
-  const repoName = project.externalLocations.githubRepoName;
+  const repoName = getRepoNameForUser(project, user);
 
   const {data: userData} = await performWithRetryNetworkErrors(
     () => github.getUser().getProfile(),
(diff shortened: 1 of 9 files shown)


src/clients/github.js
```diff
@@ -25,8 +25,12 @@ function normalizeTitle(title) {
   return titleWithoutPunctuationAndWhitespace;
 }
 
+function getRepoNameForUser(project, user) {
+  return project.externalLocations.githubRepos[user.githubUsername];
```
alexpelan: This function is basically the same as the selector `getCurrentProjectExportedRepoName` except we don't (and shouldn't) have access to the whole store here. Is there any literature on sharing these business-logicky things between redux and non-redux contexts

test/unit/sagas/clients.js
```diff
@@ -119,7 +119,7 @@ test('export repo', (t) => {
         url,
         exportType,
         scenario.project.projectKey,
-        {name},
+        {repoName: name, username: undefined},
```
alexpelan: Not great, but the username is determined inside a function from the state and I couldn't (quickly) figure out a way to inject it. This at least tests that the key is correctly produced 🤷‍♀️

alexpelan:
Oh, also, unclear if this fixes #1478, it may have a secondary cause,  but this bug was the reason that I couldn't export the repo from that snapshot when I was trying to reproduce.

src/reducers/user.js
```diff
@@ -11,19 +11,18 @@ function user(stateIn, action) {
       const {user: userData, credential, additionalUserInfo} = action.payload;
 
       const profileData = get(userData, ['providerData', 0], userData);
+      const githubUsername = get(additionalUserInfo, 'username');
```
alexpelan: unclear how this will interact with google auth.

alexpelan:
So the moment that a user clicks create snapshot, we would wipe out the associated repo name? Or would this happen during the import process? Feel free to move this to Popcode slack #dev channel if it needs more discussion

outoftime:
@alexpelan when we export snapshots we create a copy of the current project data, so it’s not a matter of _removing_ the repo name from the project as stored in the user’s workspace, just omitting that information from the copy of the project that goes into the snapshot. probably makes sense either in the [firebase client](https://github.com/popcodeorg/popcode/blob/master/src/clients/firebase.js#L34) or in the saga that calls it.

alexpelan:
Ditching this PR in favor of one where we do that, then

PR #1485, opened 2018-05-21
Archive projects 

### LINKED ISSUE #1136: Archive projects
Add the ability to archive projects, which hides them from the **My Projects** list but does not actually delete them. Gives coders the ability to keep the projects list organized as they see fit.

Specifications:

* Next to each project title, add an archive icon.
* Clicking the archive icon archives the project. Archived projects are not visible in the **My Projects** list by default.
* However, the **My Projects** interface does expose a “View Archived Projects” button if there are any archived projects. If this button is clicked, archived projects become visible in the list.
* When archived projects are visible, they have an un-archive icon in place of the archive icon. Clicking this restores them to default visibility.

---

#### Original description:

Is it possible to remove projects after they are created?

Perhaps we can have a little x to the right of the project that will ask are they sure they want to delete and then removes the project

screen shot 2017-10-16 at 6 05 16 pm

pwjablonski:
Here's another one! My main focus is still on google auth but I couldn't resist!

Closes #1136

This PR attempts to add archiving functionality to projects dropdown. 

Functionality
- Clicking on the archive icon on a project menu item archives the project. 
- Clicking view archived projects button at the end of the project list shows a list of all archived projects. 
- Clicking on the up-arrow icon on a project menu item in the archived list unarchives the project and adds it to the active project list. 

Notes/ Questions
- Archive icon doesn't quite look right. Should we replace with an X or make it bigger?
- Is there a better way to filter the projects? Should filtering be done in a selector?
- The view archived projects button at the end of the list felt natural. However, this may be hard to navigate to with a lot of unarchived projects. Alternatively, we could add the button to the hamburger menu and pop up a modal with archived projects.
- We may want to change the color of view archived projects button
- Still needs to remove/set archived to false on project export. 

may-20-2018 20-03-45


outoftime:
@pwjablonski OK! HMU on Slack whenever!

src/components/ProjectPreview.jsx
```diff
@@ -1,14 +1,50 @@
...
+            >
+              <span className="u__icon">&#xf0aa;</span>
+            </div>
+          ) :
+          (
+            <div
+              className="project-preview__archive"
+              onClick={(e) => {
+                e.stopPropagation();
+                onProjectArchived();
```
outoftime: Probably not worth worrying about in this PR, but perhaps for a followup—the disappearance of the project from the list when you click the unarchive button feels a bit abrupt—would be nice to get a fade and/or slide animation popping. (This also might fall in the set of things that are way more annoying in a strict Redux-based state architecture, although I would be surprised if there weren’t some good libraries for abstracting this away at the React level)

src/components/TopBar/ProjectPicker.jsx
```diff
@@ -21,22 +21,69 @@ const ProjectPicker = createMenu({
...
+        onClick={onToggleViewArchived}
       >
-        <ProjectPreview projectKey={projectKey} />
-      </MenuItem>
-    ));
+
+        {archivedViewOpen ?
+          <div>
+            Hide Archived Projects
+            <span className="u__icon"> &#xf077; </span>
```
outoftime: I’m not loving the up/down chevron here—probably would be fine with just the words.

src/components/TopBar/ProjectPicker.jsx
```diff
@@ -21,22 +21,69 @@ const ProjectPicker = createMenu({
...
+        key="archivedProjectButton"
+        onClick={onToggleViewArchived}
       >
-        <ProjectPreview projectKey={projectKey} />
-      </MenuItem>
-    ));
+
+        {archivedViewOpen ?
+          <div>
+            Hide Archived Projects
```
outoftime: Should be `i18n`’d!

src/components/TopBar/ProjectPicker.jsx
```diff
@@ -21,22 +21,69 @@ const ProjectPicker = createMenu({
...
+        }
+      </MenuItem>,
+    );
+
+    return items;
   },
 })(ProjectPickerButton);
 
 ProjectPicker.propTypes = {
+  archivedViewOpen: PropTypes.bool.isRequired,
```
outoftime: The name suggests that there is a discrete archived view that may or may not be visible—I would call this something like `shouldShowArchivedProjects`

src/records/Project.js
```diff
@@ -20,6 +20,7 @@ export default class Project extends Record({
   updatedAt: null,
   instructions: '',
   externalLocations: new ExternalLocations(),
+  archived: false,
```
outoftime: Prefer `isArchived` for boolean properties

src/util/projectUtils.js
```diff
@@ -24,3 +28,13 @@ export function saveCurrentProject(state) {
 
   return false;
 }
+
+export function saveProjectWithKey(state, projectKey) {
+  const userId = getCurrentUserId(state);
+  const project = getProject(state, {projectKey});
+  if (userId) {
```
outoftime: Seems like this could also check for pristine-ness? In general this method looks to be 90% the same logic as `saveCurrentProject` so probably best to abstract out the shared logic?

src/sagas/projects.js
```diff
@@ -112,6 +114,23 @@ export function* projectExported({payload: {exportType}}) {
   }
 }
 
+export function* archiveProject({payload: {projectKey}}) {
+  const state = yield select();
+  const currentProjectKey = getCurrentProjectKey(state);
+  if (projectKey === currentProjectKey) {
+    yield put(createNewProject());
```
outoftime: Would it be simpler to just not show the archive button on the current project? Or would that be confusing from a UI standpoint?

src/sagas/projects.js
```diff
@@ -112,6 +114,23 @@ export function* projectExported({payload: {exportType}}) {
...
+  }
+
+  yield call(saveProjectWithKey, state, projectKey);
+}
+
+export function* unArchiveProject({payload: {projectKey}}) {
+  const state = yield select();
+  yield put(changeProject(projectKey));
```
outoftime: This is an interesting case that pushes us right to the limits of the “actions should say what happened” rule. A couple of corollaries of that rule are:

* We should never dispatch two actions from the same execution point
* A saga should never dispatch (`put`) an action merely because it consumed a different action

The latter corollary is violated here.

In the context of my proposal to dispatch `changeCurrentProject` rather than `unArchiveProject` (and probably to remove the `unArchiveProject` action/button altogether), the “what happened” would be simply “the user clicked a project with the intent to switch to it”. Then the system would react to it by:

* Changing the current project (`currentProject` reducer)
* Un-archiving the project (`projects` reducer)
* Persisting the project (`projects` saga, here)

What complicates this is that we really only need to persist the project if we’re switching to a project that was previously archived. However, the way that `redux-saga` works is that sagas’ `take` effects always consume actions _after_ the reducers have done so. So, there’s no straightforward way to know whether the project had previously been archived by the time this code is running.

I can think of a few potential ways around the issue:

1. Create a new action along the lines of `CHANGE_AND_UNARCHIVE_CURRENT_PROJECT`, which the system handles appropriately (by doing all the things it would do for both `CHANGE_CURRENT_PROJECT` and `UNARCHIVE_PROJECT`)
2. Pass extra information in the payload of `CHANGE_CURRENT_PROJECT` to indicate that the project had been archived
3. Always persist the current project when switching to it (unless it is pristine)

My strong preference is for the third option, which I don’t believe has any downsides from a behavior standpoint—at worst we’ll simply be saving data to Firebase that’s identical to the data that’s already been saved.

That said, it is something of a lucky coincidence that we can be insensitive to the previous state of the store when deciding how the saga should behave; it’s easy to imagine a situation in which we really would need to vary behavior based on the previous state. Probably a cross-that-bridge-when-we-get-there situation, but I’m interested to hear your thoughts!

src/components/ProjectPreview.jsx
```diff
@@ -1,14 +1,50 @@
...
+          (
+            <div
+              className="project-preview__archive"
+              onClick={(e) => {
+                e.stopPropagation();
+                onProjectUnArchived();
+              }}
+            >
+              <span className="u__icon">&#xf0aa;</span>
+            </div>
```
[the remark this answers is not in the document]
pwjablonski: I am not opposed to getting rid of the unarchive button completely though I think there needs to be some other more distinct visual indicator that the project is archived besides just the current italics. Perhaps some muted text on the right that says _Archived_

src/components/TopBar/ProjectPicker.jsx
```diff
@@ -21,22 +21,69 @@ const ProjectPicker = createMenu({
...
+          >
+            <ProjectPreview projectKey={project.projectKey} />
+          </MenuItem>,
+        );
+      } else if (archivedViewOpen && project.archived) {
+        projectItems.push(
+          <MenuItem
+            isActive={project.projectKey === currentProjectKey}
+            key={project.projectKey}
+            onClick={partial(onProjectUnArchived, project.projectKey)}
```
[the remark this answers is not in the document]
pwjablonski: Indeed I was toggling between which way to go with this. This makes sense

src/components/TopBar/ProjectPicker.jsx
```diff
@@ -21,22 +21,69 @@ const ProjectPicker = createMenu({
...
+      }
+      return projectItems;
+    }, []);
+
+    items.push(
       <MenuItem
-        isActive={projectKey === currentProjectKey}
-        key={projectKey}
-        onClick={partial(onChangeCurrentProject, projectKey)}
+        isSticky
```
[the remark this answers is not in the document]
pwjablonski: Yeah! True it is not quite a menu item and would be clearer to separate out.

src/components/TopBar/ProjectPicker.jsx
```diff
@@ -21,22 +21,69 @@ const ProjectPicker = createMenu({
...
+            onClick={partial(onProjectUnArchived, project.projectKey)}
+          >
+            <ProjectPreview projectKey={project.projectKey} />
+          </MenuItem>,
+        );
+      }
+      return projectItems;
+    }, []);
+
+    items.push(
```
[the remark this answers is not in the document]
pwjablonski: Yes that is a nice addition

src/components/ProjectPreview.jsx
pwjablonski: Will look into this!

src/components/TopBar/ProjectPicker.jsx
pwjablonski: Gotcha I don't have a strong attachment to it

src/components/TopBar/ProjectPicker.jsx
pwjablonski: Yup!

src/components/TopBar/ProjectPicker.jsx
pwjablonski: Makes sense

src/records/Project.js
pwjablonski: Yup!

src/util/projectUtils.js
pwjablonski: Yes indeed!

src/sagas/projects.js
pwjablonski: Would be a bit easier to implement. I will try it out to not show the archive button on the current project

pwjablonski:
Forget to update tests

pwjablonski:
@outoftime just going to update the tests but yes ready for another look

src/components/TopBar/createMenu.jsx
```diff
@@ -37,6 +37,7 @@ MenuItem.propTypes = {
 MenuItem.defaultProps = {
   isActive: false,
   isDisabled: false,
+  isSticky: false,
```
pwjablonski: Needs to come out

pwjablonski:
@outoftime likely will have some conflicts with the `saved-indicator` PR but good for another review.

outoftime:
@pwjablonski you called it! ping me when you’ve got the conflicts resolved so I can review the final product?

pwjablonski:
@outoftime 

Bringing this up to date.  I think this is up for another round. 

I saw some talk of potentially extracting the project list into a modal. Do you think it makes sense pause this and to tack that route?

pwjablonski:
Oops got used to closing PRs recently :) Back open!

pwjablonski:
Fixed UI tests :)

src/components/ProjectPreview.jsx
```diff
@@ -1,14 +1,55 @@
...
+            'project-preview__label_archived': project.isArchived,
+          })}
+        >
+          {preview.slice(0, MAX_LENGTH)}
+        </div>
+        {
+          (function showArchived() {
+            if (isCurrentProject) {
+              return null;
+            } else if (project.isArchived) {
```
outoftime: I would expect the linter to catch this, but no need for an `else` when the `if` branch returns.

src/components/TopBar/ProjectPicker.jsx
```diff
@@ -23,22 +26,58 @@ const ProjectPicker = createMenu({
...
+    currentProjectKey,
+    projects,
+    shouldShowArchivedProjects,
+    onChangeCurrentProject,
+    onToggleViewArchived,
+  }) {
+    const visibleProjects = shouldShowArchivedProjects ?
+      projects :
+      filter(projects, ({isArchived}) => !isArchived);
+    const items = map(visibleProjects, project => (
```
outoftime: Seems like we're still only using `projectKey` in the loop—let's destructure the parameter to reduce duplication?

src/sagas/projects.js
```diff
@@ -134,15 +138,31 @@ export function* projectExported({payload: {exportType}}) {
...
   }
 }
 
+export function* archiveProject({payload: {projectKey}}) {
+  yield* saveProjectWithKey(projectKey);
+}
+
+export function* unArchiveProject({payload: {projectKey}}) {
+  const state = yield select();
+  yield call(saveProjectWithKey, state, projectKey);
```
outoftime: It doesn’t look like `saveProjectWithKey` takes `state` as a parameter?

outoftime: Also, the `yield` approach in `archiveProject` and the `call` approach here are both basically fine, but it seems like we should be consistent about it?

test/unit/reducers/projects.js
```diff
@@ -87,7 +88,11 @@ test('changeCurrentProject', (t) => {
     reducer,
     initProjects({1: true, 2: false}),
     partial(changeCurrentProject, '2'),
-    initProjects({1: true, 2: false}),
+    initProjects({1: true, 2: false}).
+      update(
+        '2',
+        editedProject => editedProject.set('isArchived', false),
```
outoftime: So I believe the idea here is to test the behavior of projects getting unarchived when you switch to them, BUT:

1. It’s not immediately clear from looking at this that that is the behavior under test
2. I’m not even sure that that behavior _is_ being tested? It seems like in the initial state both projects are not archived, so the assertion here would pass even if the reducer side of things had not been implemented (unless I am missing something)

I think the immediate fix for both issues would be a separate test that explicitly sets the initial state of the project to archived, and then explicitly expects the output state to be un-archived.

More broadly this gets right to the heart of my critique of the way our reducer tests are structured (my fault!)—because we make assertions about the entire output state, rather than making discrete assertions about the things in the state we actually care about for the given test, it’s quite difficult to divine the intent of the test a lot of the time (and many tests are implicitly testing several behaviors, which is undesirable for a unit test)

src/components/TopBar/ProjectPicker.jsx
```diff
@@ -33,12 +46,37 @@ const ProjectPicker = createMenu({
         <ProjectPreview projectKey={projectKey} />
       </MenuItem>
     ));
+
+    if (!some(projects, 'isArchived')) {
```
outoftime: I think this is backwards—we want to show the archived view toggle if some of the projects _are_ archived.

src/sagas/projects.js
```diff
@@ -134,15 +138,30 @@ export function* projectExported({payload: {exportType}}) {
...
+    yield* saveProjectWithKey(currentProject.projectKey);
+  }
+}
 
-  if (userId && currentProject && !isPristineProject(currentProject)) {
-    yield call(saveProject, userId, currentProject);
+function* saveProjectWithKey(projectKey) {
+  const state = yield select();
+  const userId = yield select(getCurrentUserId);
+  const project = yield call(getProject, state, {projectKey});
```
outoftime: It’s a little odd to call `select` twice, once with a selector and once for the whole state. Either of these would be fine:

```js
const userId = yield select(getCurrentUserId);
const project = yield select(partialRight(getProject, {projectKey}));
```

or

```js
const state = yield select();
const userId = yield call(getCurrentUserId, state);
const project = yield call(getProject, state, {projectKey});
```

I would say I have a mild preference for the first one but honestly doesn’t matter (and I probably wouldn’t even bring this up if it were the only feedback on the PR to be honest!)

outoftime:
@pwjablonski everything looks good to my eye, but it does look like there are some tests that need to be updated…

outoftime:
@pwjablonski looks like the tests are still failing

PR #1489, opened 2018-05-25
Remove externalLocations from the project before exporting to snapshot

alexpelan:
The next user that imports it won't be able to export to repo since they don't
own it.

I look forward to the discussion as to whether this should be a whitelist or blacklist


outoftime:
@alexpelan LGTM! Can see arguments both ways, but this seems the most straightforward fix for the immediate problem so I’m inclined to keep this until future requirements make it clearer that a whitelist of preferable.

PR #1507, opened 2018-06-05
Regenerate clean yarn.lock file

outoftime:
`yarn.lock` has path-dependency based on the sequence of adding and upgrading package versions, which can cause transitive dependencies to be unnecessarily duplicated and/or out of date. Regenerating a fresh `yarn.lock` file allows Yarn to construct the most efficient dependency tree given the first-order dependency constraints.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8,47 +8,11 @@
   dependencies:
     "@babel/highlight" "7.0.0-beta.44"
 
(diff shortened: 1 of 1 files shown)


outoftime:
The clean `yarn.lock` file results in an annoying peer dependency mismatch which will probably resolve itself next month. Not worth wrangling now.

PR #1511, opened 2018-06-10
Create classroom assignment

pwjablonski:
@outoftime Next Chunk of Classroom Assignment Project. 

This PR creates assignments (Draft and Active) on google classroom via classroom API. Assignments are created via Assignment Selector Modal. 

Let me know if the scope of this PR is too large and I can pare it down to just creating a draft assignment on button click.


src/sagas/ui.js
```diff
@@ -25,6 +30,29 @@ export function* projectSuccessfullySaved() {
   yield put(hideSaveIndicator());
 }
 
+export function* assignmentCreated(action) {
```
outoftime: This seems like another place where we are using a saga but don’t need to be?

src/clients/googleClassroom.js
```diff
@@ -8,3 +12,60 @@ export function createShareToClassroomUrl(snapshotKey, title) {
   uri.search = `snapshot=${snapshotKey}`;
   return BASE_URL + qs.stringify({url: uri.href, title});
 }
+
+export function createSnapshotUrl(snapshotKey) {
```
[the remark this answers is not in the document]
pwjablonski: I don't think this function exist anywhere else. The same thing happens in `SnapshotNotification.jsx` but not in a function. Perhaps we should extract into a separate module in the util folder and use in both places

src/actions/ui.js
```diff
@@ -76,3 +76,37 @@ export const showSaveIndicator = createAction(
...
+  'COURSE_SELECTED',
+);
+
+export const updateCourses = createAction(
+  'UPDATE_COURSES',
+  courses => ({courses}),
+);
+
+export const updateDate = createAction(
+  'UPDATE_DATE',
```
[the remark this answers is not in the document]
pwjablonski: Yes - definitely needs more specificity

src/clients/googleClassroom.js
```diff
@@ -8,3 +12,60 @@ export function createShareToClassroomUrl(snapshotKey, title) {
...
+
+export function createAssignmentUrl(assignemtnKey) {
+  const uri = document.createElement('a');
+  uri.href = '/';
+  uri.search = `assignment=${assignemtnKey}`;
+  return uri.href;
+}
+
+export async function getCourses() {
+  const gapi = await getGapiSync();
```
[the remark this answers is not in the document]
pwjablonski: I should get rid of the `await` I believe?

src/actions/assignments.js
```diff
@@ -0,0 +1,27 @@
...
+);
+
+export const assignmentDisplayed = createAction(
+  'ASSIGNMENT_DISPLAYED',
+  assignment => (assignment),
+);
+
+export const assignmentCreated = createAction(
+  'ASSIGNMENT_CREATED',
+  assignment => (assignment),
```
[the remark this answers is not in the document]
pwjablonski: Yes I meant to wrap in a an object

src/clients/googleClassroom.js
```diff
@@ -8,3 +12,60 @@ export function createShareToClassroomUrl(snapshotKey, title) {
...
+}
+
+export async function createClassroomAssignment(
+  accessToken,
+  workType,
+  courseId,
+  selectedDate,
+  url,
+  title,
+  assignmentState,
```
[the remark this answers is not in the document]
pwjablonski: Yes much better

src/components/AssignmentSelector.jsx
```diff
@@ -0,0 +1,151 @@
...
+  }
+  const courseOptions = map(courses, course => (
+    <option key={course.id} value={course.id}>{course.name}</option>
+  ));
+  let parsedDateString;
+  if (parsedDate) {
+    parsedDateString = parsedDate.toLocaleTimeString(
+      'en-EN',
+      {weekday: 'long', year: 'numeric', month: 'long', day: 'numeric'},
+    );
```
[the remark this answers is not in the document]
pwjablonski: Yes down to extract.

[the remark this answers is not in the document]
pwjablonski: Can you clarify the meaning of (module-local)?

src/components/AssignmentSelector.jsx
```diff
@@ -0,0 +1,151 @@
...
+import partial from 'lodash-es/partial';
+
+function AssignmentSelector({
+  assignmentNotification,
+  courses,
+  currentProjectKey,
+  dateInput,
+  isOpen,
+  parsedDate,
+  preview,
```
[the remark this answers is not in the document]
pwjablonski: Indeed!

src/components/AssignmentSelector.jsx
```diff
@@ -0,0 +1,151 @@
...
+              {courseOptions}
+            </select>
+            <span className="u__icon assignment-selector__caret">
+              &#xf0d7;
+            </span>
+          </div>
+          <div className="assignment-selector__input_field">
+            <input
+              className="assignment-selector__input"
+              placeholder="when is it due?"
```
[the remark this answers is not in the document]
pwjablonski: Yes!

src/components/AssignmentSelector.jsx
```diff
@@ -0,0 +1,151 @@
...
+              (parsedDate < Date.now())},
+            )}
+          >
+            {parsedDateString}
+          </p>
+          <p className="assignment-selector__assignment-notifcation">
+            {assignmentNotification ?
+              t(`assignment-selector.${assignmentNotification}`) :
+              null}
+          </p>
```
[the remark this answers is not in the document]
pwjablonski: Gotcha!

src/components/AssignmentSelector.jsx
```diff
@@ -0,0 +1,151 @@
...
+          >
+            {t('assignment-selector.cancel-button')}
+          </button>
+          <button
+            className={classnames(
+              'assignment-selector__button',
+              'assignment-selector__button_confirm',
+            )}
+            type="button"
+            onClick={partial(onDraftAssignment,
```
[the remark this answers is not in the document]
pwjablonski: Yup

src/components/AssignmentSelector.jsx
```diff
@@ -0,0 +1,151 @@
...
+  onAssignAssignment: PropTypes.func.isRequired,
+  onCloseAssignmentSelector: PropTypes.func.isRequired,
+  onDraftAssignment: PropTypes.func.isRequired,
+  onHandleDateInput: PropTypes.func.isRequired,
+  onSelectCourse: PropTypes.func.isRequired,
+};
+
+AssignmentSelector.defaultProps = {
+  assignmentNotification: null,
+  currentProjectKey: null,
```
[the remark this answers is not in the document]
pwjablonski: Yeah, I think the case was before project load? I will double check this scenario.

src/components/Workspace.jsx
```diff
@@ -22,6 +23,7 @@ import Output from '../containers/Output';
 
 import PopThrobber from './PopThrobber';
 
+
```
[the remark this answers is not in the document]
pwjablonski: 👍

src/components/Workspace.jsx
```diff
@@ -161,6 +163,9 @@ export default class Workspace extends React.Component {
   render() {
     return (
       <div className="layout">
+        <AssignmentSelector
+          projectKey={get(this.props, ['currentProject', 'projectKey'])}
```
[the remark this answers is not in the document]
pwjablonski: ahh yes not sure I will remove

src/reducers/ui.js
```diff
@@ -24,7 +24,15 @@ const defaultState = new Immutable.Map().
   set('workspace', DEFAULT_WORKSPACE).
   set('notifications', new Immutable.Map()).
   set('topBar', new Immutable.Map({openMenu: null})).
-  set('saveIndicatorShown', false);
+  set('saveIndicatorShown', false).
+  set('assignmentSelector', new Immutable.Map({
```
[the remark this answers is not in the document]
pwjablonski: Yes this makes sense!

src/reducers/ui.js
```diff
@@ -24,7 +24,15 @@ const defaultState = new Immutable.Map().
...
+    openModal: false,
+    selectedCourse: '',
+    courses: new Immutable.List(),
+    dateInput: '',
```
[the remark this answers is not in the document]
pwjablonski: My thought in having both was since the date input field is a controlled component. Will look at your proposal re forms below.

src/reducers/ui.js
```diff
@@ -24,7 +24,15 @@ const defaultState = new Immutable.Map().
   set('workspace', DEFAULT_WORKSPACE).
   set('notifications', new Immutable.Map()).
   set('topBar', new Immutable.Map({openMenu: null})).
-  set('saveIndicatorShown', false);
+  set('saveIndicatorShown', false).
+  set('assignmentSelector', new Immutable.Map({
+    openModal: false,
+    selectedCourse: '',
+    courses: new Immutable.List(),
```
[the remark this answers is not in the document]
pwjablonski: Yes

src/sagas/assignments.js
```diff
@@ -0,0 +1,86 @@
...
+  updateCourses,
+  assignmentNotificationCreated,
+} from '../actions/ui';
+import {
+  getCurrentProject,
+} from '../selectors';
+import {generateTextPreview} from '../util/compileProject';
+
+export function* openAssignmentSelector() {
+  const courses = yield call(getCourses);
```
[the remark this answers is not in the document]
pwjablonski: Indeed!

src/sagas/assignments.js
```diff
@@ -0,0 +1,86 @@
...
+export function* draftAssignment(
+  {payload: {projectKey, selectedCourse, selectedDate}}) {
+  yield* createAssignment(projectKey, selectedCourse, selectedDate, 'DRAFT');
+}
+
+function* createAssignment(
+  projectKey, selectedCourse, selectedDate, assignmentState) {
+  const state = yield select();
+  const user = state.get('user').toJS();
+  const project = getCurrentProject(state);
```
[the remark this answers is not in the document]
pwjablonski: Yup!

src/sagas/assignments.js
```diff
@@ -0,0 +1,86 @@
...
+        user.accessTokens['google.com'],
+        'ASSIGNMENT',
+        selectedCourse,
+        selectedDate,
+        snapshotUrl,
+        projectTitle,
+        assignmentState,
+      );
+      yield put(assignmentCreated(assignment));
+      yield put(assignmentSelectorClosed());
```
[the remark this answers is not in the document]
pwjablonski: Gotcha!

src/clients/googleClassroom.js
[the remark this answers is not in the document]
outoftime: Yep, sounds right

src/clients/googleClassroom.js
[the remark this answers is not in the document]
outoftime: Actually I think you should call `await loadAndConfigureGapi()`—it is a fundamentally async operation. I look at `getGapiSync()` as basically a workaround for certain browser restrictions, so we should only use it where we need that workaround.

src/components/AssignmentSelector.jsx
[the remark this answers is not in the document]
outoftime: In other words a function that lives in this file, that isn’t exported.

src/reducers/ui.js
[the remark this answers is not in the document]
outoftime: Yeah that makes total sense, I think it would also scan fine in the context of a clean separation between core assignment data and form properties as outlined above.

pwjablonski:
@outoftime Hey finally got some time to make progress on this. 

Still very rough but welcome feedback on the approach.

1. Redux form and immutable
2. Structure of the AssignmentCreationInterface
3. Validation in general
4. Structure of the saga

src/components/TopBar/index.jsx
```diff
@@ -153,6 +157,7 @@ TopBar.propTypes = {
   isUserAnonymous: PropTypes.bool.isRequired,
   isUserAuthenticated: PropTypes.bool.isRequired,
   isUserAuthenticatedWithGithub: PropTypes.bool.isRequired,
+  isUserAuthenticatedWithGoogle: PropTypes.bool.isRequired,
```
pwjablonski: Legacy?

src/components/TopBar/index.jsx
```diff
@@ -91,10 +93,12 @@ export default function TopBar({
         isOpen={openMenu === 'export'}
         isRepoExportInProgress={isRepoExportInProgress}
         isUserAuthenticatedWithGithub={isUserAuthenticatedWithGithub}
+        isUserAuthenticatedWithGoogle={isUserAuthenticatedWithGoogle}
```
pwjablonski: Legacy?

src/components/renderField.jsx
```diff
@@ -0,0 +1,29 @@
...
+    <div>
+      <div className="assignment-creator__input_field">
+        <input
+          className="assignment-creator__input"
+          {...props.input}
+          placeholder={props.placeholder}
+        />
+      </div>
+      {
+        (props.meta.touched && (props.meta.error &&
```
pwjablonski: This is ugly

src/components/renderField.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function renderField(props) {
```
pwjablonski: Separate out props

src/containers/AssignmentCreator.js
```diff
@@ -0,0 +1,75 @@
...
+} from '../actions/ui';
+import {
+  assignAssignment,
+  draftAssignment,
+} from '../actions/assignments';
+import {
+  getAssignmentNotification,
+  getCourses,
+  isAssignmentCreatorOpen,
+  getSelectedCourse,
```
pwjablonski: These are left over

src/containers/AssignmentCreator.js
```diff
@@ -0,0 +1,75 @@
...
+
+
+function makeMapStateToProps() {
+  return function mapStateToProps(state) {
+    return {
+      assignmentNotification: getAssignmentNotification(state),
+      courses: getCourses(state),
+      dateInput: getDateInput(state),
+      isOpen: isAssignmentCreatorOpen(state),
+      parsedDate: getParsedDate(state),
```
pwjablonski: left over

src/containers/AssignmentCreator.js
```diff
@@ -0,0 +1,75 @@
...
+    onDraftAssignment(data) {
+      const selectedCourse = data.get('course');
+      const dateInput = data.get('date');
+      dispatch(draftAssignment(
+        selectedCourse,
+        dateInput,
+      ));
+    },
+
+    onSelectCourse(coursedId) {
```
pwjablonski: same left over

src/containers/TopBar.js
```diff
@@ -21,16 +21,19 @@ import {
   isTextSizeLarge,
   isUserAnonymous,
   isUserAuthenticatedWithGithub,
+  isUserAuthenticatedWithGoogle,
```
pwjablonski: left over

src/sagas/assignments.js
```diff
@@ -0,0 +1,71 @@
...
+  const courses = yield call(getCourses);
+  yield put(updateCourses(courses));
+}
+
+export function* assignAssignment(
+  {payload: {selectedCourse, dateInput}}) {
+  yield* createAssignment(
+    selectedCourse,
+    dateInput,
+    'PUBLISHED',
```
pwjablonski: Should we only use the createassignment function and add the assignment type in when the action is dispatched?

src/sagas/ui.js
```diff
@@ -25,6 +30,29 @@ export function* projectSuccessfullySaved() {
...
+    }));
+  } else {
+    yield put(assignmentDisplayed({
+      url: 'https://classroom.google.com/',
+      exportType: 'assignment-draft',
+    }));
+  }
+}
+
+export function* updateDate(action) {
```
pwjablonski: left over

src/selectors/getDateInput.js
```diff
@@ -0,0 +1,5 @@
+export default function getDateInput(state) {
```
pwjablonski: left over

src/selectors/getParsedDate.js
```diff
@@ -0,0 +1,5 @@
+export default function getParsedDate(state) {
```
pwjablonski: left over

src/selectors/getSelectedCourse.js
```diff
@@ -0,0 +1,5 @@
+export default function getSelectedCourse(state) {
```
pwjablonski: left over

src/selectors/index.js
```diff
@@ -39,13 +39,22 @@ import isUserAuthenticatedWithGithub from './isUserAuthenticatedWithGithub';
 import isUserAuthenticatedWithGoogle from './isUserAuthenticatedWithGoogle';
 import makeGetProjectPreview from './makeGetProjectPreview';
 import makeGetResizableFlexGrow from './makeGetResizableFlexGrow';
+import isAssignmentCreatorOpen from './isAssignmentCreatorOpen';
+import getCourses from './getCourses';
+import getSelectedCourse from './getSelectedCourse';
```
pwjablonski: remove left over

src/containers/AssignmentCreator.js
```diff
@@ -0,0 +1,75 @@
...
+      projectTitle: getCurrentProjectPreview(state),
+      selectedCourse: getSelectedCourse(state),
+    };
+  };
+}
+
+function mapDispatchToProps(dispatch) {
+  return {
+    onAssignAssignment(data) {
+      const selectedCourse = data.get('course');
```
pwjablonski: Form data is stored as imutable

src/reducers/index.js
```diff
@@ -1,5 +1,6 @@
 import {combineReducers} from 'redux-immutable';
 import reduceReducers from 'reduce-reducers';
+import {reducer as formReducer} from 'redux-form/immutable';
```
pwjablonski: When changing this to just 'redux-form' the form did not work

src/components/AssignmentCreatorForm.jsx
```diff
@@ -0,0 +1,129 @@
...
+function AssignmentCreatorForm({
+  courses,
+  handleSubmit,
+  onAssignAssignment,
+  onDraftAssignment,
+  onCloseAssignmentCreator,
+}) {
+  const courseOptions = map(courses, course => (
+    <option key={course.id} value={course.id}>{course.name}</option>
+  ));
```
outoftime: These days I am more into inlining everything into the JSX and avoiding intermediate variables, but I don’t feel strongly about it.

src/components/AssignmentCreatorForm.jsx
```diff
@@ -0,0 +1,129 @@
...
+function warn(values) {
+  const warnings = {};
+  const input = values.get('date');
+  const parsedResponse = chrono.parse(input);
+  if (parsedResponse[0]) {
+    const parsedDate = parsedResponse[0].start.date();
+    const dateString = dateToString(parsedDate);
+    warnings.date = dateString;
+  }
+  return warnings;
```
outoftime: OK so this is a clever use of `warnings`, but of course this is not _actually_ a warning so it does feel a little funny. I have a have/eat cake proposal with the caveat that I have much less experience (none) working with `redux-form` than you, so my apologies if this is unrealistic and/or you’ve already tried it.

Basic idea: The form should have both a `natural_language_date` field and a `date` field. The `natural_language_date` field holds the string that the user typed, and the `date` field holds the translated version.

The `natural_language_date` field would be rendered as it is now, a normal text input. The `date` field would be rendered as a `readonly` input, styled to not look like an input at all.

Using an `onChange` handler on the `natural_language_date` `<Field>` component, we’d register a callback that calls [`change()`](https://redux-form.com/7.4.2/docs/api/props.md/#-code-change-field-string-value-any-function-code-) to update the `date` field accordingly (i.e. using `chrono.parse`).

The advantage here is that, if I’m not mistaken, nothing else would need to worry about the `natural_language_date` input, `chrono.parse` etc.—everything else (including the validation etc.) could just look at the already-parsed value of `date`.

(Might be even slicker to let `date` itself flow through the `parse`/`format` flow, not sure if that would make the code simpler or more complicated though)

src/components/renderField.jsx
outoftime: Let’s name these components the same as we would name any other, e.g. this one should probably be called something like `<AssignmentCreatorTextField>` and the next one `<AssignmentCreatorSelectField>` etc.

src/components/AssignmentCreatorForm.jsx
```diff
@@ -0,0 +1,129 @@
...
+import renderField from './renderField';
+import renderSelectField from './renderSelectField';
+
+function validate(values) {
+  const errors = {};
+  const input = values.get('date');
+  const parsedResponse = chrono.parse(input);
+
+  if (!values.get('course')) {
+    errors.course = '* Course Required';
```
outoftime: Don’t forget to `i18next`!

src/clients/googleClassroom.js
```diff
@@ -8,3 +13,56 @@ export function createShareToClassroomUrl(snapshotKey, title) {
...
+  workType,
+  courseId,
+  dateInput,
+  url,
+  title,
+  state,
+}) {
+  const gapi = await loadAndConfigureGapi();
+
+  const parsedDate = chrono.parse(dateInput);
```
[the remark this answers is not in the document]
outoftime: (Update, I believe [this](https://github.com/popcodeorg/popcode/pull/1511#discussion_r214769201) would have that effect)

src/sagas/assignments.js
```diff
@@ -0,0 +1,71 @@
...
+export function* draftAssignment(
+  {payload: {selectedCourse, dateInput}}) {
+  yield* createAssignment(selectedCourse, dateInput, 'DRAFT');
+}
+
+function* createAssignment(selectedCourse, dateInput, assignmentState) {
+  const project = yield select(getCurrentProject);
+  const snapshotKey = yield call(createProjectSnapshot, project);
+  const assignmentData = {
+    workType: 'ASSIGNMENT',
```
outoftime: Seems like the knowledge of Google Classroom’s API is leaking out a bit here? I would expect the `googleClassroom` client to encapsulate things like `workType` and just provide an interface where we can pass in the basic facts of the assignment from the application’s perspective.

src/sagas/assignments.js
```diff
@@ -0,0 +1,71 @@
...
+    dateInput,
+    url: yield call(createSnapshotUrl, snapshotKey),
+    title: yield call(generateTextPreview, project),
+    state: assignmentState,
+  };
+
+  try {
+    const assignment = yield call(createClassroomAssignment, assignmentData);
+    yield put(assignmentCreated(assignment));
+    yield put(closeAssignmentCreator());
```
outoftime: If successful assignment creation should cause the assignment modal to close, the UI reducer should consume `ASSIGNMENT_CREATED` and update its state accordingly, no?

src/sagas/assignments.js
```diff
@@ -0,0 +1,71 @@
...
+  } catch (e) {
+    yield put(assignmentNotCreated());
+    yield put(closeAssignmentCreator());
```
outoftime: Ibid. double dispatch

src/sagas/ui.js
```diff
@@ -25,6 +28,20 @@ export function* projectSuccessfullySaved() {
...
+  if (assignment.alternateLink) {
+    yield put(assignmentDisplayed({
+      url: assignment.alternateLink,
+      exportType: 'assignment',
+    }));
+  } else {
+    yield put(assignmentDisplayed({
+      url: 'https://classroom.google.com/',
+      exportType: 'assignment-draft',
+    }));
```
outoftime: It seems like we’re just consuming an action and re-dispatching a new action? Why not just have the reducers that are currently consuming `ASSIGNMENT_DISPLAYED` instead consume `ASSIGNMENT_CREATED` directly?

src/sagas/ui.js
```diff
@@ -25,6 +28,20 @@ export function* projectSuccessfullySaved() {
...
 
+export function* assignmentCreated({payload: {assignment}}) {
+  if (assignment.alternateLink) {
+    yield put(assignmentDisplayed({
+      url: assignment.alternateLink,
+      exportType: 'assignment',
+    }));
+  } else {
+    yield put(assignmentDisplayed({
+      url: 'https://classroom.google.com/',
```
outoftime: Would it be possible to at least direct-link to the course in this case?

src/selectors/getCourses.js
```diff
@@ -0,0 +1,11 @@
+export default function getCourses(state) {
+  const courses = state.getIn([
+    'ui',
+    'assignmentCreationInterface',
+    'courses',
+  ]);
+  if (courses) {
+    return courses.toJS();
```
outoftime: `.toJS()` is considered harmful (main problem is it is slow) and I’m trying to remove everywhere we currently use it in the code base. Instead we should just make sure we are using `Record`s everywhere we should be (which as far as I can see we are in this PR) and then work with those directly in components etc. In many cases the interface is the same (you can do normal property access, `for…of` loops work with Immutable collections, etc.)

src/selectors/getCurrentProjectPreview.js
```diff
@@ -0,0 +1,8 @@
+import makeGetProjectPreview from './makeGetProjectPreview';
+import getCurrentProjectKey from './getCurrentProjectKey';
+
+export default function getCurrentProjectPreview(state) {
+  const getProjectPreview = makeGetProjectPreview();
```
outoftime: I don’t think we want to make a new `getProjectPreview()` function each time `getCurrentProjectPreview()` is called. Instead, I would move this line outside the function so that we are working with the same `getProjectPreview()` function each time.

Since in this case the argument we are passing to `getProjectPreview()` will be the current project key, having the single memoized selector will have the cache behavior we expect/want.

src/records/AssignmentCreationInterface.js
```diff
@@ -0,0 +1,6 @@
+import {Record, List} from 'immutable';
+
+export default Record({
+  openModal: false,
```
outoftime: This may change anyway based on the below, but I like to keep boolean properties/predicates prefixed with `is`, e.g. `isModalOpen` for this one.

src/selectors/makeGetProjectPreview.js
```diff
@@ -5,6 +5,11 @@ import {generateTextPreview} from '../util/compileProject';
 export default function makeGetProjectPreview() {
   return createSelector(
     (state, {projectKey}) => state.getIn(['projects', projectKey]),
-    project => generateTextPreview(project.toJS()),
+    (project) => {
+      if (project) {
```
outoftime: When do we need this check?

src/util/timeFormatter.js
```diff
@@ -0,0 +1,9 @@
+export function dateToString(date) {
+  if (date) {
+    return date.toLocaleTimeString(
+      'en-EN',
+      {weekday: 'long', year: 'numeric', month: 'long', day: 'numeric'},
+    );
+  }
+  return null;
+}
```
outoftime: Would be nice to fold this into the `i18next` fold, particularly so that if we ever do support other languages we won’t have to go rooting around too many places. Seems like i18n’s [`format`](https://www.i18next.com/translation-function/formatting) hook should work nicely?

yarn.lock
```diff
@@ -8545,6 +8559,10 @@ react-immutable-proptypes@^2.1.0:
   version "2.1.0"
   resolved "https://registry.yarnpkg.com/react-immutable-proptypes/-/react-immutable-proptypes-2.1.0.tgz#023d6f39bb15c97c071e9e60d00d136eac5fa0b4"
 
+react-lifecycles-compat@^3.0.4:
```
outoftime: Hopefully `redux-form` is working on porting to the new lifecycles…!

src/components/AssignmentCreatorForm.jsx
```diff
@@ -0,0 +1,129 @@
...
+}
+
+function AssignmentCreatorForm({
+  courses,
+  handleSubmit,
+  onAssignAssignment,
+  onDraftAssignment,
+  onCloseAssignmentCreator,
+}) {
+  const courseOptions = map(courses, course => (
```
outoftime: If I’m understanding the flow here, there is going to be a period in which the `courses` have not in fact loaded, during which the `courseOptions` will be empty (lodash `map` returns `[]` if passed `null` or `undefined).

Instead, I think we should display a loading spinner for the entire form until the courses have loaded.

src/components/AssignmentCreatorForm.jsx
```diff
@@ -0,0 +1,129 @@
+import React from 'react';
+import {t} from 'i18next';
+import classnames from 'classnames';
+import map from 'lodash-es/map';
+import chrono from 'chrono-node';
+
+import PropTypes from 'prop-types';
+import {Field, reduxForm} from 'redux-form/immutable';
```
outoftime: Is `redux-form` a big contributor to bundle size? If it’s at all significant I think we should load it asynchronously, especially since it’s only used by instructors.

The use of `react-coroutine` in [this (sadly unmerged) component](https://github.com/popcodeorg/popcode/blob/19a5f331c50c3c6075d39923821941e70cf39e5c/src/components/InstructionsEditorAsync.jsx) is a great way to accomplish this.

outoftime: For the avoidance of doubt, I _don’t_ like the way that the linked component has both an `<InstructionsEditorAsync>` and a separate `<InstructionsEditor>` component—I think it should just be one component called `<InstructionsEditor>`, which is asynchronous.

src/sagas/ui.js
pwjablonski: This is what the payload returned looks like for a draft assignment.
```
assigneeMode: "ALL_STUDENTS"
associatedWithDeveloper:true
courseId:"2022089786"
creationTime:"2018-09-09T16:51:45.383Z"
creatorUserId:"xxxxxxxxxxxxxxxxxxxxxxxx"
dueDate:{year: 2018, month: 9, day: 11}
dueTime:{hours: 16}
id:"17715559980"
materials:[{…}]
state:"DRAFT"
submissionModificationMode:"MODIFIABLE_UNTIL_TURNED_IN"
title:"Page Title"
updateTime:"2018-09-09T16:51:45.325Z"
workType:"ASSIGNMENT"
```
The payload has a courseID as a string of digits, but the url to direct to the classroom looks like this `https://classroom.google.com/u/0/c/MjAyMjA4OTc4Nlpa`
where the id is a string of letters of digit. I'm not quite sure if you can extract this URL with the information given.
My only other thought is that we might be able store the classroom url when retrieving the lists of courses and pass that url along when submiting the assignment. A bit annoying but may be worth it?

src/css/application.css
```diff
@@ -76,6 +76,7 @@
   --color-dark-gray: #444;
   --color-green: #00ff85;
   --color-low-contrast-green: #a2ffd2;
+  --color-dark-green: #27ae60;
```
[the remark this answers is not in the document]
pwjablonski: I plucked it out of thin air as placeholder. I felt the popcode green as the default background for the button was too bright with whit text and against the background of the modal. Open to suggestions.

src/records/AssignmentCreationInterface.js
```diff
@@ -0,0 +1,6 @@
...
+  courses: new List(),
```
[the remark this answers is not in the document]
pwjablonski: Definitely up for moving courses out of here and dropping the `isModalOpen` into the `UiState`. May rename it to isAssignmentCreatorModalOpen`. As for where to persist the courses, I am a bit torn. I do like keeping all of the classroom related data together `googleClassroom` but also being able to easily drop the courses on log out is convenient. I don't mind either way and am happy to defer to your judgment.

src/sagas/ui.js
outoftime: Mm, good call on just storing the course URL (weird that the ID and public-facing URL seem unrelated). Seems pretty straightforward since we are already storing course info? But if it’s annoying totally reasonable as a followup as well.

src/components/AssignmentCreatorForm.jsx
pwjablonski: Got it - how would I determine whether redux-form is a  big contributor to bundle size?

package.json
```diff
@@ -199,6 +199,7 @@
     "qs": "^6.1.0",
     "react": "^16.4.2",
     "react-copy-to-clipboard": "^5.0.0",
+    "react-coroutine": "^2.0.2",
```
pwjablonski: Oops this a hanger on- I was looking at async load

src/components/AssignmentCreatorForm.jsx
```diff
@@ -1,79 +1,88 @@
...
 import {Field, reduxForm} from 'redux-form/immutable';
 
-import {dateToString} from '../util/timeFormatter';
-
-import renderField from './renderField';
-import renderSelectField from './renderSelectField';
+import AssignmentCreatorTextField from './AssignmentCreatorTextField';
+import AssignmentCreatorSelectField from './AssignmentCreatorSelectField';
 
 function validate(values) {
```
pwjablonski: Should the validation (as well as `parse` and  `format`) remain in the `AssignmentCreatorForm` or should this be extracted out somewhere?

src/containers/AssignmentCreator.js
```diff
@@ -1,47 +1,48 @@
...
 function mapDispatchToProps(dispatch) {
   return {
     onAssignAssignment(data) {
-      const selectedCourse = data.get('course');
-      const dateInput = data.get('date');
-      dispatch(assignAssignment(
-        selectedCourse,
-        dateInput,
+      const selectedCourseId = data.get('course');
+      const dueDate = data.get('date').parsedDate;
```
pwjablonski: are these okay here?

src/css/application.css
```diff
@@ -42,7 +42,7 @@
   --color-dark-gray: #444;
   --color-green: #00ff85;
   --color-low-contrast-green: #a2ffd2;
-  --color-dark-green: #27ae60;
+  --color-dark-green: #00d16d;
```
pwjablonski: Used your suggestion

src/css/application.css
```diff
@@ -923,20 +923,20 @@ body {
 
 .assignment-creator__button_confirm {
   background-color: var(--color-dark-green);
-  color: var(--color-chrome);
+  color: var(--color-low-contrast-gray);
```
pwjablonski: Switched around the color scheme here but happy for input on what looks best

src/records/Course.js
```diff
@@ -0,0 +1,26 @@
...
+  ownerId: null,
+  creationTime: null,
+  updateTime: null,
+  enrollmentCode: null,
+  courseState: null,
+  alternateLink: null,
+  teacherGroupEmail: null,
+  courseGroupEmail: null,
+  guardiansEnabled: null,
+  calendarId: null,
```
pwjablonski: Include most of the info from the assignment that comes back in the payload except for these two. Give that we are only using a couple of these items I am wondering if it makes sense to store all of these.
```
"teacherFolder": {
    object(DriveFolder)
  },
  "courseMaterialSets": [
    {
      object(CourseMaterialSet)
    }
  ],
```

src/selectors/makeGetProjectPreview.js
pwjablonski: Seems to be catching on project load but haven't tracked it down entirely

src/components/AssignmentCreatorForm.jsx
```diff
@@ -0,0 +1,141 @@
...
+
+AssignmentCreatorForm.defaultProps = {
+  parsedDate: null,
+};
+
+
+export default reduxForm({
+  form: 'assignmentCreator',
+  validate,
+})(AssignmentCreatorForm);
```
outoftime: Would it make sense to make a `src/containers/AssignmentCreatorForm.js` that did the `reduxForm` work? I don’t see any reason containers specifically have to be using `react-redux`’s `connect()` directly—applying `reduxForm` seems analogous?

src/components/AssignmentCreatorTextField.jsx
```diff
@@ -0,0 +1,50 @@
...
+          placeholder={placeholder}
+        />
+      </div>
+      <div className="assignment-creator__input_warning">
+        <span>
+          {t('assignment-creator.date', {date: parsedDate})}
+        </span>
+      </div>
+      <div>
+        { (() => {
```
outoftime: No whitespace after `{`

src/components/AssignmentCreatorTextField.jsx
```diff
@@ -0,0 +1,50 @@
...
+          if (meta.touched && meta.error) {
+            return (
+              <span className="assignment-creator__assignment-notifcation">
```
outoftime: Typo in “notification”

src/components/AssignmentCreatorTextField.jsx
outoftime: Although I think in this case I would just use boolean logic e.g.:

```js
{
  (meta.touched && meta.error) && (
    <span className="assignment-creator__assignment-notifcation">
      {meta.error}
    </span>
  )
}
```

outoftime: Also would probably destructure `meta` in the function params

src/components/TopBar/ExportMenu.jsx
```diff
@@ -33,6 +35,17 @@ const ExportMenu = createMenu({
         </MenuItem>,
       );
 
+      if (isUserAuthenticatedWithGoogle) {
+        items.push(
```
outoftime: I would probably put this last on the list

src/components/AssignmentCreatorForm.jsx
```diff
@@ -0,0 +1,141 @@
...
+function parseDate(value) {
+  const parsedResponse = chrono.parse(value);
+  let parsedDate;
+  if (parsedResponse[0]) {
+    parsedDate = parsedResponse[0].start.date();
+  }
+  return {
+    string: value,
+    parsedDate,
+  };
```
outoftime: You mentioned that this didn’t work right when you used an `Immutable.Record`—can you elaborate? Or push a version that does that so I can poke around with it?

src/css/application.css
```diff
@@ -902,6 +903,134 @@ body {
...
+.assignment-creator__button {
+  border-radius: 0.2em;
+  border-width: 0;
+  box-shadow: var(--box-shadow);
+  cursor: pointer;
+  outline: none;
+  padding: 0.75em;
+  position: relative;
+  transition: background-color 0.3s;
+  margin: 10px;
```
outoftime: Should we use `em` or `rem` here too?

src/records/Course.js
outoftime: I’d be inclined to stick with just the information we’re using—no real cost to adding more properties later if we need them.

src/records/Course.js
```diff
@@ -0,0 +1,26 @@
...
+}, 'Course') {
+  static fromJS(js) {
+    return new this(
+      Object.assign({}, js),
```
outoftime: Do we need the `Object.assign`? `Immutable.Record`’s constructor certainly shouldn’t mess with the argument it’s passed…

src/reducers/googleClassroom.js
```diff
@@ -0,0 +1,26 @@
...
+const defaultState = new GoogleClassroom();
+
+function addCourse(state, course) {
+  return state.setIn(
+    ['courses', course.id],
+    new Course(course),
+  );
+}
+
+/* eslint-disable complexity */
```
outoftime: ?

outoftime:
@pwjablonski is this ready for another round?

package-lock.json
```diff
@@ -9,36 +9,36 @@
...
-        "@babel/traverse": "^7.1.0",
-        "@babel/types": "^7.0.0",
-        "convert-source-map": "^1.1.0",
-        "debug": "^3.1.0",
-        "json5": "^0.5.0",
-        "lodash": "^4.17.10",
-        "resolve": "^1.3.2",
-        "semver": "^5.4.1",
-        "source-map": "^0.5.0"
+      "version": "7.1.2",
```
outoftime: It seems like you’ve got a bunch of version changes here that aren’t related to the work in this pull request—I think you can fix this by doing:

```sh
$ git checkout master -- package-lock.json
$ npm install
```

outoftime:
@pwjablonski let me know when this is ready for a round of review—perhaps the final one! Don’t forget to smash that request review button:

image

src/records/DateRecord.js
```diff
@@ -0,0 +1,6 @@
+import {Record} from 'immutable';
```
outoftime: Calling this `DateRecord` is inconsistent with the other records (which don’t explicitly have `Record` in their names). I’m guessing you did that to disambiguate with the native `Date`, which is a good idea, but maybe something like `FormDate` or `ParsedNaturalLanguageDate` would work?

src/sagas/assignments.js
```diff
@@ -0,0 +1,66 @@
...
+export function* createAssignment(
+  {payload: {selectedCourseId, dueDate, assignmentState}}) {
+  const project = yield select(getCurrentProject);
+  const snapshotKey = yield call(createProjectSnapshot, project);
+
+  const assignmentData = {
+    courseId: selectedCourseId,
+    dueDate,
+    url: yield call(createSnapshotUrl, snapshotKey),
+    title: yield call(generateTextPreview, project),
```
outoftime: I’m sure this works fine, but in a saga I think it’s a little easier to follow the flow if `yield`s are always in their own statement, in this case by assigning these to local variables above. I think you could also call `createProjectSnapshot` and `generateTextPreview` in parallel using the [`all` effect](https://redux-saga.js.org/docs/api/).

src/util/i18nFormatting.js
```diff
@@ -2,11 +2,14 @@ import capitalize from 'lodash-es/capitalize';
 import get from 'lodash-es/get';
 import identity from 'lodash-es/identity';
 
+import {dateToString} from './timeFormatter';
```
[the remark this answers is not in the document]
outoftime: `moment.js` does contribute to bundle size considerably BUT lucky you, it’s already in the bundle, so you get to use it for free 😄

outoftime:
> how would I determine whether redux-form is a big contributor to bundle size?

Oops @pwjablonski missed this as the comment had gotten lost in the diff.

You can enable profiling by running this (instead of the normal development server):

```js
PROFILE_BUILD=true npm run preview
```

The `npm run preview` command builds the bundle in production mode, meaning minified etc., which gives you a more realistic measure of bundle size. And the `PROFILE_BUILD` flag instructs the Webpack configuration to generate a couple of bundle analysis reports, using two different bundle visualization tools.

Once the preview server is up and running (takes a while, particularly since minifying is slow), you can go to either of:

* http://localhost:3000/profile/bundle-analyzer.html
* http://localhost:3000/profile/webpack-visualizer.html

See which one you like better—I think they’re both pretty useful!

package.json
```diff
@@ -186,10 +186,12 @@
     "@fortawesome/react-fontawesome": "^0.1.3",
     "PrettyCSS": "^0.3.17",
     "array-to-sentence": "^2.0.0",
+    "babel-plugin-transform-async-generator-functions": "^6.24.1",
```
pwjablonski: I should take this out now that we are using react-async-component?

src/clients/googleClassroom.js
```diff
@@ -8,3 +13,45 @@ export function createShareToClassroomUrl(snapshotKey, title) {
...
+      hours: dueDate.getUTCHours(),
+      minutes: dueDate.getUTCMinutes(),
+      seconds: 0,
+      nanos: 0,
+    },
+    materials: {
+      link: {
+        url,
+        title,
+        thumbnailUrl: 'N/A',
```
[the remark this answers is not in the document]
pwjablonski: Leaving this for a follow up

src/components/AssignmentCreatorSelectField.jsx
```diff
@@ -0,0 +1,40 @@
...
+    <div>
+      <div className="assignment-creator__select-field">
+        <select
+          className="assignment-creator__select"
+          {...input}
+        >
+          {children}
+        </select>
+        <span className="assignment-creator__caret">
+          <FontAwesomeIcon icon={faCaretDown} key="icon" />
```
outoftime: Ibid. `key` prop

PR #1523, opened 2018-07-28
Consolidate Firebase service into Firebase client

outoftime:
The concept of a service isn’t strongly motivated in Popcode—the idea primarily applies to dependency injection frameworks, which we don’t use.  Instead, just have the client be responsible for maintaining instances of the underlying SDK as needed. `gapi` should probably go this way as well.

The implementation in the Firebase client also encapsulates the construction of the Firebase SDK resources in a function, making it possible to construct more instances in the future.


outoftime:
@pwjablonski relevant to your interests vis-a-vis thoughts on `gapi`

PR #1524, opened 2018-07-29
Account migrate/merge (experimental only)

outoftime:
Sets up a process for migrating an old GitHub-based account into a new Google-based account, merging the data and credentials for the two accounts.

The process is initiated by attempting to link GitHub. If the GitHub account you try to link is already associated with another Firebase account, you’re prompted to merge them:

image

If you choose to proceed, there is a five second grace period during which you can cancel the operation, before anything happens (in case you clicked yes by mistake):

image

After five seconds have passed, the process begins:

image

Once it’s done, you get a success message:

image

Info-level notices are sent to Bugsnag when a migration attempt begins, and when it completes successfully. If there is an error during migration, it is handled by the saga, reported to Bugsnag, and an error message is displayed in the modal.

Currently this feature is only active in experimental mode; once a few people have tested it, we can wide-release it.

### DIFF
--- a/src/components/AccountMigrationUndoGracePeriod.jsx
+++ b/src/components/AccountMigrationUndoGracePeriod.jsx
@@ -1,0 +1,28 @@
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import React, {Fragment} from 'react';
+import {t} from 'i18next';
+
+export default function AccountMigrationUndoGracePeriod({onDismiss}) {
+  return (
+    <Fragment>
+      <p>
+        {t('account-migration.preparing')}
+      </p>
+      <div className="account-migration__buttons">
+        <button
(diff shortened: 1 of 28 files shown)


src/components/AccountMigrationUndoGracePeriod.jsx
```diff
@@ -0,0 +1,28 @@
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import React, {Fragment} from 'react';
+import {t} from 'i18next';
+
+export default function AccountMigrationUndoGracePeriod({onDismiss}) {
+  return (
+    <Fragment>
+      <p>
+        {t('account-migration.preparing')}
```
stormsweeper: I'd suggest adding a countdown timer here.

Issue #1291, opened 2017-12-17
Use Google Fonts CDN instead of self-hosting


outoftime:
We should leverage the likelihood that browsers may have cached CDN-hosted fonts from other websites, and the fact that caching headers on the Google Fonts CDN will almost certainly be more appropriate than the headers we use (since fonts have different caching characteristics than most other assets). It also removes quite a bit of complexity from the build system!

One thing to check: do fonts still work in offline mode? I suspect they should load from the browser cache fine, but we should confirm.

PR #1527, opened 2018-07-31
Moved google fonts to CDN

coryetzkorn:
Addresses issue #1291 

- Replaced local Google fonts with CDN version
- Tested local cache by disabling wifi and refreshing. Everything worked.
- Removed old Google fonts from `/dist`. Was searching for a src version, but looks like these were added to `/dist` manually and not via a build process?
- Removed reference to old Google fonts from webpack.config.


coryetzkorn:
With offline mode enabled fonts still load via the cache. They disappear after a hard refresh.

I don't think we'll see a huge difference in performance either way. We're adding an additional request in exchange for the potential benefit of leveraging Google's CDN caching. Roboto and Inconsolata are widely used and likely cached tho. The biggest benefit I see is it makes managing Google fonts easier in the future if a new weight or font is needed. Don't have to mess with bower or gulp.

Re: FontAwesome... I'd leave it local for now. The icon font should be replaced by inline SVG anyway.

outoftime:
@coryetzkorn inline SVG!? seems like a ton of work! what’s wrong with fontawesome?

outoftime:
Oh wait do you mean switching to [FontAwesome’s SVG library for React](https://fontawesome.com/how-to-use/on-the-web/using-with/react)? Because yes that does seem strictly better!!

coryetzkorn:
@outoftime Icon font is fine since it already exists... but would choose SVG for a new project for [these reasons](https://css-tricks.com/icon-fonts-vs-svg/).

That FontAwesome SVG library for React looks like a great solution! Haven't seen that before.

outoftime:
@coryetzkorn ready to merge once the linter errors are cleared up!

outoftime:
@coryetzkorn gonna have a chance to rebase latest and fix the linter errors?

Issue #1260, opened 2017-12-09
Runtime errors thrown by console statements should be mapped to user-friendly messages


outoftime:
Just as we map runtime errors in project code to user-friendly messages, we should also map them when outputting them in the console. We may need to map a few more errors than we have currently, but the one or two that we currently handle would be a good start.

omardeleo:
Hey Mat, I believe I've got a solution here. Interested in hearing your thoughts when you get a chance, thanks!

## Implementation Outline

-   Inside `async _evaluateConsoleExpression(key, expression)` in the `PreviewFrame` component, modify the arrow function passed to the caller's `error` handler.
-   Inside the arrow function, initialize variable `errorMessage`.
- If `(message === 'Loop Broken!')`, set `errorMessage = t('errors.javascriptRuntime.infinite-loop')`.
- If `(name === 'TypeError')`:
  * generate `ErrorConstructor` to construct error;
  * create variable `normalizedError` and assign it the normalized constructed error;
  * assign `errorMessage = normalizedError.message`.
 - Check to see if `errorMessage` has been assigned a value. If not, assign it the original `message` argument that was passed to the arrow function.
 - Pass `errorMessage` as argument to the `message` parameter of `this.props.onConsoleError()`.

outoftime:
Fixed by #1260

PR #1528, opened 2018-07-31
Display runtime errors in the console

ryanquincypaul:
Hijacked the console.log functionality to pass an error to the console.

### DIFF
--- a/src/reducers/console.js
+++ b/src/reducers/console.js
@@ -61,6 +61,15 @@ export default function console(stateIn, {type, payload, meta}) {
           payload.compiledProjectKey,
         ),
       );
+
+    case 'ADD_RUNTIME_ERROR':
+      return state.set(
+        payload.key,
+        new ConsoleEntry({
+          evaluatedByCompiledProjectKey: payload.compiledProjectKey,
+          error: new ConsoleError({name: "Runtime Error", message: payload.error.text})
+        }),
+      );
     case 'EVALUATE_CONSOLE_ENTRY':
       return payload.trim(' ') === '' ? state : state.setIn(
         ['history', meta.key],
--- a/src/reducers/console.js
+++ b/src/reducers/console.js
@@ -67,7 +67,10 @@ export default function console(stateIn, {type, payload, meta}) {
         payload.key,
         new ConsoleEntry({
           evaluatedByCompiledProjectKey: payload.compiledProjectKey,
-          error: new ConsoleError({name: "Runtime Error", message: payload.error.text})
+          error: new ConsoleError({
+            name: 'Runtime Error',
+            message: payload.error.text,
+          }),
         }),
       );
     case 'EVALUATE_CONSOLE_ENTRY':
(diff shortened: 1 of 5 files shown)


outoftime:
@ryanquincypaul great progress here! let me know when it’s ready for review : )

outoftime:
@ryanquincypaul I took the liberty of giving the PR a descriptive title—hope you don’t mind : )

Let me know if there’s anything I can do to help get this to the finish line!

src/reducers/console.js
```diff
@@ -34,6 +34,15 @@ export default function console(stateIn, {type, payload, meta}) {
           payload.compiledProjectKey,
         ),
       );
+
+    case 'ADD_RUNTIME_ERROR':
+      return state.set(
+        payload.key,
+        new ConsoleEntry({
+          evaluatedByCompiledProjectKey: payload.compiledProjectKey,
+          error: new ConsoleError({name: "Runtime Error", message: payload.error.text})
```
outoftime: So this is a tricky situation—right now if you type code directly into the Popcode console that results in an error, the error class and message are displayed as-is from the interpreter.

However if you were to type the exact same code into the project JavaScript, the error would be “normalized”, i.e. translated into a human-readable message (when recognized).

That seems like an odd inconsistency—I think in the immediate term we probably want to always display raw errors in the console, regardless of what triggered them. There’s already [a separate issue](https://github.com/popcodeorg/popcode/issues/1260) to then start translating the errors displayed in the console across the board.

What do you think? I believe this would require adding the original error as additional information in the `ADD_RUNTIME_ERROR` payload, but that seems reasonable enough.

ryanquincypaul: Can you give me an example? I'm not following what you want to happen. I do see they are different though. Top error is from the javascript section, the bottom is from the console. 
screenshot 2018-08-29 11 27 44

outoftime:
@ryanquincypaul following up on our Slack conversation with more detail. so right now if there is an error in a console expression, we [call back with the raw error information](https://github.com/popcodeorg/popcode/blob/master/src/components/PreviewFrame.jsx#L77) (specifically the error class and name). No human-friendly transformation occurs. However if there is a runtime error in the project code, we first [“normalize” the error](https://github.com/popcodeorg/popcode/blob/master/src/components/PreviewFrame.jsx#L111) and then call back with the normalized information. This is the reason for the inconsistency: the console reducer now sometimes gets normalized error information (from the `ADD_RUNTIME_ERROR` action) and sometimes raw error information (from the `CONSOLE_ERROR_PRODUCED` action).

Ultimately I think I made a mistake when I put the responsibility for normalizing the error in the `PreviewFrame`’s message handler. Ideally, I think, the `ADD_RUNTIME`_ERROR` action would also just be dispatched with raw error information, and then the `errors` reducer would be responsible for translating that into a human-friendly error.

If you’re not up for a restructuring of that size, I think for now it would be fine to simply modify the `ADD_RUNTIME_ERROR` payload to take the raw error information as an additional property, and use that in the `console` reducer.

WDYT?

ryanquincypaul:
@omardeleo is picking this up on https://github.com/popcodeorg/popcode/pull/1686

PR #1531, opened 2018-07-31
Handle up/down arrows in console

razzius:
This code is still pretty rough, but works. It should be updated with actions and reducers, but for a first pass, I used `let` and local variables, and some ++ and --. Fortunately, the history values themselves are already in the store, so that part was fine.

I welcome feedback on the code as-is, but I don't expect it to be merged until all state is in redux.

This could also use some tests.


outoftime:
@razzius nice! what you’ve got so far definitely seems to be heading in the right direction—as you point out of course Popcode is strictly all-Redux for state management so I’m excited to see how you handle that. Did you have an idea of the overall Redux information flow/structure? I could see going about it a few different ways…

webpack.config.js
```diff
@@ -33,7 +33,10 @@ const babelrc = {
     'react',
     ['env', {targets, modules: false}],
   ],
-  plugins: ['syntax-dynamic-import'],
+  plugins: [
+    'syntax-dynamic-import',
+    'syntax-object-rest-spread',
```
razzius: Strangely, I saw this in the yarn.lock but I got an error until I enabled it here. Is object spread the preferred paradigm for partial state updates?

src/selectors/index.js
```diff
@@ -1,7 +1,11 @@
 import getAllProjectKeys from './getAllProjectKeys';
 import getAllProjects from './getAllProjects';
 import getCompiledProjects from './getCompiledProjects';
-import getConsoleHistory from './getConsoleHistory';
+import {
+  getConsoleHistory,
+  getPreviousConsoleInput,
+  getPreviousHistoryIndex,
+} from './getConsoleHistory';
```
razzius: This breaks the pattern of `export default` in existing selectors but it seemed overly split up to have each of these functions in their own file. In my own limited experience I've been defining these inline

src/selectors/getConsoleHistory.js
```diff
@@ -1,3 +1,11 @@
-export default function getConsoleHistory(state) {
-  return state.get('console');
+export function getConsoleHistory(state) {
```
razzius: This breaks the pattern of `export default` in existing selectors but it seemed overly split up to have each of these functions in their own file. In my own limited experience I've been defining these inline

src/actions/console.js
```diff
@@ -24,6 +24,14 @@ export const clearConsoleEntries = createAction(
   'CLEAR_CONSOLE_ENTRIES',
 );
 
+export const setPreviousHistoryIndex = createAction(
```
razzius: I opted for 2 actions each with a single argument, but this could also have been a single `historyPagingState` object with index and current (temporary) input

src/containers/Console.js
```diff
@@ -52,6 +58,14 @@ function mapDispatchToProps(dispatch) {
       dispatch(editorFocusedRequestedLine());
     },
 
+    onSetPreviousHistoryIndex(index) {
+      dispatch(setPreviousHistoryIndex({index}));
+    },
+
+    onStoreCurrentConsoleInput(input) {
+      dispatch(setCurrentConsoleInput({input}));
```
razzius: Since the argument is called `payload` it seemed appropriate to wrap these in objects, but later while debugging I noticed payload was sometimes a primitive. Either way works I think

src/components/ConsoleInput.jsx
```diff
@@ -57,10 +62,65 @@ export default class ConsoleInput extends Component {
...
+            onStoreCurrentConsoleInput(editor.getValue());
+          }
+          const newPreviousHistoryIndex = previousHistoryIndex + 1;
+
+          onSetPreviousHistoryIndex(newPreviousHistoryIndex);
+
+          const historyIndex = history.size - newPreviousHistoryIndex;
+          const {expression} = history.toList().get(historyIndex);
+          editor.setValue(expression);
+          editor.clearSelection();
```
outoftime: In general we want data flow in the application to work like this:

* Component receives props from Redux store, renders itself accordingly
* User does something, action is dispatched to Redux store
* Redux store updates its state based on action
* New props flow out to components from updated Redux state
* Components render themselves based on new props

In this case we are short-circuiting this process by having the editor update its internal state in the handler for the user action—thus the editor component’s UI state is no longer a function merely of props, but also of the internal updates that have been made to editor state based on user actions. This is only possible at all because this component uses an ACE editor instance, which provides a sort of imperative/mutable bubble inside the otherwise purely declarative React-Redux structure.

Since ultimately we want the component’s UI state to be purely a function of its props, I think we’ll need to have a concept of “current value of the console input” living in the Redux store. Then this component can receive that information as a prop, and call `setValue()` any time the received prop does not match the value currently held in the editor instance. This is precisely how the `<Editor>` component operates.

That information flow does impose certain additional requirements on the data shape in Redux, specifically that we need to store the current value of the console input, and update it every time the user adds or removes a character.

Although the proposed information flow probably feels like a lot of extra ceremony, my intuition is that it’ll actually make the code overall simpler—in particular I don’t think the event handlers in this component will need to do much heavy lifting, and the reducers and selectors can take on manageable chunks of logic which in combination produce the right effect.

A qualification that is in order: having “the component’s UI state…purely a function of its props” doesn’t strictly describe the status quo of the console in production today, since the value of the editor input is not described anywhere in props. This comes down to the choice between [controlled components](https://reactjs.org/docs/forms.html#controlled-components) vs. [uncontrolled](https://reactjs.org/docs/uncontrolled-components.html); my feeling is that controlled components are always preferable, but uncontrolled components are acceptable as long as the application code never needs to change the value of the input (a rule which is anyway broken [here](https://github.com/popcodeorg/popcode/blob/master/src/components/ConsoleInput.jsx#L63); I’m sorry!!)

src/reducers/console.js
```diff
@@ -2,7 +2,11 @@ import {OrderedMap} from 'immutable';
 
 import {ConsoleEntry, ConsoleError} from '../records';
 
-const initialState = new OrderedMap();
+const initialState = {
+  history: new OrderedMap(),
+  previousHistoryIndex: 0,
+  previousConsoleInput: '',
+};
```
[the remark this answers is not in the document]
outoftime: Also, thinking about the high-level structure here—not a huge change from what you’ve got but I would probably model this as:

* `entries` (what you’re calling `history`, although either is fine)
* `currentInputEntryIndex` (or something—I don’t love this): I would make this `null` unless/until the user starts navigating up through previous entries.
* `nextConsoleEntry`: This would be what the user typed in the console before navigating up through previous entries; null if they have not navigated up.
* `currentInputValue`: Whatever is currently in the input editor. See my comment on the `<ConsoleInput>` component for motivation.

Another approach would be to store the `nextConsoleEntry` simply as the last item in the list of entries, presumably with an additional boolean property in `ConsoleEntry` to indicate that an entry is in-progress. Ultimately though I think storing it as its own entity is less complicated and no worse at modeling the essential state of the application.

src/reducers/console.js
```diff
@@ -33,22 +35,65 @@ export default function console(stateIn, {type, payload, meta}) {
...
+          set('historyEntryIndex', null).
+          set('nextConsoleEntry', null).
+          set('currentInputValue', state.nextConsoleEntry);
+      }
+      const historyIndex = relevantHistory.size - newHistoryEntryIndex;
+      const {expression} = relevantHistory.get(historyIndex);
+      return state.
+        set('historyEntryIndex', newHistoryEntryIndex).
+        set('currentInputValue', expression);
+    }
```
outoftime: So I think we can probably cut down the amount of code needed here if we don’t handle the up and down cases totally separately, something like:

* If the current history index is `null`, set it to 0.
* Then increment/decrement based on the value of `direction`
* Then check for each boundary case and handle appropriately
* Then handle the non-boundary case

Perhaps also some tweaks to how we represent the state overall might make this operation simpler, namely:

* Go back to representing the default history index as 0
* Keep the `nextConsoleEntry` up to date all the time (in response to the `CONSOLE_CHANGE` action, as long as the index is at 0)

Just throwing ideas out there—not 100% sure that they will simplify the code but maybe worth playing with!

src/selectors/getConsoleHistory.js
```diff
@@ -1,3 +1,3 @@
 export default function getConsoleHistory(state) {
-  return state.get('console');
+  return state.get('console').history;
```
outoftime: `state.getIn(['console', 'history'])` is a bit more idiomatic for Immutable

src/selectors/getCurrentInputValue.js
```diff
@@ -0,0 +1,3 @@
+export default function getCurrentInputValue(state) {
+  return state.get('console').currentInputValue;
```
outoftime: Ibid. `getIn`

src/selectors/getCurrentInputValue.js
```diff
@@ -0,0 +1,3 @@
+export default function getCurrentInputValue(state) {
```
outoftime: Let’s be more specific here since all selectors live in a single namespace, e.g. `getCurrentConsoleInputValue`.

src/components/ConsoleInput.jsx
```diff
@@ -57,10 +73,27 @@ export default class ConsoleInput extends Component {
...
+      editor.commands.addCommand({
+        name: 'historyNext',
+        bindKey: 'Down',
+        exec: () => {
+          onNavigateConsoleHistory('DOWN');
+        },
+      });
+
       session.on('change', ({action, lines}) => {
         if (action === 'insert' && lines.length === 2) {
```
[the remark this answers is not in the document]
razzius: Interesting. It'd be great to get rid of the `lines.length === 2` part which that might allow.

I'll pass on updating this for now, though, and attempt that refactor as a separate PR.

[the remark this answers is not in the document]
outoftime: Sounds right to me!

src/reducers/console.js
```diff
@@ -12,19 +10,23 @@ export default function console(stateIn, {type, payload, meta}) {
...
+          ).set(
+            'evaluatedByCompiledProjectKey',
+            payload.compiledProjectKey,
+          ),
         ),
-      );
-
+      ).set('currentInputValue', '').
+        set('nextConsoleEntry', null).
+        set('historyEntryIndex', null);
```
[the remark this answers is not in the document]
razzius: True

src/reducers/console.js
```diff
@@ -33,22 +35,65 @@ export default function console(stateIn, {type, payload, meta}) {
...
+      return state.set('history', state.history.set(
         meta.key,
         new ConsoleEntry({
           value: payload.value,
           evaluatedByCompiledProjectKey: payload.compiledProjectKey,
         }),
+      ));
+    case 'NAVIGATE_CONSOLE_HISTORY': {
+      const relevantHistory = state.history.toList().filter(
+        entry => entry.expression !== null,
```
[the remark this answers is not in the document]
razzius: I wish !== was just != and isNull hides this. Seeing as isNull is more-used in this codebase I'm happy to change this

src/reducers/console.js
```diff
@@ -12,20 +10,20 @@ export default function console(stateIn, {type, payload, meta}) {
...
         ),
-      );
-
+      ).delete('currentInputValue').
+        delete('nextConsoleEntry').
+        delete('historyEntryIndex');
     case 'CONSOLE_ERROR_PRODUCED':
-      return state.update(
-        payload.key,
+      return state.updateIn(['history', payload.key],
```
outoftime: Minor thing, but project style is to have the array on its own line (basically it’s all-or-nothing in terms of multiline arguments)

src/reducers/console.js
```diff
@@ -35,20 +33,62 @@ export default function console(stateIn, {type, payload, meta}) {
         ),
       );
     case 'EVALUATE_CONSOLE_ENTRY':
-      return payload.trim(' ') === '' ? state : state.set(
-        meta.key,
-        new ConsoleEntry({expression: payload}),
+      return payload.trim(' ') === '' ? state : state.updateIn(
```
outoftime: In this case, since you are replacing the value rather than modifying it, you can just use `setIn`

src/reducers/console.js
```diff
@@ -35,20 +33,62 @@ export default function console(stateIn, {type, payload, meta}) {
...
       );
     case 'CLEAR_CONSOLE_ENTRIES':
       return initialState;
+    case 'CONSOLE_CHANGE':
+      return state.set('currentInputValue', payload.value);
     case 'CONSOLE_LOG_PRODUCED':
-      return state.set(
-        meta.key,
-        new ConsoleEntry({
+      return state.updateIn(
```
outoftime: Ibid. `setIn`

src/reducers/console.js
```diff
@@ -35,20 +33,62 @@ export default function console(stateIn, {type, payload, meta}) {
...
+
+      const firstUp = newHistoryEntryIndex === 1 &&
+        payload.direction === 'UP';
+
+      if (firstUp) {
+        return updatedState.set('nextConsoleEntry', state.currentInputValue);
+      }
+
+      return updatedState;
+    }
```
outoftime: OK so this is definitely looking better, I do want to make one more pitch for always maintaining the `nextConsoleEntry` though!! In particular, I think it would allow you to remove these lines:

  ```js
const firstUp = newHistoryEntryIndex === 1 &&
  payload.direction === 'UP';

if (firstUp) {
  return updatedState.set('nextConsoleEntry', state.currentInputValue);
}
```

Which seems like a pretty big win, not as much in terms of literal lines of code removed as just dropping a special case and making the whole thing easier to reason about.

If you still don’t think it’s worth it though I will shut up about it : )

File: src/reducers/console.js [hunk 19]
```diff
@@ -35,20 +33,62 @@ export default function console(stateIn, {type, payload, meta}) {
...
+        state.historyEntryIndex + 1 :
+        state.historyEntryIndex - 1;
+
+      if (newHistoryEntryIndex < 0 ||
+          newHistoryEntryIndex === relevantHistory.size) {
+        return state;
+      }
+
+      const downToPreviousInput = newHistoryEntryIndex === 0 &&
+        payload.direction === 'DOWN';
```
outoftime: I don’t think there’s any particular need to check this condition explicitly—the boundary guards just above should make this redundant.

src/reducers/console.js
```diff
@@ -35,20 +33,62 @@ export default function console(stateIn, {type, payload, meta}) {
...
+      const relevantHistory = state.history.toList().filter(
+        entry => entry.expression !== null,
+      );
+
+      const newHistoryEntryIndex = payload.direction === 'UP' ?
+        state.historyEntryIndex + 1 :
+        state.historyEntryIndex - 1;
+
+      if (newHistoryEntryIndex < 0 ||
+          newHistoryEntryIndex === relevantHistory.size) {
```
outoftime: Much as you check `<0` for the lower bound, I would check `>== relevantHistory.size` for the upper bound to make it explicit that you are enforcing a range of allowed values.

src/reducers/console.js
```diff
@@ -35,20 +33,62 @@ export default function console(stateIn, {type, payload, meta}) {
...
+        ['history', meta.key],
+        () => new ConsoleEntry({
           value: payload.value,
           evaluatedByCompiledProjectKey: payload.compiledProjectKey,
         }),
       );
+    case 'NAVIGATE_CONSOLE_HISTORY': {
+      const relevantHistory = state.history.toList().filter(
+        entry => entry.expression !== null,
+      );
```
outoftime: I’d be inclined to extract this into a selector using `reselect` to cache the last value so that we don’t have to recalculate it on every navigation unless the console history itself has changed.

src/reducers/console.js
```diff
@@ -35,20 +33,62 @@ export default function console(stateIn, {type, payload, meta}) {
...
+
+      const newHistoryEntryIndex = payload.direction === 'UP' ?
+        state.historyEntryIndex + 1 :
+        state.historyEntryIndex - 1;
```
outoftime: It would be pretty slick to define an [enum](https://github.com/popcodeorg/popcode/blob/master/src/enums/index.js) with explicit values for `UP` and `DOWN` (which the [`enum` library](https://www.npmjs.com/package/enum) lets you do by initializing the enum with an object) so that we can just call `payload.direction.value` here rather than having to do a conditional.

src/reducers/console.js
```diff
@@ -35,20 +33,62 @@ export default function console(stateIn, {type, payload, meta}) {
...
+          set('currentInputValue', state.nextConsoleEntry);
+      }
+
+      const historyIndex = relevantHistory.size - newHistoryEntryIndex;
+      const {expression} = relevantHistory.get(historyIndex);
+      const updatedState = state.
+        set('historyEntryIndex', newHistoryEntryIndex).
+        set('currentInputValue', expression);
+
+
```
outoftime: Project style forbids multiple consecutive blank lines

src/reducers/console.js
razzius: Nice, didn't know there was such a method

src/reducers/console.js
razzius: Makes sense

src/reducers/console.js
razzius: This is handling the user pressing down to get their previously typed input, so  it's not like the no-op out of bounds boundary cases handled above

src/reducers/console.js
razzius: 👀 ❗️

src/reducers/console.js
razzius: My hesitation with doing that is that it makes it look like the index might exceed relevantHistory.size, which the code prevents. I suppose it's inconsistent the way I have it, though, since I'm checking < 0 and === size.

I found [inRange](https://lodash.com/docs/4.17.10#inRange) in the lodash docs so I'm happy to use that to say "if the index is not in the range, return the state with no updates"

outoftime: `inRange` seems like the ideal have/eat cake!

src/reducers/console.js
outoftime: Sorry, just to be clear—I don’t think the second part of the conjunction (`payload.direction === 'DOWN'`) is needed; the only way to get to `0` is down, right?

src/reducers/console.js
razzius: For me, my mental model is this: "up and down arrows update the state related to history, and typing directly updates the current input value"

I'll admit that a conditional update, which necessitates the temporary variable updatedState, isn't great, but I think it's worth it to keep the CONSOLE_CHANGED handling as simple as

```
    case 'CONSOLE_CHANGE':
      return state.set('currentInputValue', payload.value);
```

src/reducers/console.js
razzius: I think recalculation fine here - caching would complicate the code and I don't expect performance to be an issue. If performance measurements indicated that the ui was noticeable less responsive due to this (which I'm aware is extra important for students using old browsers and hardware) I'd happily look for such a solution, but I'll pass on implementing that for now.

src/reducers/console.js
razzius: I do like enums but I think the real refactoring win for this code would be to split up and down navigation into different actions. That way, there's no need for a conditional to figure out if the index should be incrementing or decrementing. It would also mean we don't have to use constant strings like 'UP' nor enums.

If you agree that would be beneficial, such an approach would necessitate a shared function that does the history lookup (splitting out this part is itself a good step IMO, since the `case 'NAVIGATE_CONSOLE_HISTORY':` is already 38 lines). Could I put such a function in the top level of the `src/reducers/console.js` module (not being exported)?

src/reducers/console.js
razzius: Even better - no external method dependency! `Immutable.List.has` can tell whether or not the index I'm at is in range.

src/reducers/console.js
razzius: Oh I see, you're referring to the `=== 'DOWN'`. Right you are.

src/reducers/console.js
outoftime: I don’t necessarily agree that it would complicate the code on balance—more indirection, yes, but also more self-describing code (via the selector’s name) and using a pattern that’s standard in the codebase. BUT I am receptive to the argument that it’s a premature optimization 🙂

razzius:
@outoftime I made changes in response to your comments, and added a bit of my own refactoring. I realized that if I add the next console input to the history and reverse that list, I can index into it using the history entry index, which to me is a lot nicer than manually doing the size - index and having 2 index variables that are going opposite directions. The `nextConsoleEntry` sticks around in the state in this approach even after it becomes the current input after pressing down, but it doesn't cause any problems and will be updated to the correct value when the user presses us again.

I found that `Immutable.List.of(1).has(-1)` returns `true` so I couldn't use .has to see if the index was out of range, and did end up pulling in `inRange` from lodash. Interestingly, the negative indexing would allow not sorting the expression history, and doing `expressionHistory.get(-newHistoryEntryIndex)`. I find that less readable than explicitly reversing the history.

Now I'm thinking yet another way to go about this would be to treat the history as a stack rather than a list/queue as it is now essentially reversing the history as it's built.

At this point, I'm happy with the code, though I notice for interactions as simple as typing a long input line, the ui starts to lag. There's definitely performance improvements to be had in the `NAVIGATE_CONSOLE_HISTORY` handling but I'm not sure what to make of the lag when just typing, since the handler there is just 1 line. Maybe that's just because I have redux devtools open. I'll keep experimenting. I'm sure there are patterns to debounce actions.

razzius:
Yeah, interacting with the app without devtools open is nice and smooth 😎

src/reducers/console.js
razzius: For reference, here's what it would look like to keep the next input value always in sync:

```js
case 'CONSOLE_CHANGE': {
  if (state.historyEntryIndex === 0) {
    return state.set('nextInputValue', payload.value).set('currentInputValue', payload.value)
  }
  return state.set('currentInputValue', payload.value);
}
```

What all this makes me wonder is if there's a way to apply an update unconditionally and apply an update conditionally without duplicated code or temporary variables. I guess helper functions could help

outoftime: The [`withMutations()`](https://facebook.github.io/immutable-js/docs/#/Map/withMutations) method might be a good way to apply an update conditionally without needing a temporary variable?

src/reducers/console.js
```diff
@@ -35,20 +34,55 @@ export default function console(stateIn, {type, payload, meta}) {
...
+    case 'NAVIGATE_CONSOLE_HISTORY': {
+      const newHistoryEntryIndex = payload.direction === 'UP' ?
+        state.historyEntryIndex + 1 :
+        state.historyEntryIndex - 1;
+
+      const expressionHistory = state.history.toList().
+        map(entry => entry.expression).
+        filter(expression => expression !== null).
+        concat(state.nextConsoleEntry).
+        reverse();
```
outoftime: I don’t feel strongly about it, but [`Immutable.List#get()`](https://facebook.github.io/immutable-js/docs/#/List/get) does allow you to index into the end of a list with negative numbers, which would maybe retain the pleasantness while avoiding the need to reverse the list every time.

razzius:
Looking into that CONSOLE_INPUT_CHANGE action spamming. One definite issue is https://github.com/ajaxorg/ace/issues/503 - calling setValue also dispatches an ace change event, but it looks like this can be handled in the application logic

src/reducers/console.js
razzius: It seems like not a clear win to me, since -0 is 0 so I can't just use negative index.

razzius:
Now the extra actions are gone - there's a single action per user action - and I'm happy with the code as is. I know there are some suggestions that you've made @outoftime that I passed on, so if you feel strongly about any of theme let me know and I'm happy to make them.

src/reducers/console.js
outoftime: Good call

razzius:
Update: I missed your comment approving splitting navigate console history into 2 distinct actions. I carried that out using a module-local function and am happy with the result. I also experimented with `withMutations` and found that I liked that more than a temporary variable `updatedState`.

razzius:
Even better, I made an actual helper function since that's what withMutations was doing with an anonymous function.

outoftime:
@razzius I like the general direction here, but it seems like your last commit broke the functionality (the same error that’s causing tests to fail is also popping up in the app)

src/components/ConsoleInput.jsx
```diff
@@ -57,10 +73,41 @@ export default class ConsoleInput extends Component {
...
-        if (action === 'insert' && lines.length === 2) {
-          onInput(editor.getValue().replace('\n', ''));
-          editor.setValue('', 0);
+        const programmaticEdit = ['historyNext', 'historyPrevious'].includes(
+          editor.curOp.command.name,
+        );
+
+        if (programmaticEdit) {
+          return;
+        }
```
outoftime: I think we can actually drop this check—the check for `value !== this.props.currentInputValue` below seems to catch every case I can think of, doesn’t require specific knowledge of the kinds of things that might cause the input to change, and doesn’t require inspecting a transient state property on the editor. Unless there is a case I am missing?

src/reducers/console.js
```diff
@@ -1,8 +1,36 @@
-import {OrderedMap} from 'immutable';
+import inRange from 'lodash-es/inRange';
 
-import {ConsoleEntry, ConsoleError} from '../records';
+import {ConsoleState, ConsoleEntry, ConsoleError} from '../records';
 
-const initialState = new OrderedMap();
+const initialState = new ConsoleState();
+
+function updateConsoleForHistoryIndex(index, state) {
```
outoftime: Generally in reducers `state` is the first argument so for consistency let’s switch the order here?

src/reducers/console.js
```diff
@@ -1,8 +1,36 @@
...
+  }
+
+  const expression = expressionHistory.get(index);
+
+  return state.
+    set('historyEntryIndex', index).
+    set('currentInputValue', expression);
+}
+
+function setNextConsoleEntry(newHistoryEntryIndex, state) {
```
outoftime: Ibid. argument order

src/reducers/console.js
```diff
@@ -35,20 +62,38 @@ export default function console(stateIn, {type, payload, meta}) {
...
-        meta.key,
+      return state.setIn(
+        ['history', meta.key],
         new ConsoleEntry({
           value: payload.value,
           evaluatedByCompiledProjectKey: payload.compiledProjectKey,
         }),
       );
+    case 'PREVIOUS_CONSOLE_HISTORY': {
+      const newHistoryEntryIndex = state.historyEntryIndex + 1;
```
outoftime: I don’t feel strongly about it, but you could inline these variables and remove the need for the weird case statement block form.

src/components/ConsoleInput.jsx
razzius: I agree that this isn't the most elegant solution, but it seems that due to the internal design of ace, this is the best approach.

The problem appears to me that when I have typed `x` in the console and hit the up arrow that will go to the last entry (let's call that `y`), in order to remove the x, ace will dispatch a `change` with action `remove` that will set the value to empty string, then another change event that will set the value to `y`. The first `remove` calls `componentDidUpdate` with empty string as the value, which is why the state input value would switch to empty string and back to the actual value whenever the user goes up and down in the history, in a previous edition of this PR.

Unfortunately the `remove` which sets the value to empty string is indistinguishable from the user deleting the input manually, unless we look at `editor.curOp.command.name` to see if it's happening during a history action.

This check isn't strictly necessary, since dispatching extra actions: x -> empty string -> y still gets to y, but especially while debugging having the input switch to empty string was annoying and with this check every user action can correspond to a single redux action.

It'd be great if calling `setValue` didn't also trigger `change`, but it does, so we have to prevent the update from triggering other updates using this check.

PR #1534, opened 2018-08-03
⚠️ In Progress: Console Pretty Print ⚠️

### LINKED ISSUE #1252: Console should pretty-print values

alessbell:
screenshot 2018-08-03 at 4 17 50 pm

Closes #1252. Revisiting this today while attending the [Write/Speak/Code](https://writespeakcode.com) conference :) 

Got some basic pretty printing working and wanted to get some feedback...

I'm serializing the input with [`serialize-javascript`](https://github.com/yahoo/serialize-javascript), then rendering each argument in an `<Inspector />`  component using [`react-inspector`](https://github.com/xyc/react-inspector).

I landed at this solution after realizing I needed to be able to handle multiple arguments. In `handleConsoleLogs.js` on L10 where previously we were calling

```javascript
params: prettyPrint(args).join(' ')
```

joining the args meant that `console.log(1, 2)` would eventually try to deseralize `""1" "2""` which would throw a `SyntaxError`, hence passing the serialized array through.

Still need to look into tests, add a missing `key` prop on `<Inspector />` (I see the linter doesn't want me to use the index--uuid ok?), and polish up the styling by more finely-tuning the spacing, among other things.

IIRC pretty printing objects isn't a high priority as they're not taught in the first year curriculum, but being able to more easily distinguish strings and numbers seems like a clear UX win. Curious about everyone's thoughts here!


alessbell:
Thanks for the feedback @outoftime! I _thought_ eval might be a non-starter but figured I'd pick up where I left off. Just added a commit that replaces `serialize-javascript` + eval with `super-json` as you suggested which produces the same output as the PR screenshot.

I did notice an issue, however: DOM elements aren't serialized properly and return `Object {}` whereas Popcode currently logs something like `<h1 class="myclass">...</h1>`... seems like we could write a custom serializer, though I don't have a clear idea wrt implementation.

Also just found [`console-feed`](https://github.com/samdenty99/console-feed). Looks like it could be interesting but would potentially involve more foundational changes to the way we handle logs.

src/components/ConsoleOutput.jsx
```diff
@@ -23,7 +32,14 @@ export default function ConsoleOutput({entry, isActive}) {
         }
       >
         {chevron}
-        {value}
+        {Object.keys(deserialize(value)).map(
```
outoftime: I think we can probably do this more straightforwardly with `deserialize(value).map()`, or the lodash equivalent? Both should pass the array index as the second argument to the iterator function.

src/components/ConsoleOutput.jsx
```diff
@@ -2,9 +2,18 @@ import isNil from 'lodash-es/isNil';
 import React from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import Inspector from 'react-inspector';
```
outoftime: How big is the footprint of `react-inspector`? Generally I try to avoid loading any library dependencies in the main bundle that aren’t needed to bootstrap a fresh Popcode environment. Since this is a UI library and asynchronously rendering React components is kind of a pain, I’m down to do it in a fast-follow if the library itself is small, although [this (sadly unmerged) PR](https://github.com/popcodeorg/popcode/pull/1426/files#diff-038c11f73da9e0721eb03924327efa1cR1) has some very nice prior art on async loading a component so as to async render a library dependency!

src/components/ConsoleOutput.jsx
```diff
@@ -2,9 +2,18 @@ import isNil from 'lodash-es/isNil';
...
+import superJson from 'super-json';
```
outoftime: Same question about the footprint of `super-json`, although should be simpler to lazy-load it if it’s moved to the JSChannel handler, since that’s already an asynchronous context.

src/css/application.css
```diff
@@ -623,6 +623,11 @@ body {
 .console__row {
   padding-left: 1em;
   position: relative;
+  display: inline-flex !important;
+}
+
+.console__row > li {
```
outoftime: Just to make sure, the `<li>` in question is coming from the `<Inspector>`, right?

src/previewSupport/prettyPrint.js
```diff
@@ -1,5 +1,7 @@
-import inspect from 'object-inspect';
+import superJson from 'super-json';
+
+const myJson = superJson.create();
 
 export default function prettyPrint(value) {
```
outoftime: Honestly we probably don’t need `prettyPrint` to be its own module (or even its own function) at this point—I would just inline it in `handleConsoleLogs`

src/previewSupport/prettyPrint.js
```diff
@@ -1,5 +1,7 @@
-import inspect from 'object-inspect';
+import superJson from 'super-json';
+
+const myJson = superJson.create();
```
outoftime: Same comment as above re `my`

Issue #1403, opened 2018-03-30
Add StrictMode to React component root


outoftime:
React 16.3 introduces [`<StrictMode>`](https://reactjs.org/docs/strict-mode.html), which provides extra runtime warnings for deprecated or potentially buggy behavior. We should add it to the root component.

alexpelan:
blah. so every one of our StrictMode errors is currently caused by `react-redux` and our connected components. This is fixed in `react-redux@next` (see: https://github.com/reduxjs/react-redux/issues/897 ). Unfortunately, Popcode doesn't seem to boot with this version of `react-redux` (see: https://github.com/reduxjs/react-redux/issues/965 and others).

The good news is I think we can just wait for the next cut of `react-redux`, `yarn add` that, and we'll be good then.

PR #1549, opened 2018-08-21
Add React StrictMode Globally

alexpelan:
Seems to work, we get some console errors now.

Fixes: #1403

### DIFF
--- a/src/components/Application.jsx
+++ b/src/components/Application.jsx
@@ -40,11 +40,13 @@ class Application extends React.Component {
     }
 
     return (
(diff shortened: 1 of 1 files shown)


outoftime:
@alexpelan wait I swear I remember a thorough writeup about how it’s not really possible to fix the StrictMode errors until a new version of, I think, redux comes out? Where was that? Or was I dreaming?

Anywho, assuming I was not dreaming, I’ll just merge this once we no longer have library dependencies in the way ; )

PR #1550, opened 2018-08-22
Dismiss Snapshot Notification after Project is Updated

### LINKED ISSUE #1048: Snapshot notification should disappear when project changes
Right now, if you:

* Do some stuff
* Create a snapshot
* Leave the snapshot copy prompt open
* Make more modifications to the project

You might be tempted to think that clicking the copy button in the snapshot notification would give you a snapshot of the latest state of your project. But in fact, you’ll still get a link to a snapshot of the state of the project when you clicked **Snapshot**.

Simple fix, I think, is to just make the snapshot notification disappear as soon as the project is changed in some way.

alecmerdler:
### Description

Deletes notification from Redux store in the `UPDATE_PROJECT_SOURCE` reducer.

### Screenshots

popcode-dismiss-notification

Fixes #1048

### DIFF
--- a/src/reducers/ui.js
+++ b/src/reducers/ui.js
@@ -38,7 +38,9 @@ export default function ui(stateIn, action) {
       return state.set('isEditingInstructions', false);
 
     case 'UPDATE_PROJECT_SOURCE':
-      return state.set('isTyping', true);
+      return state.
+        set('isTyping', true).
+        deleteIn(['notifications', 'snapshot-created']);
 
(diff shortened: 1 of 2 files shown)


alecmerdler:
Per feedback by @outoftime, I will update this to mutate the state in the reducer instead of firing another action.

alecmerdler:
@outoftime Made the necessary change to the reducer and updated the test. Thanks!

PR #1551, opened 2018-08-22
Add subtle edit button to instructions bar

### LINKED ISSUE #1396: Add an edit button in the instructions pane
Idea is to add an extra edit button so that it is easier to edit instructions.

This should ideally be something pretty subtle—maybe an icon that appears in the corner of the instructions pane only when you’re hovering over it.

greenberga:
Closes #1396

### DIFF
--- a/src/components/Workspace.jsx
+++ b/src/components/Workspace.jsx
@@ -30,6 +30,7 @@ export default class Workspace extends React.Component {
       this,
       '_handleUnload',
       '_handleClickInstructionsBar',
+      '_handleClickInstructionsEditButton',
     );
     this.columnRefs = [null, null];
   }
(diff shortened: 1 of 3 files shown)


greenberga:
I had actually meant to bring up the changing bar width! Seemed like I didn't have to though 🙂 Any idea what might be causing that? It's weird because the characters _seem_ to have the same width...

All good ideas! Will get to implementing them.

outoftime:
@greenberga soooo weirdly now I can’t repro the width changing!? even though i still have the same commit checked out…

anywho, also occurred to me that a quick transition on the opacity (fade in/out) would be nice ; )

greenberga:
I can still reproduce the changing width in Firefox...I wonder if it's just because the icon is a different size? Let me try switching icons and see if that fixes it.

greenberga:
@outoftime okay, I think I did all the things! Would you mind taking another look when you have the time?

PR #1556, opened 2018-09-10
fontawesome-react

### LINKED ISSUE #1533: Switch to Font Awesome’s SVG implementation
Font Awesome now offers an SVG implementation with an [out-of-box React library](https://fontawesome.com/how-to-use/on-the-web/using-with/react). This seems [strictly better](https://css-tricks.com/icon-fonts-vs-svg/) than using an icon font (h/t @coryetzkorn for pointing this out).

chasestarr:
closes #1533 

it's convenient to look at icon 'names' rather than unicodes :)
seems like it's fine to remove the `fonts` gulp task?


chasestarr:
i was thinking that as well... kind of hidden at the moment. let me try to add within the components and see what it feels like

unsure if there are downsides to 'adding' an icon to the 'library' multiple times, but i'm guessing it will be fine

src/components/Console.jsx
```diff
@@ -9,6 +16,10 @@ import {EditorLocation} from '../records';
 import ConsoleEntry from './ConsoleEntry';
 import ConsoleInput from './ConsoleInput';
 
+library.add(faBan);
+library.add(faChevronDown);
+library.add(faChevronUp);
```
outoftime: @chasestarr my apologies! i should have been clearer about what i was proposing. i think the simplest/clearest thing might be to avoid the `library` concept altogether and just directly use the imported icons in the JSX, e.g.:

```js
import {
  faBan,
  faChevronDown,
  faChevronUp,
} from '@fortawesome/free-solid-svg-icons';
import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';

// …

<FontAwesomeIcon icon={ban} />
```

This follows the [explicit import](https://github.com/FortAwesome/react-fontawesome#explicit-import) approach mentioned in their documentation.

Would that work?

chasestarr: 👍 yes that is much better

Issue #304, opened 2016-10-22
Project test suites


outoftime:
https://trello.com/c/O49A0kcA/31-project-test-suites

Teacher can include a test suite with an assignment, which gives automatic feedback to the student

pwjablonski:
I took an initial shot at this issue in #1557. Closing down the PR for now but feel free to take a look and reopen.

PR #1557, opened 2018-09-11
Testing framework

pwjablonski:
Hey @outoftime

I’ve been thinking a bunch about assessment recently. I’m increasingly seeing the usefulness of having an approachable assessment/testing platform within Popcode that would afford us the flexibility to create self-grading assignments. Address  #304 that was logged 2 years ago.

Over Labor Day I tried taking a shot at making a simple testing framework integrated into Popcode. It’s a simple modal-based UI for both creating the tests and another to view the test results. I chose tape as the framework given that it was an easy implementation and we already require it for the Unit Tests.

There is one LARGE question I have about this approach and a couple of smaller mostly UI related ones.

The questions boil down to I don’t know much about the dangers of eval(). I figured since the console input implementation use eval() in the sandboxed iframe this would be the correct model to follow. Most of all, I don’t know how to adequately scope the function so as to not cause any potential vulnerabilities.

All this to say I was really excited to take this on but I am not tied to any of the approaches taken thus far. Also, realize have a lot of other PR’s outstanding as well as the next step in assignment feature so if this doesn’t seem to be on the right track or if I’m barreling down the wrong road please let me know!

Looking forward to hearing your thoughts!

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,14 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Launch Program",
+            "program": "${file}"
+        }
+    ]
+}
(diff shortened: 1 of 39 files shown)


pwjablonski:
Spiking this PR.  Reach out if you are interested in talking about it or reopening it.

PR #1582, opened 2018-10-02
Batch console.logs every 1 second for infinite loop handling

### LINKED ISSUE #1354: Loop breaking doesn't work if there's a console.log() in the loop body
Start with: https://popcode.org/?snapshot=05697db4-3317-43aa-b8c7-eb829c9e4992 . Popcode notices pretty quickly (within about a second) that there's an infinite loop, stops executing, and tells you about it. If you change `i--` to something else that's syntactically valid (e.g. `i` or `i -= 1`), it'll respond reasonably quickly as well.

Now uncomment the `console.log(i);` statement. Depending on how quickly after that you try to type, either the tab will hang for a few seconds and then your typed characters will come through, or if you wait long enough the tab will hang unrecoverably and you have to close it.

---

Further information: This isn’t strictly a loop-breaking issue but rather one with performance of `console.log()` when it’s invoked a large number of times in quick succession. For instance, a finite loop that issues a `console.log()` 1,000,000 times will freeze Popcode. Initial research indicates that the problem is the large number of `CONSOLE_LOG_PRODUCED` actions being dispatched, yielding enough spins through React’s update cycle to overwhelm the browser.

I believe the simple answer here is just to debounce the log messages and dispatch them in batches. This could either be done at the Redux level or, probably more simply, in the component container.

wylieconlon:
Closes #1354

This is an optimization focused on making infinite loop handling more performant for students. The root cause of the performance bottleneck appears to be React's handling of thousands of actions flooding the system during an infinite loop with `console.log` calls inside, which caused the browser UI to lock up before the loop protection code could kick in.

This fixes the bottleneck by adding a debounced queue in front of the log action, so that actions are batched every second and then rendered to the Popcode console.

This replaces the `CONSOLE_LOG_PRODUCED` action with the `CONSOLE_LOG_BATCH_PRODUCED` action.


src/actions/console.js
```diff
@@ -37,8 +37,7 @@ export const consoleInputChanged = createAction(
   value => ({value}),
 );
 
-export const consoleLogProduced = createAction(
-  'CONSOLE_LOG_PRODUCED',
-  (value, compiledProjectKey) => ({value, compiledProjectKey}),
-  (_value, _compiledProjectKey, key = uuid().toString()) => ({key}),
+export const consoleLogBatchProduced = createAction(
+  'CONSOLE_LOG_BATCH_PRODUCED',
+  actions => ({actions}),
```
outoftime: Echoing other comment, I’d probably call this `entries` rather than `actions`.

src/containers/Preview.js
```diff
@@ -27,7 +29,27 @@ function mapStateToProps(state) {
...
 
+function generateConsoleLogDispatcher(dispatch, timeout) {
+  let queue = [];
+
+  function flushQueue() {
+    dispatch(consoleLogBatchProduced(queue));
+    queue = [];
+  }
+
+  const timer = throttle(flushQueue, timeout, {
```
outoftime: I don’t think `timer` really describes it, maybe something like `throttledFlushQueue`, `flushQueueWithThrottling`, `flushQueueSoon`, etc.?

src/containers/Preview.js
```diff
@@ -27,7 +29,27 @@ function mapStateToProps(state) {
...
+  });
+
+  return function addLogEntry(value, compiledProjectKey) {
+    queue.push({value, compiledProjectKey, key: uuid().toString()});
+    timer();
+  };
+}
+
 function mapDispatchToProps(dispatch) {
+  const logDispatcher = generateConsoleLogDispatcher(dispatch, 1000);
```
outoftime: I would still give this an imperative name like `dispatchLogEntry` etc.

src/reducers/console.js
```diff
@@ -73,14 +73,19 @@ export default function console(stateIn, {type, payload, meta}) {
...
-      return state.setIn(
-        ['history', meta.key],
-        new ConsoleEntry({
-          value: payload.value,
-          evaluatedByCompiledProjectKey: payload.compiledProjectKey,
-        }),
-      );
+    case 'CONSOLE_LOG_BATCH_PRODUCED': {
+      let previousMap = state.get('history');
+      payload.actions.forEach((action) => {
```
outoftime: This isn’t really a collection of _actions_—objects with a `type`, a `payload`, and possibly a `meta`—but rather just a collection of console entry information. So I’d refer to them as something like `entry` heren (and probably destructure as well)

src/reducers/console.js
```diff
@@ -73,14 +73,19 @@ export default function console(stateIn, {type, payload, meta}) {
...
+        previousMap = previousMap.set(
+          action.key,
+          new ConsoleEntry({
+            value: action.value,
+            evaluatedByCompiledProjectKey: action.compiledProjectKey,
+          }),
+        );
+      });
+      return state.set('history', previousMap);
```
outoftime: In terms of the nicest way to do this in Immutable, I think `state.update()` is always preferable in this situation, and I do think this is probably a case where Immutable’s `withMutations` would be appropriate. As I understand it [`withMutations` is primarily indicated](https://facebook.github.io/immutable-js/docs/#/OrderedMap/withMutations) in situations where you are making a bunch of discrete changes to an immutable data structure, and it allows you to avoid creating a ton of intermediate objects that are immediately thrown away. Seems ideally suited to this situation.

src/containers/Preview.js
wylieconlon: Done

src/reducers/console.js
wylieconlon: Done

src/reducers/console.js
```diff
@@ -73,14 +73,17 @@ export default function console(stateIn, {type, payload, meta}) {
...
+    case 'CONSOLE_LOG_BATCH_PRODUCED': {
+      const newHistory = state.get('history').withMutations((map) => {
+        payload.entries.forEach(({value, compiledProjectKey, key}) => {
+          map.set(key, new ConsoleEntry({
+            value,
+            evaluatedByCompiledProjectKey: compiledProjectKey,
+          }));
+        });
+      });
+      return state.set('history', newHistory);
```
outoftime: This is the smallest thing, but I think in this situation it’s more elegant to use `update()` like so:

```js
return state.update('history', history => history.withMutations((map) => {
  payload.entries.forEach(({value, compiledProjectKey, key}) => {
    map.set(key, new ConsoleEntry({
      value,
      evaluatedByCompiledProjectKey: compiledProjectKey,
    }));
  });
}));
```

Just avoids the need to create the intermediate `newHistory` variable.

wylieconlon:
@outoftime Updated the reducer based on your feedback, thanks!

outoftime:
@wylieconlon FWIW, I wrote a loop that prints 10,000,000 messages to the console—it finished pretty quickly but the environment is not terribly usable afterwards if you’ve got the console open. The situation is much improved, and the important thing is that it does halt and show an error message in the case of an infinite loop. But I think we’d need to stop dumping all the messages into the DOM at once if we wanted to be truly resilient to the situation.

Issue #1137, opened 2017-10-17
New pristine project should appear in My Projects


outoftime:
Currently, if you’re logged in and click **New Project**, the project you just created does not appear in the **My Projects** dropdown until you modify it in some way. (As long as the new project is pristine, there is simply no project selected in **My Projects**).

While this does reflect the fact that projects are not persisted until they are modified, I think it’s overall confusing. Instead, we should show the new project in the list immediately (but keep the persistence logic the same).

outoftime:
Closed by #1583

PR #1583, opened 2018-10-03
#1137 New pristine project should appear in My Projects

jwang1919:
When creating a new project assigned `updatedAt` attribute to be `new Date()` instead of `null`

### DIFF
--- a/src/components/ProjectPreview.jsx
+++ b/src/components/ProjectPreview.jsx
@@ -1,20 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import moment from 'moment';
-import {t} from 'i18next';
 
 const MAX_LENGTH = 50;
 
 export default function ProjectPreview({preview, project}) {
+  const projectTimestamp = project.updatedAt ?
+    (
+      <div className="project-preview__timestamp">
+        {moment(project.updatedAt).fromNow()}
+      </div>
+    ) : null;
+
   return (
     <div>
       <div className="project-preview__label">
         {preview.slice(0, MAX_LENGTH)}
       </div>
-      <div className="project-preview__timestamp">
-        {project.updatedAt ?
-          moment(project.updatedAt).fromNow() : t('top-bar.pristine-project')}
-      </div>
+      {projectTimestamp}
     </div>
   );
 }
--- a/src/components/ProjectPreview.jsx
+++ b/src/components/ProjectPreview.jsx
@@ -5,19 +5,16 @@ import moment from 'moment';
 const MAX_LENGTH = 50;
 
 export default function ProjectPreview({preview, project}) {
-  const projectTimestamp = project.updatedAt ?
-    (
-      <div className="project-preview__timestamp">
-        {moment(project.updatedAt).fromNow()}
-      </div>
-    ) : null;
-
   return (
     <div>
       <div className="project-preview__label">
         {preview.slice(0, MAX_LENGTH)}
       </div>
-      {projectTimestamp}
+      {project.updatedAt && (
+        <div className="project-preview__timestamp">
+          {moment(project.updatedAt).fromNow()}
+        </div>
+      )}
     </div>
   );
 }
--- a/src/components/ProjectPreview.jsx
+++ b/src/components/ProjectPreview.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import moment from 'moment';
+import {t} from 'i18next';
 
 const MAX_LENGTH = 50;
 
@@ -11,7 +12,8 @@ export default function ProjectPreview({preview, project}) {
         {preview.slice(0, MAX_LENGTH)}
       </div>
       <div className="project-preview__timestamp">
-        {moment(project.updatedAt).fromNow()}
+        {project.updatedAt ?
+          moment(project.updatedAt).fromNow() : t('top-bar.pristine-project')}
       </div>
     </div>
   );
--- a/src/records/Project.js
+++ b/src/records/Project.js
@@ -17,7 +17,7 @@ export default class Project extends Record({
   sources: new Sources(),
   enabledLibraries: new Set(),
   hiddenUIComponents: new Set(['console']),
-  updatedAt: new Date(),
+  updatedAt: null,
   instructions: '',
   externalLocations: new ExternalLocations(),
 }) {
--- a/src/records/Project.js
+++ b/src/records/Project.js
@@ -17,7 +17,7 @@ export default class Project extends Record({
   sources: new Sources(),
   enabledLibraries: new Set(),
   hiddenUIComponents: new Set(['console']),
-  updatedAt: null,
+  updatedAt: new Date(),
   instructions: '',
   externalLocations: new ExternalLocations(),
 }) {
--- a/src/selectors/getAllProjectKeys.js
+++ b/src/selectors/getAllProjectKeys.js
@@ -3,8 +3,15 @@ import {createSelector} from 'reselect';
 export default createSelector(
   state => state.get('projects'),
   projects => projects.
-    filter(project => project.updatedAt).
-    sortBy(project => -project.updatedAt).
+    sort((firstProj, secondProj) => {
+      if (firstProj.updatedAt === null) {
+        return -1;
+      }
+      if (secondProj.updatedAt === null) {
+        return 1;
+      }
+      return secondProj.updatedAt - firstProj.updatedAt;
+    }).
     keySeq().
     toJS(),
 );
--- a/src/selectors/getAllProjectKeys.js
+++ b/src/selectors/getAllProjectKeys.js
@@ -5,8 +5,9 @@ export default createSelector(
   state => state.get('projects'),
   projects => projects.
     sort((
-      {updatedAt: firstProjectUpdatedAt,
-        updatedAt: secondProjectUpdatedAt}) => {
+      {updatedAt: firstProjectUpdatedAt},
+      {updatedAt: secondProjectUpdatedAt},
+    ) => {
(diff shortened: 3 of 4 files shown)


src/records/Project.js
```diff
@@ -17,7 +17,7 @@ export default class Project extends Record({
   sources: new Sources(),
   enabledLibraries: new Set(),
   hiddenUIComponents: new Set(['console']),
-  updatedAt: null,
+  updatedAt: new Date(),
```
outoftime: @jwang1919 this is a good thought, but unfortunately it has far-reaching consequences that I don’t think we want. In particular, the concept of a “pristine project”, namely one that does not have an `updatedAt` set, is used to make decisions in several areas of the codebase, including:

* Whether to persist a project to Firebase in various situations
* Whether to keep a project listed in the **My Projects** menu after you switch away from it

This change inadvertently removes the whole “pristine” concept from the application by making it a state that a project can never be in.

The most obvious undesirable effect can be observed by checking out this branch and repeatedly pressing **New Project**: you’ll note that your projects list fills up with each new project you create. Whereas in the current released version of Popcode, a new project will only stick around in the menu if you modify it before switching away from it.

So, instead of making it so that projects are never pristine, I think we just want to change the logic specific to the **My Projects** menu, telling it to include the current project in the list even if the current project is pristine.

jwang1919:
Okay made the changes. I removed the filter on `project.updatedBy` so that all projects will be displayed. Also sorted the projects so that Pristine Projects have priority.

src/selectors/getAllProjectKeys.js
```diff
@@ -3,8 +3,15 @@ import {createSelector} from 'reselect';
 export default createSelector(
   state => state.get('projects'),
   projects => projects.
-    filter(project => project.updatedAt).
-    sortBy(project => -project.updatedAt).
+    sort((firstProj, secondProj) => {
```
outoftime: We can make this a bit terser by destructuring here

src/components/ProjectPreview.jsx
```diff
@@ -11,7 +12,8 @@ export default function ProjectPreview({preview, project}) {
         {preview.slice(0, MAX_LENGTH)}
       </div>
       <div className="project-preview__timestamp">
-        {moment(project.updatedAt).fromNow()}
+        {project.updatedAt ?
+          moment(project.updatedAt).fromNow() : t('top-bar.pristine-project')}
```
outoftime: The “pristine project” terminology is used in the codebase but isn’t really intended to be user-facing—how about we just omit this element altogether in the pristine case?

src/selectors/getAllProjectKeys.js
```diff
@@ -3,8 +3,15 @@ import {createSelector} from 'reselect';
...
+      if (firstProj.updatedAt === null) {
+        return -1;
```
outoftime: Technically should be 0 if they are _both_ null right?

src/selectors/getAllProjectKeys.js
```diff
@@ -3,8 +3,15 @@ import {createSelector} from 'reselect';
 export default createSelector(
   state => state.get('projects'),
   projects => projects.
-    filter(project => project.updatedAt).
-    sortBy(project => -project.updatedAt).
+    sort((firstProj, secondProj) => {
+      if (firstProj.updatedAt === null) {
```
outoftime: Project style is to prefer `isNull` to the explicit equality check

src/selectors/getAllProjectKeys.js
jwang1919: There shouldn't be a situation with 2 pristine projects

src/selectors/getAllProjectKeys.js
jwang1919: Ah okay can do

src/selectors/getAllProjectKeys.js
jwang1919: Can you be more specific?

src/components/ProjectPreview.jsx
jwang1919: Okay

src/selectors/getAllProjectKeys.js
outoftime: That’s correct now, of course, but do we want to rely on that always being true? Right now this sorting algorithm would produce non-deterministic sorting results if there ever were a situation where there could be two projects with null `updatedAt`s. Seems worth the fairly minimal additional code to handle that case?

src/selectors/getAllProjectKeys.js
```diff
@@ -1,10 +1,26 @@
 import {createSelector} from 'reselect';
+import isNull from 'lodash-es/isNull';
 
 export default createSelector(
   state => state.get('projects'),
   projects => projects.
-    filter(project => project.updatedAt).
-    sortBy(project => -project.updatedAt).
+    sort((...args) => {
+      const [firstProj, secondProj] = args;
```
outoftime: My preference is to just spell full words out unless there is a _really_ good reason not to (so `Project` here and below)

src/components/ProjectPreview.jsx
```diff
@@ -5,14 +5,19 @@ import moment from 'moment';
...
+
   return (
     <div>
       <div className="project-preview__label">
         {preview.slice(0, MAX_LENGTH)}
       </div>
-      <div className="project-preview__timestamp">
-        {moment(project.updatedAt).fromNow()}
-      </div>
+      {projectTimestamp}
```
outoftime: This is something of a moving target but at this point I’ve come around to trying to inline JSX conditionals whenever possible so that it reads more like a “template”, in this case something like:

```jsx
{project.updatedAt && (
  <div className="project-preview__timestamp">
    {moment(project.updatedAt).fromNow()}
  </div>
)}
```

src/selectors/getAllProjectKeys.js
jwang1919: :( and I thought i was being concise with the variable naming.

src/components/ProjectPreview.jsx
jwang1919: Yeah noticed this as well. Tried to follow how conditional JSX was written in other parts. Okay.

src/selectors/getAllProjectKeys.js
```diff
@@ -1,10 +1,26 @@
...
 
 export default createSelector(
   state => state.get('projects'),
   projects => projects.
-    filter(project => project.updatedAt).
-    sortBy(project => -project.updatedAt).
+    sort((...args) => {
+      const [firstProject, secondProject] = args;
+      const firstProjectUpdatedAt = firstProject.updatedAt;
+      const secondProjectUpdatedAt = secondProject.updatedAt;
```
outoftime: Ack I’m sorry @jwang1919 I totally missed this—so there is some destructuring going on here but it was not what I had in mind. Generally we want to use destructuring to make the code more concise, avoid repetition, and make explicit what parts of an object’s shape we care about. So in this case, since we only care about the `updatedAt` field, we can destructure right in the function params and avoid three intermediate variables:

```js
sort(({updatedAt: firstProjectUpdatedAt}, {updatedAt: secondProjectUpdatedAt}) => {
  const isFirstProjectNull = isNull(firstProjectUpdatedAt);
  // etc.
});
```

src/selectors/getAllProjectKeys.js
```diff
@@ -1,10 +1,26 @@
...
+      const isFirstProjectNull = isNull(firstProjectUpdatedAt);
+      const isSecondProjectNull = isNull(secondProjectUpdatedAt);
```
outoftime: I would probably call these `isFirstProjectPristine` etc. to avoid confusion (the project itself is not null).

src/selectors/getAllProjectKeys.js
```diff
@@ -1,10 +1,26 @@
...
+      if (isFirstProjectNull && isSecondProjectNull) {
+        return 0;
```
outoftime: Forget if I mentioned this already but you could nest this inside the below `if` statement, so:

```js
if (isFirstProjectNull) {
  if (isSecondProjectNull) {
    return 0;
  }
  return -1;
}
if (isSecondProjectNull) {
  return 1;
}
// etc.
```

jwang1919: Yeah I was thinking of something like this, but then I thought the code would be unclear this way.

src/selectors/getAllProjectKeys.js
jwang1919: Hmm.. didn't know I can do it this way. Sure can.

outoftime:
@jwang1919 ready for another review?

image

jwang1919:
omg i was just trying to look for this stupid request review button

jwang1919:
@outoftime I don't have that option... I'm clicking on the button and it's not doing anything

src/selectors/getAllProjectKeys.js
```diff
@@ -1,10 +1,25 @@
...
+import isNull from 'lodash-es/isNull';
 
 export default createSelector(
   state => state.get('projects'),
   projects => projects.
-    filter(project => project.updatedAt).
-    sortBy(project => -project.updatedAt).
+    sort((
+      {updatedAt: firstProjectUpdatedAt,
+        updatedAt: secondProjectUpdatedAt}) => {
```
outoftime: Hey dude—testing it out I noticed the sort looked off, and the destructuring here isn’t quite right. You want to destructure the two args as separate objects. (Actually wish the linter caught this as right now you’re destructuring the same property twice, which is clear lint). So:

```js
sort((
  {updatedAt: firstProjectUpdatedAt},
  {updatedAt: secondProjectUpdatedAt},
) => {
```

jwang1919: Oops! First time destructurer here.

PR #1587, opened 2018-10-09
Set the Secure flag on Popcode's session token cookie

andy11:
Sets the [Secure flag](https://www.owasp.org/index.php/SecureFlag) to `true` on the Popcode's session token cookie.

It would be nice to also set the [HttpOnly flag](https://www.owasp.org/index.php/HttpOnly), but this is not supported: https://github.com/js-cookie/js-cookie/issues/344.

NOTE: This should not be considered an important security issue, just a minor nit that's probably worth merging eventually :-)

### DIFF
--- a/src/clients/firebase.js
+++ b/src/clients/firebase.js
@@ -262,7 +262,8 @@ export function setSessionUid() {
     Cookies.set(
       VALID_SESSION_UID_COOKIE,
       uid,
-      {expires: new Date(Date.now() + SESSION_TTL_MS)},
+      {expires: new Date(Date.now() + SESSION_TTL_MS),
+       secure: true},
     );
   }
 }


src/clients/firebase.js
```diff
@@ -262,7 +262,10 @@ export function setSessionUid() {
     Cookies.set(
       VALID_SESSION_UID_COOKIE,
       uid,
-      {expires: new Date(Date.now() + SESSION_TTL_MS)},
+      {
+        expires: new Date(Date.now() + SESSION_TTL_MS),
+        secure: true,
```
outoftime: @andy11 this seems like a reasonable change, but unfortunately it does break behavior in development, where people will typically not be running Popcode over https. So, we should make this behavior configurable, by:

* Introducing an environment variable e.g. `REQUIRE_SECURE_SESSION_COOKIE`
* Passing it through to the client-side code [via Webpack](https://github.com/popcodeorg/popcode/blob/master/webpack.config.js#L64-L73) (with a default of `false`)
* Adding it with value `true` in our Travis configuration (I will do this once the code is ready)
* Exposing it as a boolean via the [`config` module](https://github.com/popcodeorg/popcode/blob/master/src/config.js)

The specific behavior is retaining the user session on refresh—you’ll find that right now, if you log in and then refresh the page with this branch checked out, you’ll be logged back out.

PR #1598, opened 2018-10-22
WIP Element highlighter

wylieconlon:
This continues work from @pwjablonski on #1025

### DIFF
--- a/src/components/Editor.jsx
+++ b/src/components/Editor.jsx
@@ -129,8 +129,26 @@ class Editor extends React.Component {
 
   _startNewSession(source) {
     const session = createAceSessionWithoutWorker(this.props.language, source);
+    const cursor = session.selection.lead;
     session.on('change', () => {
       this.props.onInput(this._editor.getValue());
+    });
+    session.selection.on('changeCursor', () => {
+      this.props.onCursorChange(
+        this._editor.getValue(),
+        cursor,
+        this.props.language,
+      );
+    });
+    this._editor.on('blur', () => {
+      this.props.onEditorBlurred();
+    });
+    this._editor.on('focus', () => {
+      this.props.onEditorFocused(
+        this._editor.getValue(),
+        cursor,
+        this.props.language,
+      );
     });
     session.setAnnotations(this.props.errors);
     this._editor.setSession(session);
@@ -147,6 +165,9 @@ Editor.propTypes = {
   requestedFocusedLine: PropTypes.instanceOf(EditorLocation),
   source: PropTypes.string.isRequired,
   textSizeIsLarge: PropTypes.bool.isRequired,
+  onCursorChange: PropTypes.func.isRequired,
+  onEditorBlurred: PropTypes.func.isRequired,
+  onEditorFocused: PropTypes.func.isRequired,
   onInput: PropTypes.func.isRequired,
   onRequestedLineFocused: PropTypes.func.isRequired,
 };
--- a/src/components/PreviewFrame.jsx
+++ b/src/components/PreviewFrame.jsx
@@ -47,6 +47,7 @@ class PreviewFrame extends React.Component {
     const {consoleEntries, isActive} = this.props;
 
     if (this._channel && isActive) {
+      this._postFocusedSelectorToFrame(this.props.focusedSelector);
       for (const [key, {expression}] of consoleEntries) {
         if (!prevConsoleEntries.has(key) && expression) {
           this._evaluateConsoleExpression(key, expression);
@@ -135,6 +136,13 @@ class PreviewFrame extends React.Component {
     this.props.onConsoleLog(printedValue, compiledProjectKey);
   }
 
+  _postFocusedSelectorToFrame(selector) {
+    this._channel.notify({
+      method: 'highlightElement',
+      params: selector,
+    });
+  }
+
   _attachToFrame(frame) {
     if (!frame) {
       if (this._channel) {
@@ -168,11 +176,16 @@ class PreviewFrame extends React.Component {
 PreviewFrame.propTypes = {
   compiledProject: PropTypes.instanceOf(CompiledProjectRecord).isRequired,
   consoleEntries: ImmutablePropTypes.iterable.isRequired,
+  focusedSelector: PropTypes.string,
   isActive: PropTypes.bool.isRequired,
   onConsoleError: PropTypes.func.isRequired,
   onConsoleLog: PropTypes.func.isRequired,
   onConsoleValue: PropTypes.func.isRequired,
   onRuntimeError: PropTypes.func.isRequired,
+};
+
+PreviewFrame.defaultProps = {
+  focusedSelector: null,
 };
 
 export default PreviewFrame;
--- a/src/containers/EditorsColumn.js
+++ b/src/containers/EditorsColumn.js
@@ -9,10 +9,13 @@ import {
   isTextSizeLarge,
 } from '../selectors';
 import {
+  currentCursorChanged,
   editorFocusedRequestedLine,
   hideComponent,
   updateProjectSource,
   unhideComponent,
+  editorBlurred,
+  editorFocused,
 } from '../actions';
 
 function mapStateToProps(state) {
@@ -41,6 +44,33 @@ function mapDispatchToProps(dispatch) {
     onRequestedLineFocused() {
       dispatch(editorFocusedRequestedLine());
     },
+
+    onEditorCursorChange(source, cursor, language) {
+      dispatch(
+        currentCursorChanged(
+          source,
+          cursor,
+          language,
+        ),
+      );
+    },
+
+    onEditorBlurred() {
+      dispatch(
+        editorBlurred(),
+      );
+    },
+
+    onEditorFocused(source, cursor, language) {
+      dispatch(
+        editorFocused(
+          source,
+          cursor,
+          language,
+        ),
+      );
+    },
+
   };
 }
 
--- a/src/previewSupport/handleElementHighlights.js
+++ b/src/previewSupport/handleElementHighlights.js
@@ -1,0 +1,65 @@
+import throttle from 'lodash-es/throttle';
+
+import channel from './channel';
+
+const RESIZE_THROTTLE = 250;
+let highlightSelector = null;
+
+const handleWindowResize = throttle(() => {
+  updateCovers(highlightSelector);
+}, RESIZE_THROTTLE);
+
+window.addEventListener('resize', handleWindowResize);
+
+function getOffsetFromBody(element) {
+  if (element === document.body) {
+    return {top: 0, left: 0};
+  }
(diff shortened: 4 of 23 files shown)


wylieconlon:
I'm feeling pretty good about this code overall, but I'm not sure what's causing the test failure- it wasn't failing locally for me. I think it's ready for another pass, but it might be better to talk about the overall approach again @outoftime

src/components/Editor.jsx
```diff
@@ -129,9 +129,27 @@ class Editor extends React.Component {
 
   _startNewSession(source) {
     const session = createAceSessionWithoutWorker(this.props.language, source);
+    const cursor = session.selection.lead;
```
outoftime: Seems like `lead` as a property may not be part of the public API—the [docs](https://ace.c9.io/#nav=api&api=selection) only mention `getSelectionLead()`

src/components/Editor.jsx
```diff
@@ -129,9 +129,27 @@ class Editor extends React.Component {
...
     session.on('change', () => {
       this.props.onInput(this._editor.getValue());
     });
+    session.selection.on('changeCursor', () => {
+      this.props.onCursorChange(
+        this._editor.getValue(),
+        cursor,
+        this.props.language,
+      );
+    });
```
outoftime: In general I think facts about UI state should either be completely invisible to application (redux) state, or completely controlled by/synchronized with Redux. So in this case, once application state is concerned with the cursor position, it should also control cursor position. Practically I think this just means we should be taking cursor position and focus as props in this component, watching for changes, and updating the editor components if they’re out of sync.

This also means we can replace the never-terribly-appealing `REQUEST_FOCUSED_LINE` mechanism with a simple action to move the cursor to a particular location at the Redux level.

src/components/PreviewFrame.jsx
```diff
@@ -47,6 +47,7 @@ class PreviewFrame extends React.Component {
     const {consoleEntries, isActive} = this.props;
 
     if (this._channel && isActive) {
+      this._postFocusedSelectorToFrame(this.props.focusedSelector);
```
outoftime: Should we only do this if the focused selector has in fact changed?

src/containers/EditorsColumn.js
```diff
@@ -41,6 +44,33 @@ function mapDispatchToProps(dispatch) {
...
+          source,
+          cursor,
+          language,
+        ),
+      );
+    },
+
+    onEditorBlurred() {
+      dispatch(
+        editorBlurred(),
```
outoftime: Should this still contain a `language` argument? As it is I think we rely on ACE dispatching the `blur` event before the `focus` event in the case where the user clicks from one editor to another.

src/previewSupport/handleElementHighlights.js
```diff
@@ -0,0 +1,65 @@
...
+function addCovers(selector) {
+  const elements = document.querySelectorAll(selector);
+  for (const element of elements) {
+    const cover = document.createElement('div');
+    const rect = element.getBoundingClientRect();
+    let offset = {top: rect.top, left: rect.left};
+    if (element.offsetParent === null) {
+      cover.style.position = 'fixed';
+    } else if (element !== document.body ||
+      element !== document.documentElement) {
```
outoftime: It seems like every element is either not `<html>` or not `<body>`?

src/previewSupport/handleElementHighlights.js
```diff
@@ -0,0 +1,65 @@
...
+      offset = getOffsetFromBody(element);
+    }
+    document.body.appendChild(cover);
+    cover.classList = '__popcode-highlighter';
+    cover.style.left = `${offset.left}px`;
+    cover.style.top = `${offset.top}px`;
+    cover.style.width = `${element.offsetWidth}px`;
+    cover.style.height = `${element.offsetHeight}px`;
+    cover.classList.add('fade');
```
outoftime: Why do we need a second class here? Can’t we use the `__popcode-highlighter` class for all the rules we want to apply?

src/reducers/selectorLocations.js
```diff
@@ -0,0 +1,26 @@
+import Immutable from 'immutable';
+
+import {SelectorLocations} from '../records';
+
+const emptyMap = new Immutable.Map();
+
+export default function reduceProjects(stateIn, action) {
```
outoftime: `reduceSelectorLocations`

outoftime: Also, on my list is to refactor all existing reducers to use `handleActions`, but for now I’d at least like to use it for new reducers (e.g. [`resizableFlex`](https://github.com/popcodeorg/popcode/blob/master/src/reducers/resizableFlex.js))

src/sagas/projects.js
```diff
@@ -55,6 +57,44 @@ export function* changeCurrentProject() {
...
+      /* webpackChunkName: "mainAsync" */
+      '../util/getCssSelectorLocations',
+    ),
+  );
+  return module.default;
+}
+
+export function* parseCurrentProjectSource() {
+  const getJsSelectorLocations = yield call(importGetJsSelectorLocations);
+  const getCssSelectorLocations = yield call(importGetCssSelectorLocations);
```
outoftime: Can use an `all` effect to do these in parallel

src/reducers/selectorLocations.js
```diff
@@ -0,0 +1,26 @@
...
+    state = emptyMap;
+  } else {
+    state = stateIn;
+  }
+
+  switch (action.type) {
+    case 'UPDATE_SELECTOR_LOCATIONS':
+      return state.set(
+        'selectors',
+        new SelectorLocations(action.payload.selectors),
```
outoftime: Although there are some specific exceptions, we usually want to store all non-scalar data in Redux using `Immutable` data structures—so here we should convert each list of locations into an `Immutable` collection (probably a little `Record` representing the location).

src/selectors/getSelectorLocationsForLanguage.js
```diff
@@ -0,0 +1,6 @@
+export default function getSelectorLocationsForLanguage(state, language) {
+  return state.getIn([
+    'selectorLocations',
+    'selectors',
+  ])[language] || null;
```
outoftime: Why not `getIn(['selectorLocations', 'selectors', language])`?

src/reducers/selectorLocations.js
```diff
@@ -0,0 +1,26 @@
...
+  if (stateIn === undefined) {
+    state = emptyMap;
+  } else {
+    state = stateIn;
+  }
+
+  switch (action.type) {
+    case 'UPDATE_SELECTOR_LOCATIONS':
+      return state.set(
+        'selectors',
```
outoftime: I don’t think there’s any need for an additional level of nesting here—isn’t the `SelectorLocations` record itself sufficient to represent this slice of the state?

src/sagas/projects.js
```diff
@@ -131,6 +171,10 @@ export default function* projects() {
       'UPDATE_PROJECT_SOURCE',
       'UPDATE_PROJECT_INSTRUCTIONS',
     ], updateProjectSource),
+    throttle(100, [
+      'UPDATE_PROJECT_SOURCE',
```
outoftime: For `UPDATE_PROJECT_SOURCE`, wouldn’t it be more efficient to call a different saga that only re-parses and calculates the selector locations for that language?

src/util/getCssSelectorLocations.js
```diff
@@ -0,0 +1,19 @@
...
+
+    rootNode.walkRules((rule) => {
+      selectors.push({
+        selector: rule.selector,
+        loc: rule.source,
+      });
+    });
+
+    return selectors;
+  } catch (e) {
```
outoftime: Should we be looking for a specific error that indicates invalid CSS syntax? I wouldn’t want to inadvertently swallow an error thrown by a bug in our code.

src/util/getJsSelectorLocations.js
```diff
@@ -0,0 +1,38 @@
...
+            loc: expressionLoc,
+            selector: selector.value,
+          });
+        }
+      },
+    };
+    traverse(ast, visitor, null);
+
+    return foundMatches;
+  } catch (e) {
```
outoftime: Same question about blanket `catch` here

src/util/getJsSelectorLocations.js
```diff
@@ -0,0 +1,38 @@
...
+      CallExpression(path) {
+        // Only matches jquery or querySelector with string literals
+        const expressionLoc = path.node.loc;
+        const [selector] = path.node.arguments;
+        const callee = path.get('callee').toString();
+
+        if (
+          (
+            callee === '$' ||
+            callee.indexOf('querySelector') !== -1
```
outoftime: I would use `callee.startsWith` (this is a fairly new method but should be polyfilled)

templates/highlighter.css
```diff
@@ -0,0 +1,21 @@
+.__popcode-highlighter {
+  z-index: 2000000;
```
outoftime: I’m not an enormous fan of arbitrarily large `z-indexes` to accomplish “definitely above everything else”—when we’re constructing the highlighter, can we just find the topmost ancestor element that has a `z-index`, and set the highighter’s `z-index` to one plus that?

templates/highlighter.css
```diff
@@ -0,0 +1,21 @@
...
+  box-sizing: content-box;
+  background-color: #00b8ff;
+  opacity: .5;
+  transition-property: opacity, background-color;
+  transition-duration: 300ms, 2.3s;
+}
+
+.__popcode-highlighter.fade {
+  background-color: #ffffff00;
+}
```
outoftime: I love this effect! Seems like it would be a little more intuitive to describe it using an animation rather than a transition, though? I found it confusing when I initially noticed that the highlighter was being given two class names at the same time…

src/reducers/selectorLocations.js
outoftime: Also generally we want to have unit tests for all reducers and sagas

test/unit/util/getCssSelectorLocations.js
```diff
@@ -0,0 +1,23 @@
...
+
+import getCssSelectorLocations from
+  '../../../src/util/getCssSelectorLocations';
+import {css} from '../../data/acceptance.json';
+
+const [source] = css;
+
+test('finds all selectors in css', (assert) => {
+  const selectors = getCssSelectorLocations(source);
+  assert.deepEqual(selectors.map(s => s.selector), [
```
outoftime: You can do the map a little more cutely with lodash, e.g. `map(selectors, 'selector')`

outoftime:
Great stuff! Feels like very strong progress and the structure is mostly there—only major thing would be to make the cursor/focus state fully controlled by Redux rather than just a one-way data flow. Can’t wait to release this, it’s going to be huge!!

PR #1599, opened 2018-10-23
Validate that markup is inside <html> or <body>

### LINKED ISSUE #1581: No error for text/tags outside the body
No error for putting plain text outside of the body, as well as plain text or other tags outside of the html tags.
Would also be great if we could detect plain text (not in an enclosing tag) _inside_ the body, tho that seems more difficult. A lot of students struggle with this, since their text does show up on the page, they don't see a problem.

See: https://popcode.org/?snapshot=0a260e86-b67b-4dc4-b2d1-d3d5e145db98

wylieconlon:
Closes #1581 

screenshot 2018-10-23 12 47 11


package-lock.json
```diff
@@ -2569,7 +2569,7 @@
     },
     "bl": {
       "version": "1.1.2",
-      "resolved": "https://registry.npmjs.org/bl/-/bl-1.1.2.tgz",
+      "resolved": "http://registry.npmjs.org/bl/-/bl-1.1.2.tgz",
```
outoftime: Can you check this out from master?

src/validations/html/rules/MarkupOutsideContainer.js
```diff
@@ -0,0 +1,75 @@
...
+//
+// * Any tags or text outside the <html> element
+// * Any text that is inside the <html> element but not part
+//   of the <head> or <body> tags
+
+import findLastIndex from 'lodash-es/findLastIndex';
+
+import Code from './Code';
+
+export default class MismatchedTag {
```
outoftime: `TopLevelDocumentStructure`, `DocumentContainment`, etc.? Also be sure to name the class and the file the same

src/validations/html/rules/MarkupOutsideContainer.js
```diff
@@ -0,0 +1,75 @@
...
+  constructor() {
+    this._openTagStack = [];
+    this._locationsWithOutsideTags = [];
+  }
+
+  openTag(location, {name}) {
+    if (
+      this._isOutsideAcceptableContainer() &&
```
outoftime: Let’s decompose this into something like `_isOutsideDocumentElement()` and `isDirectChildOfDocumentElement()` and then pair each directly with the below acceptable tag names

src/validations/html/rules/MarkupOutsideContainer.js
```diff
@@ -0,0 +1,75 @@
...
+      this._openTagStack.splice(openIndex);
+    }
+  }
+
+  text(location, text) {
+    const trimmedText = text.trim();
+    if (trimmedText && this._isOutsideAcceptableContainer()) {
+      this._locationsWithOutsideTags.push({
+        location,
+        type: 'text',
```
outoftime: I don’t have a strong opinion as to exactly what the right place is to make the split, but we should avoid having the word `text` be bold/monospace in the error message (the actual copy that it currently produces happens to be fine)

src/validations/html/rules/MarkupOutsideContainer.js
```diff
@@ -0,0 +1,75 @@
...
+
+    const isOpened = openIndex >= 0;
+    if (isOpened) {
+      this._openTagStack.splice(openIndex);
+    }
+  }
+
+  text(location, text) {
+    const trimmedText = text.trim();
+    if (trimmedText && this._isOutsideAcceptableContainer()) {
```
outoftime: We should show an error for text nodes directly inside `<head>` as well

src/validations/html/rules/MarkupOutsideContainer.js
wylieconlon: @outoftime I've been thinking about the actual error messages a bit more, and I think my preference is to show the following two messages:

```
You are not allowed to write text outside of the <body> tag
You are not allowed to place `<p>` tags outside of the <body> tag
```

I think it will always indicate that you should be placing it inside the `<body>` tag. I think this also simplifies the error message names to `invalid-text-outside-body` and `invalid-tag-outside-body`

Thoughts on this naming?

src/validations/html/rules/MarkupOutsideContainer.js
wylieconlon: Can discuss below, but for now I'm going with `NodeOutsideBody` because this is really an error about things that are only allowed in the body

src/validations/html/rules/MarkupOutsideContainer.js
wylieconlon: Done

src/validations/html/rules/MarkupOutsideContainer.js
wylieconlon: Done

wylieconlon:
I think this is ready for another review @outoftime

src/validations/html/rules/NodeOutsideBody.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  text(location, text) {
+    const trimmedText = text.trim();
+    if (trimmedText && (
+      this._isTopLevel() ||
+      this._isInsideHtmlTag() ||
+      this._isDirectChildOfHead()
+    )) {
+      const matchingWhitespaces = text.match(/^(\W+)/gu);
```
outoftime: Don’t we want `\s` rather than `\W`? I think the latter will match things like punctuation as well as whitespace

src/validations/html/rules/NodeOutsideBody.js
```diff
@@ -0,0 +1,101 @@
...
+      let replacementLocation = location;
```
outoftime: Small thing, but I prefer to avoid `let` wherever possible—here I would probably just do an `if`/`else` and `push` the right thing in each branch. Don’t feel strongly about it though.

wylieconlon: Updated

wylieconlon:
@outoftime This is ready to review!

PR #1606, opened 2018-11-05
Upgrade to Babel 7

outoftime:
Upgrades to Babel 7:

* Replace all relevant packages with latest `@babel/` namespaced equivalents
* Extract base babel configuration into `babel.config.js`, which is now a first-class feature; this is imported into `webpack.config.js`


wylieconlon:
I will take this over as I'm working on some dependent features

PR #1615, opened 2018-11-25
[Shen picking up from cricklet] adding auto-indentation to popcode via `js-beautify`

dshen6:
Picking up from https://github.com/popcodeorg/popcode/pull/1215.

I made the linter happy, and fixed merge conflicts

Re: selection (https://github.com/popcodeorg/popcode/pull/1215#issuecomment-418926412), I'll dig deeper, but it doesn't automatically format all three sections(html,css,js) any more - it just formats the currently focused section. (gifs on the old pr show this behavior, but I couldn't get a working sandbox of that). From my testing, the cursor selection seems fine, adding an arg to `setValue()` isn't necessary. See gif.

Is this ready to merge? I'm *very* keen to start telling students "`ctrl + i` and see what happens :D"


dshen6:
prettify

src/components/Editor.jsx
```diff
@@ -137,6 +163,64 @@ class Editor extends React.Component {
...
+    );
+
+    selection.setSelectionRange(newRange);
+
+    event.preventDefault();
+  }
+
+  _handleKeyPress(event) {
+    if (this.Beautify &&
+      event.key === 'i' && (event.metaKey || event.ctrlKey) &&
```
razzius: I think `|| event.ctrlKey` is redundant since we also check `&& !event.ctrlKey` on the next line

src/reducers/projects.js
```diff
@@ -118,6 +118,14 @@ export default function reduceProjects(stateIn, action) {
         action.meta.timestamp,
       );
 
+    case 'PROJECT_BEAUTIFIED':
+      return state.setIn([action.payload.projectKey, 'sources'],
```
dshen6: `[action.payload.projectKey.sources]` doesn't actually do what we want - i think we still need to hard code 'sources' here, since the structure is `projectkey: { sources: {...}}`

src/components/Editor.jsx
```diff
@@ -59,7 +70,7 @@ class Editor extends React.Component {
     if (projectKey !== prevProjectKey) {
       this._startNewSession(source);
     } else if (source !== prevSource && source !== this._editor.getValue()) {
-      this._editor.setValue(source);
+      this._editor.setValue(source, 1);
```
dshen6: here's where we set selection to end, so that it's not all highlighted.

src/sagas/projects.js
```diff
@@ -87,6 +92,36 @@ export function* updateProjectSource() {
...
+export function* loadAndBeautifyProjectSource() {
+  const beautify = yield call(importBeautify);
+
+  const state = yield select();
+  const currentProject = getCurrentProject(state);
+
+  const beautifiedSourcesMap = new Map();
+  for (const language of Reflect.ownKeys(currentProject.sources)) {
+    const source = currentProject.sources[language];
+    const newSource = format(
```
dshen6: oh i'm reading more into sagas, i think i should actually be forking here, hold on

src/util/formatter.js
```diff
@@ -0,0 +1,48 @@
...
+  unescape_strings: false,
+  wrap_line_length: 0,
+  e4x: false,
+  comma_first: false,
+  operator_position: 'before-newline',
+  /* don't add new lines before head,body,/html https://github.com/beautify-web/js-beautify */
+  extra_liners: [],
+};
+
+export function format(Beautify, code, mode, opts = {}) {
```
outoftime: I know this is kind of picky, but I’m wary of having two variable/parameter names which are basically “the same thing but one is abbreviated”. I would claim there is always a descriptive way to distinguish them! In this case I might call the parameter `optionsIn` or `optionOverrides`.

src/sagas/projects.js
```diff
@@ -87,6 +92,38 @@ export function* updateProjectSource() {
...
+  return retryingFailedImports(
+    () => import(
+      /* webpackChunkName: "mainAsync" */
+      'js-beautify',
+    ),
+  );
+}
+
+export function* loadAndBeautifyProjectSource() {
+  const beautify = yield call(importBeautify);
```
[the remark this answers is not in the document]
dshen6: i was following importing `validations` in `errors.js`

src/sagas/projects.js
```diff
@@ -87,6 +92,38 @@ export function* updateProjectSource() {
...
+        beautify,
+        source,
+        language);
+    }),
+  );
+  const sourcesMap = new Map();
+  allFormattedSources.map(({formatted, language}) => {
+    sourcesMap.set(language, formatted);
+  });
+  const projectSources = new ProjectSources(sourcesMap);
```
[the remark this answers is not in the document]
dshen6: I think it's a good rule to follow, and matters more when you have more data than this particular instance

src/sagas/projects.js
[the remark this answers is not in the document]
outoftime: Got it—yeah it’s kind of a judgement call in each case, but in general I try to ask the question “what is the biggest chunk I can bite off with an async load?”—so in this case the `formatter` module itself can be loaded asynchronously with no major downsides that I can think of, and I think would make the interface a little simpler since you no longer need to pass in a `beautify` instance?

src/clients/googleAnalytics.js
```diff
@@ -1,4 +1,4 @@
-import ReactGA from 'react-ga';
+import {initialize, pageview} from 'react-ga';
```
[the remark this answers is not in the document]
dshen6: fixed

package.json
```diff
@@ -176,15 +176,16 @@
   "bugs": "https://trello.com/b/ONaFg6wh/popcode",
   "license": "MIT",
   "dependencies": {
-    "@firebase/app": "^0.3.4",
+    "@firebase/app": "^0.3.5",
```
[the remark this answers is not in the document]
dshen6: done

src/components/Editor.jsx
```diff
@@ -1,19 +1,15 @@
...
 
-import {createAceEditor, createAceSessionWithoutWorker} from '../util/ace';
 import {EditorLocation} from '../records';
+import {createAceEditor, createAceSessionWithoutWorker} from '../util/ace';
 
-import 'brace/ext/searchbox';
-import 'brace/mode/html';
-import 'brace/mode/css';
-import 'brace/mode/javascript';
-import 'brace/theme/monokai';
```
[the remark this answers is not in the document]
dshen6: not sure, and i'm not sure how to figure it out. when i add them back in it says: 

Could not find a declaration file for module 'brace/mode/html'

src/sagas/projects.js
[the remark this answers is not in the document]
dshen6: i don't fully understand how a static load works. i understand the entire point of this is to have a dynamic import, but when is the code to load Beautify actually run?

src/components/Editor.jsx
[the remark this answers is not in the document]
dshen6: yea looks like this dates back from before this particular pr

dshen6:
@outoftime ah something is wrong with the way npm is installing dependencies. how do i access the actual error logs?

outoftime:
hey @dshen6 looking forward to taking a look at all the latest changes soon but in the meantime, I just merged in a [freshly-generated `package-lock.json`](https://github.com/popcodeorg/popcode/pull/1634) so would you mind merging that in and then letting `npm` update it with the additional dependencies in this PR?

src/components/Editor.jsx
```diff
@@ -137,6 +143,14 @@ class Editor extends React.Component {
     this._editor.moveCursorTo(0, 0);
     this._resizeEditor();
   }
+
+  _handleKeyPress(event) {
+    if (
+      event.key === 'i' && event.metaKey &&
+      !event.altKey && !event.ctrlKey) {
+      this.props.onAutoFormat();
```
outoftime: Hey so I’m not sure if this is an issue outside of Linux, but on my laptop when I do meta-I it indents the text but also inserts an `i` character at the end of the code. Relatedly, I think the command should probably be ctrl-I on Windows and Linux, and Command-I on Mac OS (which I am guessing it already is).

Playing around with it, I think we can call `preventDefault()` on the event, which deals with the insertion issue. However I wonder if it would make more sense to use ACE’s [`commands.addCommand()`](https://ace.c9.io/demo/keyboard_shortcuts.html) method, which adds the keyboard shortcut at the ACE level. It looks like that would make it easier to semantically specify different Mac and non-Mac keyboard shortcuts, and I’m guessing would implicitly handle the character-insertion bug. WDYT?

(Weirdly that method is not actually mentioned in the ACE documentation as far as I can tell, but it definitely exists and the linked example is on the ACE site, so it seems legit to use)

src/sagas/projects.js
```diff
@@ -87,6 +90,27 @@ export function* updateProjectSource() {
   yield* saveCurrentProject();
 }
 
+export function* loadAndBeautifyProjectSource() {
+  const state = yield select();
+  const currentProject = getCurrentProject(state);
+  const allFormattedSources = yield all(
+    Reflect.ownKeys(currentProject.sources).map((language) => {
+      const source = currentProject.sources[language];
```
outoftime: Why not use lodash `map(currentProject.sources)` here? That will invoke the iterator with `(source, language)` as arguments.

src/sagas/projects.js
```diff
@@ -87,6 +90,27 @@ export function* updateProjectSource() {
   yield* saveCurrentProject();
 }
 
+export function* loadAndBeautifyProjectSource() {
+  const state = yield select();
+  const currentProject = getCurrentProject(state);
```
outoftime: You can collapse these two lines into one by doing `const currentProject = yield select(getCurrentProject)`

src/sagas/projects.js
```diff
@@ -87,6 +90,27 @@ export function* updateProjectSource() {
...
+      const source = currentProject.sources[language];
+      return call(
+        beautifySource,
+        source,
+        language,
+      );
+    }),
+  );
+  const sourcesMap = new Map();
+  allFormattedSources.map(({formatted, language}) => {
```
outoftime: No need to use `map()` here since you are not capturing the return value.

src/sagas/projects.js
```diff
@@ -87,6 +90,27 @@ export function* updateProjectSource() {
...
+    Reflect.ownKeys(currentProject.sources).map((language) => {
+      const source = currentProject.sources[language];
+      return call(
+        beautifySource,
+        source,
+        language,
+      );
+    }),
+  );
+  const sourcesMap = new Map();
```
outoftime: I think there’s a good argument that a `Map` is a more appropriate data structure for this use case than a plain object, but given that we use plain objects in action payloads to communicate the same overall data structure elsewhere in the code, I think it’s best to be consistent here. I think the code in the reducer will be fine with it either way. Thoughts?

src/util/beautifySource.js
```diff
@@ -0,0 +1,57 @@
...
+  /* don't add new lines before head,body,/html https://github.com/beautify-web/js-beautify */
+  extra_liners: [],
+};
+
+const importingBeautify = importBeautify();
+
+async function importBeautify() {
+  return retryingFailedImports(() => import(
+    /* webpackChunkName: "mainAsync" */
+    'js-beautify',
```
outoftime: You’ll find that the latest version of the babel parser (which is now a dependency) considers the trailing comma here a syntax error (which it is according to the language spec). Unfortunately eslint has not caught up with that yet, so my temporary solution is just to remove the trailing comma and add an `eslint-disable-line` (you’ll see this pattern in all other dynamic imports in the code now).

src/util/beautifySource.js
```diff
@@ -0,0 +1,57 @@
...
+  ));
+}
+
+export default async function beautifySource(code, mode) {
+  const Beautify = await importingBeautify;
+  return {
+    formatted: formatCode(Beautify, code, mode),
+    language: mode,
```
outoftime: Still wondering why we need to return the `language` property here given that it is also an argument to the function? (Meaning it is information the caller of this function must already have)

dshen6: yea, i'm wasn't sure how to wire it up in `loadAndBeautifyProjectSource`, which uses `call` - tried to have this object creation logic here, but you can't `return { formatted: call(...) }`

dshen6:
yea, would you mind pushing the package-lock fix? thanks!

src/sagas/projects.js
dshen6: yea i remember trying an object and it not working for me, but seems to work now 🤷‍♀️

src/sagas/projects.js
dshen6: i think i followed suit from errors.js, should probably change that one too

src/components/Editor.jsx
dshen6: ace way seems good i'll look into it

src/components/Editor.jsx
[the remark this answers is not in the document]
dshen6: i mean that this was never related to these beautify changes at all, and this is just general cleanup

outoftime:
OK just pushed!

dshen6:
screen shot 2019-02-16 at 4 17 33 pm

outoftime:
@dshen6 `npm install` !

dshen6:
ah, that works, but it adds a lot of files to the package-lock :(

src/components/Editor.jsx
dshen6: got it working, mind testing on linux?

outoftime:
Works real nice on linux now!! :+1:

src/clients/googleAnalytics.js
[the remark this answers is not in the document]
dshen6: oh this was to fix a linter

src/sagas/projects.js
```diff
@@ -87,6 +91,26 @@ export function* updateProjectSource() {
...
+    ),
+  );
+  const sourcesMap = {};
+  allFormattedSources.forEach(({formatted, language}) => {
+    sourcesMap[language] = formatted;
+  });
+  yield put(projectBeautified(currentProject.projectKey, sourcesMap));
+}
+
+
```
outoftime: One extra whitespace line

src/sagas/projects.js
```diff
@@ -87,6 +91,26 @@ export function* updateProjectSource() {
...
+  const allFormattedSources = yield all(
+    map(
+      currentProject.sources,
+      (source, language) => call(
+        beautifySource,
+        source,
+        language,
+      ),
+    ),
+  );
```
outoftime:
```js
const sourcesMap = yield all(
  reduce(
    currentProject.sources,
    (calls, source, language) => Object.assign(calls, {
      [language]: call(
        beautifySource,
        source,
        language,
      )
    }),
    {},
  ),
);
```

PR #1624, opened 2018-12-13
Add a docker-compose override file for bind mounting dependencies

outoftime:


outoftime:
This seems to be already merged…

PR #1630, opened 2019-01-12
Use Babel to transpile modern JS for older browsers

### LINKED ISSUE #1620: Compile modern JavaScript for compatibility with older browsers
Currently, Popcode doesn’t prevent the use of modern JavaScript (ES6+) syntax and features—[this snapshot](https://popcode.org/?snapshot=a04ed3c3-c4ad-40d8-adda-83831b669602) works just fine on modern Chrome, for instance—but it doesn’t do anything to make newer JavaScript work on older browsers. Since many of our students are stuck on old browser versions, this creates an inconsistency in what kind of code they can actually write, and presents an obstacle to teaching modern JavaScript.

So, let’s pull Babel into the application and compile the students’ JavaScript down to a spec that the current browser can handle. A few thoughts/challenges:

* Ideally, we’d only do as much compilation as the current browser requires. In theory this should be entirely possible with `babel-preset-env`, though we will need to be careful about not making the bundle enormous with a big browser compatibility data file. If that’s unavoidable with `preset-env`, we can just compile everything down to a version that works on all browsers that Popcode supports (which is of course what Popcode’s own application runtime does)
* We should make sure to include the runtime polyfills as well as compiling the code, although I think that’s already baked in to support the `previewSupport` runtime.
* This may be as good a time as any to introduce web workers to the Popcode universe—I suspect the compilation step could have substantial performance implications if run synchronously in the document’s execution context. I’ve been wanting to move things like validations to web workers for a while anyway.
* One thing to look out for is making sure we are still reporting the correct line number in the student’s JS source for runtime errors. This will probably require generating inline source maps. If this is at all complicated, we can punt it to a followup (but should probably do the initial release in experimental mode)
* Once we have the general framework in place, also as a followup, we should do the same thing for CSS, adding automatic browser prefixing and support for compiling newer CSS features for older-browser support where possible

wylieconlon:
This is now using the preferred technique:

1. Use webpack `val-loader` to load dependencies at build time
1. Get the list of babel/preset-env plugins by running babel on an empty string
1. Generate a stub function which runs babel with the predefined plugins

I'm not sure there's a way to get jest to test against the `val-loader` generated code, this is potentially causing the tests to not be as complete as otherwise.

Closes #1620


src/util/compileProject.js
```diff
@@ -165,7 +165,37 @@ async function addJavascript(
...
+        [
+          presetEnv,
+          {
+            targets: 'chrome > 23',
+          },
+        ],
+      ],
+    });
+  } catch (e) {
+    return;
```
wylieconlon: We are catching syntax errors here before outputting it- this is probably not the ideal behavior, but as I write tests I'll flesh out the failure case

outoftime: hmm, I think the validator should detect all syntax errors? and if there is a static validation error then we don’t run `compileProject` at all? might be some nuance i am missing tho

wylieconlon:
@outoftime It seems like the esprima parser we're using doesn't understand a few kinds of syntax. Two options I see for resolving the issues, my preference is a:

a) Ignore the parse errors for now, and follow up with a switch from esprima to @babel/parser and validation tests
b) Update the JS validations in this PR and add more tests

Here are some of the errors I've found:

```
const { a, ...rest }  = { a: 'hello', c: 'world'};
```

`JavaScript doesn’t expect a ... where you wrote one here.`

```
async function hello() {}
```

```
On line 1:
It looks like there's something on this line that doesn't belong there

On line 1:
You haven't declared async. Try: var async;
```

wylieconlon:
@outoftime Esprima is working fine for regular ES5 syntax, but seems not to understand some of the more recent syntax like the spread operator and async await. I haven't investigated the cause, so it could be as simple as flipping a flag, or it could be a larger issue- and I think it's probably a larger issue. To include your suggestion:

a) Merge this PR first, then deal with validation and error reporting later
b) Update this PR to include validation and error reporting of ES6+ syntax
c) Work on the system to validate ES6+ syntax which is merged before merging this PR

My preference is a) because I think these two things are independent. This PR is going to work fine for most new syntax, such as `let` and `const`, without causing issues, and I think we can improve that later.

outoftime:
@wylieconlon OK so what I’m not clear on is whether it is practical to just ignore esprima errors triggered by modern ES features, while keeping all the existing esprima validation working. If yes then I am definitely down for (a)

outoftime:
(Also I do think it’s worth a quick check to make sure there isn’t a simple way to get esprima to do the right thing)

wylieconlon:
Unfortunately I don't see a chart comparing features directly, but you can see for yourself at https://astexplorer.net/ - esprima doesn't seem to understand the object spread operator (it does handle array spreads), but it works fine for most of the other syntax I threw at it. So maybe it's one of the other tools in the popcode validation chain that doesn't like `async` functions or the `**` exponentiation operator.

wylieconlon:
The generated Babel bundle is huge largely due to its use of `preset-env`- maybe by specifically transpiling the ES6+ features we want, we could reduce the bundle size?

outoftime:
Seems worth looking into—specifically we would need to ship with the same transforms that `preset-env` actually uses to compile the bundle. I wonder if there is any way to introspect that?

wylieconlon:
Sure, so there is a debug option for preset-env, so if we wanted to actually copy it we could use all of these dependencies:

```
@babel/preset-env: `DEBUG` option

Using targets:
{
  "chrome": "23"
}

Using modules transform: auto

Using plugins:
  transform-template-literals { "chrome":"23" }
  transform-literals { "chrome":"23" }
  transform-function-name { "chrome":"23" }
  transform-arrow-functions { "chrome":"23" }
  transform-block-scoped-functions { "chrome":"23" }
  transform-classes { "chrome":"23" }
  transform-object-super { "chrome":"23" }
  transform-shorthand-properties { "chrome":"23" }
  transform-duplicate-keys { "chrome":"23" }
  transform-computed-properties { "chrome":"23" }
  transform-for-of { "chrome":"23" }
  transform-sticky-regex { "chrome":"23" }
  transform-dotall-regex { "chrome":"23" }
  transform-unicode-regex { "chrome":"23" }
  transform-spread { "chrome":"23" }
  transform-parameters { "chrome":"23" }
  transform-destructuring { "chrome":"23" }
  transform-block-scoping { "chrome":"23" }
  transform-typeof-symbol { "chrome":"23" }
  transform-new-target { "chrome":"23" }
  transform-regenerator { "chrome":"23" }
  transform-exponentiation-operator { "chrome":"23" }
  transform-async-to-generator { "chrome":"23" }
  proposal-async-generator-functions { "chrome":"23" }
  proposal-object-rest-spread { "chrome":"23" }
  proposal-unicode-property-regex { "chrome":"23" }
  proposal-json-strings { "chrome":"23" }
  proposal-optional-catch-binding { "chrome":"23" }
  transform-named-capturing-groups-regex { "chrome":"23" }

Using polyfills: No polyfills were added, since the `useBuiltIns` option was not set.
ok 165 should be truthy
# javascript is not mangled for new browsers
@babel/preset-env: `DEBUG` option

Using targets:
{
  "chrome": "71"
}

Using modules transform: auto

Using plugins:
  syntax-async-generators { "chrome":"71" }
  syntax-object-rest-spread { "chrome":"71" }
  syntax-json-strings { "chrome":"71" }
  syntax-optional-catch-binding { "chrome":"71" }

Using polyfills: No polyfills were added, since the `useBuiltIns` option was not set.
```

wylieconlon:
@outoftime Any feedback on the possible approach of manually using the transformations listed above?

outoftime:
@wylieconlon yes—I definitely think we should, although we should find some way to make it at least semi-automatic, to wit:

* Write code to generate a module file that imports the appropriate transform/syntax modules, based on the output of `preset-env`’s debug mode. That module should also contain (probably as a named export) a checksum of the contents of `config/browsers.json`
* Add a script to the build process that verifies that the checksum exported from the generated module matches the checksum of the current contents of `config/browsers.json`

wylieconlon:
I think the current state of this code is that it's functional but not simple. Going to work on a separate PR to attempt to simplify

wylieconlon:
@outoftime This is now updated, can you please take a look?

outoftime:
Yep can you smash that Request Review button?

wylieconlon:
With sourcemaps, errors now report correctly in the devtools console but not the popcode console:

Screenshot 2019-04-21 17 31 28

wylieconlon:
Also, bundled size of the async `babel` code is ~350kb gzipped, 1.5Mb without Gzip. It's still pretty large, and I bet we could reduce it further with either:

- Bumping up our minimum supported browsers
- Selectively excluding transformations that students are unlikely to use, such as module-related ones

src/services/babel-browser.gen.js
```diff
@@ -0,0 +1,68 @@
...
+    ],
+  });
+
+  const keys = result.options.plugins.map(
+    plugin => plugin.key,
+  ).filter(
+    key => key.indexOf('/') !== 0,
+  );
+
+  const plugins = keys.map(key => `require("@babel/plugin-${key}")`).
```
outoftime: I suppose it’s ultimately neither here nor there, but if we fully lean into the premise that a `.gen.js` file is emitting a `code` property that stands in for actual source code we would put in a typical `.js` file, wouldn’t we want to do static `import`s here?

src/services/babel-browser.gen.js
```diff
@@ -0,0 +1,68 @@
...
+    join(',\n');
+
+  return {
+    code: `
+var babel = require('@babel/core');
```
outoftime: We could at _least_ do a static import here, right?

src/services/babel-browser.gen.js
```diff
@@ -0,0 +1,68 @@
...
+    sourceMaps: 'inline',
+    sourceFileName: 'popcodePreview.js',
+  }).then(function (result) {
+    return result.code;
+  }).catch(function (error) {
+    throw new Error(error);
+  });
+}
+
+module.exports = { babelWithEnv: babelWithEnv };
```
outoftime: Similarly we could do an ES6 export here?

src/util/compileProject.js
```diff
@@ -156,16 +154,28 @@ async function addPreviewSupportScript(doc) {
...
+  let source = javascript;
+
+  const {babelWithEnv} = await retryingFailedImports(
+    () => import(
+      /* webpackChunkName: "babel" */
+      '../services/babel-browser.gen' // eslint-disable-line comma-dangle
+    ),
+  );
+
+  source = await babelWithEnv(source);
```
outoftime: Seems like we could just do `const source = await babelWithEnv(javascript)` here? (No need to reassign `source`)

outoftime:
> Selectively excluding transformations that students are unlikely to use, such as module-related ones

Yeah we should definitely do this since Popcode simply does not support using modules…

outoftime:
> Bumping up our minimum supported browsers

Yeah definitely! I am down to bump minimum versions up to the empirically minimum version people are actually using Popcode with regularly any time.

wylieconlon:
@outoftime I've found some issues with errors and sourcemapping that are making this a larger issue:

1. The current mechanism to get the line number is the `lineno` parameter of `window.onerror`, which does not respect sourcemaps
2. Browsers also ignore sourcemaps when generating `error.stack`- only the devtools console uses this
3. Almost all libraries related to sourcemaps depend on https://github.com/mozilla/source-map
4. mozilla/source-map requires a piece of WASM code that is a core rust library they use
5. WASM is not supported by our target browsers.
6. The WASM file can either be loaded from unpkg or from ArrayBuffer
7. The sourcemap format itself is pretty small, and the only zero-dependency library I can see is this one: https://github.com/Rich-Harris/sourcemap-codec
8. Even if we use the `sourcemap-codec` library, we will need to write some higher level helper functions to modify sourcemaps.

So overall I think the most promising direction is to build very small helper functions directly on top of the sourcemap spec- I'm worried about introducing a dependency on WASM.

wylieconlon:
Depends on the code from https://github.com/popcodeorg/loop-breaker/pull/10

config/browsers.json
```diff
@@ -1,6 +1,6 @@
 {
   "firefox": "28",
-  "chrome": "23",
-  "chromium": "23",
+  "chrome": "38",
+  "chromium": "38",
```
wylieconlon: Chrome 38 is the lowest version that's been used with any frequency since January, according to analytics. It's also a version that bumped up a lot of ES6 standard library support.

src/components/PreviewFrame.jsx
```diff
@@ -85,17 +85,15 @@ class PreviewFrame extends React.Component {
   }
 
   _runtimeErrorLineOffset() {
-    const firstSourceLine = this.props.compiledProject.source.
-      split('\n').indexOf(sourceDelimiter) + 2;
-
-    return firstSourceLine - 1;
+    return this.props.compiledProject.source.
+      split('\n').findIndex(l => l.indexOf(sourceDelimiter) !== -1) + 1;
```
wylieconlon: Still using the delimiter because the line numbers come from `outerHTML` instead of from source directly

src/components/PreviewFrame.jsx
```diff
@@ -105,15 +103,38 @@ class PreviewFrame extends React.Component {
...
+      }
     }
 
     this.props.onRuntimeError({
       reason: normalizedError.type,
       text: normalizedError.message,
       raw: normalizedError.message,
-      row: line - 1,
-      column: error.column,
+      // Convert back to zero indexed
```
outoftime: I think this is pretty self-explanatory given your descriptive variable naming

src/components/PreviewFrame.jsx
outoftime: Just to make sure I’m understanding—the notion here is that us having a sourcemap doesn’t change the browser’s behavior with respect to line numbers in the error, which is who-knows-what since this code is in a `<script>` tag, right? What was good for the goose is still good for the gander?

src/services/babel-browser.gen.js
```diff
@@ -0,0 +1,78 @@
...
+  const browserListForBabel = {};
+
+  for (const browser in browserList) {
+    if (browser !== 'chromium') {
+      browserListForBabel[browser] = browserList[browser];
+    }
+  }
+
+  // preset-env doesn't give an introspection API, so we have to get it from
+  // the metadata on transformed code
```
outoftime: Can we self-document instead of commenting? E.g. wrap this in a function called `extractPluginsFromNoopCompilationResult` or something? (Extracting a function or two would probably be good from a composition standpoint anyway)

src/services/babel-browser.gen.js
```diff
@@ -0,0 +1,78 @@
...
+  const result = babel.transformSync('', {
+    sourceType: 'script',
+    presets: [
+      [
+        presetEnv,
+        {
+          targets: browserListForBabel,
+          // The module for regenerator causes an import error, so we ignore it
```
outoftime: Ibid. I think we could self-document by defining a constant like `ERRORING_TARGETS_THAT_ARE_SAFE_TO_EXCLUDE`?

src/services/babel-browser.gen.js
```diff
@@ -0,0 +1,78 @@
...
+function babelWithEnv(source, inputSourceMap) {
+  return transformAsync(source, {
+    sourceType: 'script',
+    plugins: [
+${pluginNames}
+],
+    sourceMaps: 'both',
+    sourceFileName: 'popcodePreview.js',
+    inputSourceMap,
+  }).then(function (result) {
```
outoftime: Any reason not to use `async`/`await` in this code? The generated source gets compiled the same as any normal source file right?

src/util/compileProject.js
```diff
@@ -156,28 +156,48 @@ async function addPreviewSupportScript(doc) {
...
+
+  let code = javascript;
+  let sourceMap;
+
   if (breakLoops) {
     const {'default': loopBreaker} = await retryingFailedImports(
       () => import(
-        /* webpackChunkName: "mainAsync" */
-        'loop-breaker' // eslint-disable-line comma-dangle
+        /* webpackChunkName: "babel" */
```
outoftime: Maybe call it something like `jsCompiler` since it’s not strictly just Babel?

src/components/PreviewFrame.jsx
wylieconlon: Updated

wylieconlon:
@outoftime The error you found is totally real, and has been happening since #1674 was merged. Reverting that change fixes the issue in Browserstack for me. So I think the fix for that issue should be merged separately.

wylieconlon:
This now depends on code from https://github.com/popcodeorg/popcode/pull/1705

wylieconlon:
Tested against Chrome 38 and found no issues

wylieconlon:
This depends on #1705 for babel-ing all node_modules so I'll wait until that's in

webpack.config.js
```diff
@@ -14,17 +14,18 @@ const babel = require('@babel/core');
...
-const babelLoaderConfig = Object.assign({}, babelrc, {
-  cacheDirectory: true,
-  cacheIdentifier: JSON.stringify({
-    babel: babel.version,
-    'babel-loader': babelLoaderVersion,
-    debug: process.env.DEBUG,
-    env: process.env.NODE_ENV || 'development',
-  }),
-});
+const babelLoaderConfig = Object.assign(
```
outoftime: Seems like we can skip the `Object.assign` altogether at this point?

webpack.config.js
```diff
@@ -160,6 +161,10 @@ module.exports = (env = process.env.NODE_ENV || 'development') => {
           use: ['val-loader'],
           enforce: 'pre',
         },
+        {
+          test: /\.gen\.js$/u,
+          use: [{loader: 'babel-loader', options: babelLoaderConfig}],
+        },
```
outoftime: Can you explain why we need this, given the next rule should also match? Is it pre loader vs. normal loader?

src/services/babel-browser.gen.js
```diff
@@ -0,0 +1,77 @@
...
+  const keys = getPluginNamesFromPresetEnv();
+
+  const pluginImports = keys
+    .map((key, index) => `import plugin${index} from\n'@babel/plugin-${key}';`)
+    .join('\n');
+
+  const pluginNames = keys.map((key, index) => `plugin${index}`).join(',\n');
+
+  return {
+    code: `
```
outoftime: It seems mildly preferable to split this into two files—one of which, the `.gen.js`, just exports the list of plugin names; and the other which imports it and contains everything below. That way we avoid having substantial code in a string literal.

outoftime:
OK, here goes! Feel free to open up a followup PR if any of the above feedback speaks to you.

Issue #1340, opened 2018-01-11
Minimize preview button


outoftime:
In the summer 2017 redesign, we in haste removed the old minimize button from the layout and did not replace it. I think a down chevron on the right side of the title bar, next to the refresh button, would do just fine. All the actual code to handle showing/hiding the preview is still in place so we should just need a button to actually do it.

alexpelan:
I can do this one

outoftime:
Apparently someone at GitHub forgot a beginning-of-line anchor in their regexp…

PR #1631, opened 2019-01-16
Add minimize button to preview window

alexpelan:
I think the only controversial decision to note here is that I decided
that the preview frame should still be rendered, just not visible. It
seems useful to log to the console via your code without showing the
produced UI, for example.

Fixes: #1340

### DIFF
--- a/src/components/Preview.jsx
+++ b/src/components/Preview.jsx
@@ -1,18 +1,24 @@
 import {
+  faChevronDown,
+  faChevronUp,
   faExternalLinkAlt,
   faSyncAlt,
 } from '@fortawesome/free-solid-svg-icons';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import get from 'lodash-es/get';
+import partial from 'lodash-es/partial';
 import PropTypes from 'prop-types';
 import ImmutablePropTypes from 'react-immutable-proptypes';
 import React from 'react';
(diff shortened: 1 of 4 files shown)


alexpelan:
Open:
screen shot 2019-01-15 at 11 07 40 pm

And closed:
screen shot 2019-01-15 at 11 07 49 pm

alexpelan:
@outoftime since this is in service of maximizing editors, I think some inconsistent behavior isn't a problem. I propose hiding the preview also hides the console.

Regarding maximizing a whole window editor - that feels like a bigger lift in the codebase - if we collapse the preview below the javascript editor when it's collapsed, you should be able to close 2 of the 3 editors and get a pretty close to fullscreen experience. 

As for 2,  I don't think it's a big deal.

outoftime:
@alexpelan OK if you’re down then I am!

alexpelan:
New expanded:
screen shot 2019-02-19 at 8 13 19 pm
And collapsed:
screen shot 2019-02-19 at 8 13 11 pm

alexpelan:
@outoftime ready for another review!

alexpelan:
@outoftime that all makes sense.  I'm gonna close up this PR.  

My plan is to implement `CollapsedColumn` first in a separate PR before tackling the `Workspace` logic changes.

PR #1633, opened 2019-02-02
Upgrade redux-saga and redux-saga-test-plan to 1.0.1

wylieconlon:
Delay and debounce are now effects.

This upgrade is waiting on jfairbank/redux-saga-test-plan#242 to be merged

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -142,6 +142,13 @@
       "integrity": "sha512-7jGW8ppV0ant637pIqAcFfQDDH1orEPGJb8aXfUozuCU3QqX7rX4DA8iwrbPrR1hcH0FTTHz47yQnk+bl5xHQA==",
       "requires": {
         "regenerator-runtime": "^0.12.0"
+      },
+      "dependencies": {
+        "regenerator-runtime": {
+          "version": "0.12.1",
+          "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.12.1.tgz",
+          "integrity": "sha512-odxIc1/vDlo4iZcfXqRYFj0vpXFNoGdKMAUieAlFYO6m/nl5e9KR/beGf41z4a1FI+aQgtjhuaSlDxQ0hmkrHg=="
+        }
       }
     },
     "@babel/template": {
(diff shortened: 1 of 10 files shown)


wylieconlon:
@outoftime I've been trying to debug this test failure and I think I need some more eyes on it- could you take a look at the failure?

wylieconlon:
Okay, I found the source of the issue. One of the breaking changes in redux-saga 1.0.0 is a change to the effect structure:

```
refactor effect structure from {[IO]: true, [type]: payload } to { [IO]: true, type, payload } to get rid of dynamic type property. Could affect you if implement custom monitor for saga effects.
```

This affects a bunch of the tests, because there key `effect.SELECT` no longer exists, and no longer has the payload. So I'm fixing a bunch of failing tests as a result.

package-lock.json
```diff
@@ -17643,7 +17779,7 @@
         },
         "htmlparser2": {
           "version": "3.3.0",
-          "resolved": "https://registry.npmjs.org/htmlparser2/-/htmlparser2-3.3.0.tgz",
+          "resolved": "http://registry.npmjs.org/htmlparser2/-/htmlparser2-3.3.0.tgz",
```
outoftime: @wylieconlon seems like there is quite a bit of noise in the `package-lock` diff, both these switches from `https` to `http`, and dependencies where `"bundled": true` is replaced  by `resolved` and `integrity` properties. I’m not expert enough in NPM to know exactly why that is going on, and would definitely accept a confident “this is kind of inevitable because of how NPM works”, but just wondering if it is maybe avoidable.

wylieconlon:
@outoftime I found the discussion of the issue here, and the issue was in the NPM registry last summer. The proposed solution in these threads is upgrading to the latest NPM version and regenerating package-lock, so I think this will be fine once you merge your changes. https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285

outoftime:
@wylieconlon OK just merged in the fixed `package-lock`!

wylieconlon:
@outoftime I've just updated the package-lock to be as "default" as possible, it was generated by:

- Installing npm 6.7.0, with node 8.12
- Removing the node_modules directory
- Doing `npm install`
- Seeing `http` in some URLs, I ran: `sed -i -e 's/http:\/\//https:\/\//g' package-lock.json`

I would appreciate it if you would try recreating this installation, because I'm not confident that npm is behaving reliably here.

outoftime:
@wylieconlon heyo, just pushed an updated `package-lock.json` that now contains a sensible number of expected changes from master. Here’s the procedure I went through:

```bash
$ git checkout master
$ git pull
$ rm -r node_modules
$ npm install
$ git status # confirm package-lock.json is unchanged
$ git checkout saga-upgrade
$ git checkout master -- package-lock.json
$ npm install
$ git add package-lock.json
$ git diff --cached master -- package-lock.json # confirm that diff is sensible
```

The main different between what you did and what I did, I think, is that I made sure that the layout of `node_modules` was a direct result of installing the `package-lock.json` in master, before introducing a change to `package.json` and regenerating `package-lock.json`. Annoying that all these hoops need to be jumped through but seems at least logically consistent…

PR #1634, opened 2019-02-11
Regenerate package-lock.json from scratch with latest npm

outoftime:
Upgrades npm to latest 6.7.0 (and Node to latest 8.X while we’re      in there), then regenerate `package-lock.json` by temporarily      locking all first-order dependencies to exact versions, deleting      `package-lock.json` and the contents of node-modules, and then running an `npm install`.

Various people have had issues with `package-lock` containing a      ton of unrelated changes after installing new dependencies; the      previous `package-lock` if I’m not mistaken was derived from one      converted from the `yarn.lock` file we used with Yarn. Hopefully      a freshly-generated one should behave better.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
-FROM node:8.12.0
+FROM node:8.15.0
 
 RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - && \
     sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' && \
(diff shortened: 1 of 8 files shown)


wylieconlon:
Based on what I've just read, I'm not sure this extra script is necessary- https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285

outoftime:
I just put the script into the repo in case I need to do the same procedure in the future. Note that this PR does not actually lock the `package.json` dependency versions—but I did do so temporarily, before regenerating `package-lock.json`, so that the regenerated `package-lock` would point to the exact same first-order dependencies as it did previously.

PR #1638, opened 2019-02-16
Back to yarn

outoftime:
It’s proving a huge hassle to get `npm` to do reasonable things with `package-lock.json` and it still feels quite non-deterministic. Back to `yarn`.

My original complaint with `yarn` was that it generates installations that don’t pass `yarn check`; it turns out that `yarn check` is considered harmful and proposed for removal anyway. (`npm` does not have an equivalent)


wylieconlon:
Thanks!

Issue #1461, opened 2018-05-08
background-position-x reported as an unknown CSS property


outoftime:
`background-position-x` is an entirely legitimate CSS property, but Popcode says it isn’t. Probably need an upstream patch to PrettyCSS.

PR #1639, opened 2019-02-18
#1461 Updated PrettyCSS library and acceptance testing

jwang1919:
So before merging... I seem to have noticed that the acceptance testing is not actually testing anything.

I manually updated `acceptance.json` because `update-validation-acceptance` script wasn't working for me and intentionally tried to get it to fail by running `yarn test` by not including the library update.

@outoftime Any thoughts? Does this have to do with the most recent yarn change?

### DIFF
--- a/package.json
+++ b/package.json
@@ -188,7 +188,7 @@
     "@fortawesome/free-brands-svg-icons": "^5.3.1",
     "@fortawesome/free-solid-svg-icons": "^5.3.1",
     "@fortawesome/react-fontawesome": "^0.1.3",
-    "PrettyCSS": "^0.3.17",
+    "PrettyCSS": "^0.3.19",
     "array-to-sentence": "^2.0.0",
     "bowser": "^1.9.4",
     "brace": "^0.11.0",
--- a/test/unit/validations/css.js
+++ b/test/unit/validations/css.js
@@ -36,6 +36,14 @@ test('valid text-shadow declaration', validationTest(
   css,
 ));
 
+test('valid background-position declarations', validationTest(
+  `img {
+     background-position-x: 10px;
+     background-position-y: 15%;
+  }`,
+  css,
+));
+
 test('bogus flex value', validationTest(
   '.flex-item { flex: bogus; }',
   css,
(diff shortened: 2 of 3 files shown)


outoftime:
wait @jwang1919 how come you updated the acceptance tests in the first place? if the goal is just to test that your change allows `background-position-x` etc., typically you’d want to just add a test to `test/unit/validations/css.js`. The acceptance tests are pulled from a ScriptEd curriculum from a few years ago, the idea being that Popcode should never error on starter/solution code from the curriculum. (Obviously I should be keeping them up-to-date with each successive year’s curriculum…my bad)

outoftime:
(Also I just manually edited `acceptance.json` to introduce a validation error, and it does cause the test suite to fail…)

outoftime:
What’s happening with `update-validation-acceptance`? Working fine for me…

jwang1919:
Hmm... Maybe it has something to do with windows:
```
D:\Projects\ScriptEd\popcode>node script/update-validation-acceptance
fs.js:129
  throw new ERR_INVALID_CALLBACK();
  ^

TypeError [ERR_INVALID_CALLBACK]: Callback must be a function
    at maybeCallback (fs.js:129:9)
    at Object.writeFile (fs.js:1137:14)
    at Object.<anonymous> (D:\Projects\ScriptEd\popcode\script\update-validation-acceptance:22:4)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:279:19)
```

jwang1919:
So short update, I think I need to fix some of the scripts because `update-validation-acceptance` works on my Mac but definitely fails on my Windows PC (Windows 10, idk if that even matters).

PR #1641, opened 2019-02-20
Missing closing title tag generates validation error

Iula79:
fixing the missing title tag

### DIFF
--- a/src/validations/html/slowparse.js
+++ b/src/validations/html/slowparse.js
@@ -114,12 +114,8 @@ function findChildNode({childNodes}, nodeName) {
 }
 
 function emptyTitleElementDetector(_, root) {
-  /*function not running if opening title is there
-     head has 3 nodes if title is there with opening and closing tag, if it is only opening it thinks there is only one node with text*/
   const html = findChildNode(root, 'HTML');
   const head = html ? findChildNode(html, 'HEAD') : null;
-  // eslint-disable-next-line no-console
-  console.log('head', head);
   const title = head ? findChildNode(head, 'TITLE') : null;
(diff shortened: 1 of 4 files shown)


src/validations/html/slowparse.js
```diff
@@ -66,6 +71,7 @@ const errorMap = {
   UNCLOSED_TAG: error => ({
     reason: 'unclosed-tag',
     payload: {tag: error.openTag.name},
+
```
jwang1919: fyi @outoftime will freak out about this

Iula79: removed

src/validations/mergeValidations.js
```diff
@@ -4,9 +4,12 @@ import flatten from 'lodash-es/flatten';
 import flatMap from 'lodash-es/flatMap';
 import sortBy from 'lodash-es/sortBy';
 import omit from 'lodash-es/omit';
+import uniqWith from 'lodash-es/uniqWith';
+
 
 function filterErrors(errors) {
-  const groupedErrors = groupBy(errors, 'reason');
+  const dedupedErrors = uniqWith(errors, e => e.reason && e.row);
```
outoftime: I don’t think this is doing quite what you intend—[`uniqWith`](https://lodash.com/docs/4.17.11#uniqWith) passes two arguments to the callback, which are two values to be compared. So, your function should take two parameters, and then return `true` if the `reason` and `row` properties have the same value across those two parameters.

Iula79: hi thanks for this I am testing now

Issue #1442, opened 2018-05-02
Stop using bower


outoftime:
We currently use Bower to install the libraries that are available to project code, e.g. jQuery, lodash, etc. This is a good fit since Bower by default ships packages in the old “put a global variable on `window`” form, which is exactly what Popcode projects expect.

However, Bower itself discourages people from using it, so we should probably migrate away from it. Ideally, we could just install the same packages via `npm` and load the pre-built standalone script into the preview frame. But there’s no guarantee that’s possible—lodash, for instance, does not appear to ship with a pre-built standalone script in its npm package.

A couple of other options (all kind of complicated and not terribly appealing) come to mind:

* Create a webpack entrypoint for each library, then load the desired ones directly from the preview frame using `<script>` tags. This has the advantage of maintaining the expected behavior associated with loading scripts the old-fashioned way
* Bundle all the libraries as part of the preview support script, but have the preview expose only the ones that are enabled as properties on `window`. This might get tricky with evaluation order, but has the advantage of not creating a proliferation of webpack entrypoints, and avoids the possibly vexing task of capturing the right reference to a bundle entrypoint to use in the `<script>` tag

alexpelan:
not fixed completely yet, reopening

PR #1646, opened 2019-02-21
Source jQuery from npm, not bower

alexpelan:
This one was straightforward - jQuery's npm package ships with a dist
version that can be easily dropped in where the bower one was before.

The first of many steps for #1442

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -491,6 +491,10 @@
     "no-regex-spaces": [
       "error"
     ],
+    "no-restricted-imports": [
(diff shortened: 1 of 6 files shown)


outoftime:
LGTM, but can we also add a [`no-restricted-imports`](https://eslint.org/docs/rules/no-restricted-imports) ban on importing jQuery into the application bundle? I would hate for anyone to get the wrong impression…

alexpelan:
@outoftime done!

PR #1667, opened 2019-03-13
Upgrade inline-style-prefixer to latest

outoftime:
This now requires code generation to support older browsers.  Fortunately, it's pretty easy to do the code gen on the fly using the `val-loader`, although not with `inline-style-prefixer`'s `generator` module as written. I have submitted a [pull request](https://github.com/rofrischmann/inline-style-prefixer/pull/175) to export the method that generates code and returns a string, which this PR assumes is available.


outoftime:
@wylieconlon ran into another situation in which we need to generate code to compile into the bundle, and it looks like [`val-loader`](https://github.com/webpack-contrib/val-loader) gives us the appealing option of generating the code at compile time, rather than checking generated code into source control. Note that the module processed by `val-loader` can return an async function, which I'm assuming would be necessary in the case of generating `preset-env` imports.

src/services/inlineStylePrefixer/prefixData.gen.js
```diff
@@ -0,0 +1,21 @@
...
+} = require('inline-style-prefixer/lib/generator');
+
+const browserList = require('../../../config/browsers.json');
+
+module.exports = () => {
+  const browserListForInlineStylePrefixer = {};
+
+  for (const browser in browserList) {
+    if (browser !== 'chromium') {
+      browserListForInlineStylePrefixer[browser] = Number(browserList[browser]);
```
wylieconlon: Converting versions to a number is throwing a warning for me, saying:

```
Warning, the following targets are using a decimal version:

  safari: 7.1

We recommend using a string for minor/patch versions to avoid numbers like 6.10
getting parsed as 6.1, which can lead to unexpected behavior.```

Issue #1434, opened 2018-04-25
Replace ACE with CodeMirror


outoftime:
[CodeMirror](https://codemirror.net) is a promising-looking alternative to ACE that is used in a [lot of projects](https://codemirror.net/doc/realworld.html), including the developer tools for Chrome, Firefox, and Safari.

Let’s investigate and see whether CodeMirror would alleviate any of the major ACE pain points:

1. It’s a huge portion of the overall bundle footprint. I’m guessing we load a fair amount of unused code.
2. It uses a proprietary module system, and is only compatible with CommonJS/ES6 modules at all via the third-party `brace` port. This is probably a major contributor to the first issue.
3. Preventing the inclusion of every available language plugin in the bundle requires webpack configuration level hacks.

For full release we need to make sure all these features from the status quo are maintained:

- [ ] Display errors as annotations in the gutter
- [x] Ability to customize editor behavior, especially to disable auto-closing of brackets
- [x] Built in syntax highlighting with fully customizable color scheme
- [ ] Start/end bracket highlighting
- [x] Undo
- [x] Find-in-code
- [ ] Support for auto-formatting (capture the keyboard shortcut, etc.)
- [ ] Support regular vs. large text mode
- [ ] Focus on error location when error is clicked in output
- [ ] Use it for the console input as well
- [ ] Highlight current line

wylieconlon:
I'm planning to investigate this more thoroughly and recommend a specific editor option, such as CodeMirror or [Monaco](https://microsoft.github.io/monaco-editor/) based on the work I'm doing to integrate VSCode-style language intelligence

wylieconlon:
A summary of my findings is that both CodeMirror and Monaco support each of the requirements you've described above. CodeMirror has fewer of those features enabled in its default setting, but the addons are easy to use and can be styled to a point. Monaco has a bunch of additional features included out of the box, including use of service workers, but I tested with that disabled and it just acted like a very basic text editor. The biggest difference is bundle size- CodeMirror is a significantly smaller codebase. Monaco is probably around 5-7Mb of code depending on how many extensions are included. So the basic idea is that Monaco is a very full-featured editor, which has many customization options out of the box. CodeMirror requires more tweaking and state management to get working, but all the features of the larger library are achievable.

[...]

PR #1670, opened 2019-03-14
Experimental CodeMirror editor

outoftime:
Creates an experimental `<CodeMirrorEditor>` component, which accepts the same (actually currently a subset of) the props as `<Editor>` but uses CodeMirror rather than ACE as the underlying editor implementation.

This is an initial spike and is not intended to be full-featured or release-ready. The specific goals of this pull request are:

* Render an editor of the correct size, with a fairly sensible/familiar look and feel
* Keep editor state in sync with underlying current project state

We do get several other things for free/by default with CodeMirror. The full list of features that are required for parity with the ACE implementation can be found in the checklist in #1434.

In this pull request (and until that checklist is complete), the CodeMirror editor is only used in experimental mode (when the `experimental` query param is present in the URL). In order to avoid bloating the bundle unnecessarily, the `<CodeMirrorEditor>` component is loaded asynchronously and lazily, in its own chunk, only in experimental mode. This means that the interface takes slightly longer to fully load in experimental mode.


outoftime:
@wylieconlon relevant to your interests? mind taking a look?

wylieconlon:
Yes, I'll take a look as soon as I can. Thanks for putting this out!

src/components/AceEditor.jsx
```diff
@@ -0,0 +1,176 @@
...
+import 'brace/mode/html';
+import 'brace/mode/css';
+import 'brace/mode/javascript';
+import 'brace/theme/monokai';
+
+const RESIZE_THROTTLE = 250;
+const NORMAL_FONTSIZE = 14;
+const LARGE_FONTSIZE = 20;
+
+class AceEditor extends React.Component {
```
wylieconlon: I notice that you're setting `this._editor` to a DOM element, but don't initialize it in the constructor. I think it might help the clarity of the code to set it to null first, and then to a DOM element. What do you think?

src/components/CodeMirrorEditor.jsx
```diff
@@ -0,0 +1,90 @@
...
+};
+
+export default function CodeMirrorEditor({
+  language,
+  projectKey,
+  source,
+  onInput,
+}) {
+  const containerRef = useRef();
+  const editorRef = useRef();
```
wylieconlon: [The docs](https://reactjs.org/docs/hooks-reference.html#useref) show that you're supposed to provide an initial value- I would suggest doing `useRef(null)` to make this clear

src/components/CodeMirrorEditor.jsx
```diff
@@ -0,0 +1,90 @@
...
+  useEffect(() => {
+    const {current: container} = containerRef;
+
+    const editor = editorRef.current = CodeMirror(
+      container,
+      {
+        lineNumbers: true,
+        lineWrapping: true,
+      },
+    );
```
wylieconlon: I know there isn't explicitly a `CodeMirror.Editor.remove()` function, but is it worth removing the DOM element on cleanup of this effect?

src/components/CodeMirrorEditor.jsx
```diff
@@ -0,0 +1,90 @@
+import CodeMirror from 'codemirror';
+import isNil from 'lodash-es/isNil';
```
wylieconlon: Not actually used

src/components/CodeMirrorEditor.jsx
```diff
@@ -0,0 +1,90 @@
...
+  source,
+  onInput,
+}) {
+  const containerRef = useRef();
+  const editorRef = useRef();
+  const docsCacheRef = useRef(new LRU(3));
+  const mode = CODEMIRROR_MODES_MAP[language];
+
+  useEffect(() => {
+    const {current: container} = containerRef;
```
wylieconlon: I would normally avoid stylistic comments, but this one is bothering me. Why destructure when the key isn't used? I prefer `const container = containerRef.current;` and think it would simplify all the effects code

src/components/CodeMirrorEditor.jsx
```diff
@@ -0,0 +1,90 @@
...
+
+    const editor = editorRef.current = CodeMirror(
+      container,
+      {
+        lineNumbers: true,
+        lineWrapping: true,
```
wylieconlon: You haven't provided a theme name or mode to the initialize call- I think you can, since the props are available here

src/components/CodeMirrorEditor.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const containerRef = useRef();
+  const editorRef = useRef();
+  const docsCacheRef = useRef(new LRU(3));
+  const mode = CODEMIRROR_MODES_MAP[language];
+
+  useEffect(() => {
+    const {current: container} = containerRef;
+
+    const editor = editorRef.current = CodeMirror(
+      container,
```
wylieconlon: In this case, I don't think you need to unwrap the reference at all. `containerRef.current` would work fine here

src/components/CodeMirrorEditor.jsx
```diff
@@ -0,0 +1,90 @@
...
+    const docKey = `${mode}:${projectKey}`;
+    if (!docsCache.has(docKey)) {
+      docsCache.set(docKey, new CodeMirror.Doc('', mode));
+    }
+    const doc = docsCache.get(docKey);
+
+    if (doc !== editor.getDoc()) {
+      editor.swapDoc(new CodeMirror.Doc('', mode));
+    }
+  }, [language, mode, projectKey]);
```
wylieconlon: This is a nice one!

outoftime:
@wylieconlon thanks for the thoughtful and detailed code review! i actually had a long flight yesterday and I think I got the CM editor close to parity with functionality with ACE, but didn't get a chance to push it until this morning. But I will definitely go through your review, respond to stuff, and make changes as appropriate : )

gulpfile.js
```diff
@@ -28,6 +28,7 @@ const srcDir = 'src';
 const distDir = 'dist';
 const stylesheetsDir = path.join(srcDir, 'css');
 const highlightStylesheetsDir = 'node_modules/highlight.js/styles';
+const codemirrorStylesheet = 'node_modules/codemirror/lib/codemirror.css';
```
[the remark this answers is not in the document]
outoftime: I agree that the default scheme isn't amazing, but I actually went through all the themes that ship with the package and didn't find any I thought were overall better than the default (I also liked the IDE-derived themes but there were things I didn't like that I don't remember now; I think they were mostly in the way they themed HTML). This [repository of themes](http://farhadg.github.io/code-mirror-themes/) looks promising though...

Anyway, seems like something we can continue to ponder between initial experimental-mode release and actually making this the official editor.

.eslintrc
```diff
@@ -922,6 +926,7 @@
   "plugins": [
     "import",
     "react",
+    "react-hooks",
```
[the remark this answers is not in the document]
outoftime: So my basic thinking is:

* Hooks are good and class components are bad (more on that below)
* We should therefore not introduce any new class components, and when we need the features that require either hooks or class components in a new component, we should use hooks
* And soon we should probably transition fully to hooks even though the React folks say not to go all crazy rewriting components for hooks just yet
* So, with that in mind, this is the first time since hooks became a thing that we have a new component that needs to use hooks. So, this would be the exact time to "start supporting hooks" (I don't see much argument for, say, introducing the eslint plugin when there are no actual hooks to lint)

gulpfile.js
```diff
@@ -68,6 +69,7 @@ gulp.task('css', () => {
     src([
       path.join(bowerComponents, 'normalize-css/normalize.css'),
       path.join(highlightStylesheetsDir, 'github.css'),
+      codemirrorStylesheet,
       path.join(stylesheetsDir, '**/*.css'),
     ]).
```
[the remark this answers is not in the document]
outoftime: Are you cool with what is there now? I'm not exactly sure what the important difference is between what I had before and what you wrote in the above comment...

package.json
```diff
@@ -218,6 +219,7 @@
     "loadjs": "^3.5.4",
     "lodash-es": "^4.17.11",
     "loop-breaker": "^0.1.0",
+    "lru-cache": "^5.1.1",
```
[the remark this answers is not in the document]
outoftime: Used to cache `CodeMirror.Doc` instances. Basically I want to hold on to the last few, so that the undo buffer is retained if you're switching back and forth between a couple projects, but don't want to build up an indefinitely growing collection of them.

src/components/AceEditor.jsx
```diff
@@ -0,0 +1,176 @@
...
+import noop from 'lodash-es/noop';
+
+import {EditorLocation} from '../records';
+import {createAceEditor, createAceSessionWithoutWorker} from '../util/ace';
+
+import 'brace/ext/searchbox';
+import 'brace/mode/html';
+import 'brace/mode/css';
+import 'brace/mode/javascript';
+import 'brace/theme/monokai';
```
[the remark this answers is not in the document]
outoftime: I don't actually know, and I'm not sure if this line of code even does anything. In CodeMirror, `monokai` is a dark theme...

src/components/AceEditor.jsx
outoftime: Eh this component only shows up in the diff because I renamed it; I don't think it's worth agonizing over the code therein since the goal is to delete it soon anyway.

src/components/CodeMirrorEditor.jsx
outoftime: The docs don't make a particularly strong claim that the initial value is required, I don't think, and [this example from the Hooks FAQ](https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables) shows `useRef()` called without an argument, so I think it's legit. And semantically it makes sense, just like a `let` that declares a variable without assigning it, thus giving it an `undefined` initial value.

src/components/CodeMirrorEditor.jsx
outoftime: It's a good thought, but I don't think so. This is a zero-dependency effect, meaning it will only be run once, after the component is initially rendered. Thus, the only time this would need to be cleaned up is when the component itself is torn down. And cleanup for a CodeMirror editor merely entails removing its container element from the DOM (see [the docs for `cm.getWrapperElement()`](https://codemirror.net/doc/manual.html#api_misc)).

src/components/CodeMirrorEditor.jsx
outoftime: Quite right! I ended up rewriting most of these already, but will seek out any lingerers. I also [added an entry to the contributor guide](https://github.com/popcodeorg/popcode/wiki/Contributor-Guide#destructuring-should-reduce-repetition-without-reducing-clarity) proposing some guidelines for when destructuring is useful.

src/components/CodeMirrorEditor.jsx
outoftime: As for a theme name, as I mentioned, I have yet to land on a theme that is obviously preferable to the default theme.

For the mode, it would feel good to put it here, but that would make the `language` prop an effect dependency, which we definitely don't want here! Though practically we would never expect `language` to change, if it did in fact change, we should not tear down and re-create the editor instance; we should just swap in a new `Doc` instance, with the appropriate mode. And that's exactly what the following effect hook does.

src/components/CodeMirrorEditor.jsx
outoftime: Something about refs makes me _always_ want to unwrap them, basically so I don't have to look at the ref itself any more than necessary. Also, in some cases it actually does matter (e.g. in effect cleanup functions) so I think it's simpler to just have a global rule of "always unwrap ref values into their own variable" and then you don't have to worry about it.

src/components/CodeMirrorEditor.jsx
outoftime: Thanks!

outoftime:
> Overall I like the direction of this, my biggest concern is whether we can easily test the various effects without relying too heavily on Ace or CodeMirror APIs. My secondary concern is that a class component might have easier-to-follow logic

So after reading [Dan Abramov's Big Blog Post](https://overreacted.io/a-complete-guide-to-useeffect/) I am pretty strongly convinced that class components are something approaching an antipattern (specifically the notion of `this.props` does not make sense). Have you had a chance to read it? If/when you _have_ read it if you are still unconvinced, I am happy to make my own case using his mental model as a framework for the argument : )

Anyway, I believe this concludes all comment responses I owe you, looking forward to hearing your thoughts!

gulpfile.js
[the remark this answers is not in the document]
wylieconlon: The default scheme is specifically less contrasty than these other options, which made it noticeable when switching back and forth. Totally fine with me to update later

.eslintrc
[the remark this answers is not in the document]
wylieconlon: Since writing this I'm now developing almost entirely in hooks, I think I'll write up some thoughts on that on a higher-up comment.

gulpfile.js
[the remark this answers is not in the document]
wylieconlon: Based on our discussion of adding stylesheets later this is fine

package.json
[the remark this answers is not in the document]
wylieconlon: Yes, makes sense

src/components/CodeMirrorEditor.jsx
wylieconlon: That wasn't obvious to me on first review, I'd appreciate a comment on why the mode is missing

src/css/application.css
```diff
@@ -540,6 +553,25 @@ body {
...
+/* postcss-bem-linter: ignore */
+.CodeMirror-matchingbracket {
+  color: inherit !important;
+  border: 1px solid var(--color-light-gray);
+  border-radius: 1px;
+}
+
+/* postcss-bem-linter: ignore */
+.CodeMirror-activeline-background {
+  background: var(--color-low-contrast-gray) !important;
```
wylieconlon: I hadn't realized we were using css variables, cool!

src/components/__tests__/CodeMirrorEditor.test.jsx
```diff
@@ -0,0 +1,134 @@
+import CodeMirror from 'codemirror';
+import findLast from 'lodash-es/findLast';
+import last from 'lodash-es/last';
+import React from 'react';
+import ShallowRenderer from 'react-test-renderer/shallow';
+import TestRenderer, {act} from 'react-test-renderer';
+
+// eslint-disable-next-line import/no-extraneous-dependencies
```
outoftime: There’s a [fix](https://github.com/benmosher/eslint-plugin-import/pull/794) in the master branch of `eslint-plugin-import` that has yet to be released; once that lands this disable will no longer be necessary (and should itself trigger a lint error for no unnecessary disables)

src/components/AceEditor.jsx
```diff
@@ -0,0 +1,176 @@
...
+    this.props.onRequestedLineFocused();
+  }
+
+  _resizeEditor() {
+    this._editor.resize();
+  }
+
+  _scrollToLine(lineNumber) {
+    const shouldCenterVertically = true;
+    const shouldAnimate = true;
```
jwang1919: If these two consts will always be true, why not just directly set both to `true` as arguments? If it's supposed to be configurable in the future, these values should be located in a properties/config file.

src/css/application.css
jwang1919: Can you avoid using `!important`? Also applies to line 565,

src/components/AceEditor.jsx
outoftime: Not sure/good point but this is basically just moving the old `Editor` component into a new place, and we’re deprecating it anyway, so I don’t think there’s any need to worry about this question

src/css/application.css
outoftime: I don’t think so—generally using `!important` is bad but in this case we are overriding a style in a vendor stylesheet so I don’t think there is any alternative.

outoftime:
N.B. I did a quick apples-to-apples on this and it looks like the bundle size will be about 10% smaller with CodeMirror fully replacing Ace.

PR #1671, opened 2019-03-15
Improve eslint config

outoftime:
Moves all deeply-nested `.eslintrc` override files into project root `.eslintrc`, as `"override"` entries. Also sorts the top-level keys in `.eslintrc`.

I like using `"overrides"` rather than putting override files in directories for a couple of reasons:

* In some cases overrides don't map neatly to the directory structure
* It keeps everything in one place

This also introduces overrides for config files and scripts that are run in Node rather than compiled by babel / run in the browser, which allows us to remove the eslint configuration comments from those files.

Finally, rules that have no configuration are just given as a severity string, rather than a one-element array, resulting in a more compact configuration file.


outoftime:
FYI @jwang1919 this is extracted/inspired by your jest/logic PR, my plan is to hold this one until yours is ready to go, and then:

* merge this to master
* rebase yours on top of it
* merge yours to master

Sound reasonable?

jwang1919:
@outoftime My PR is ready, can you merge this one?

outoftime:
@jwang1919 done!

PR #1672, opened 2019-03-25
Completed redux-logic transition for linkGithubIdentity code

jwang1919:
Followed same code conventions of when I did unlinkGithubIdentity. Replicated the same tests but in Jest and added a few more assertions.


src/logic/linkGithubIdentity.js
```diff
@@ -0,0 +1,37 @@
...
+
+export default createLogic({
+  type: 'LINK_GITHUB_IDENTITY',
+  async process() {
+    try {
+      const {user: userData, credential} = await linkGithub();
+      await saveCredentialForCurrentUser(credential);
+      return identityLinked(userData, credential);
+    } catch (e) {
+      switch (e.code) {
```
outoftime: I know the saga did the same thing (my fault!), but I think it’s a little silly to use a `switch` statement with only one case and a default. Let’s just use an `if`/`else`?

src/logic/__tests__/linkGithubIdentityTest.js
```diff
@@ -0,0 +1,72 @@
...
+    const githubProfile = {login: 'popcoder'};
+
+    linkGithub.mockRejectedValue(error);
+    getProfileForAuthenticatedUser.mockResolvedValue(githubProfile);
+
+    const {type, payload: {credential: {
+      providerId, accessToken,
+    }}} = await linkGithubIdentity.process();
+    expect(linkGithub).toHaveBeenCalledWith();
+    expect(getProfileForAuthenticatedUser).toHaveBeenLastCalledWith(
```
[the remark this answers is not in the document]
jwang1919: That was a mistake, fixing it

src/logic/__tests__/linkGithubIdentityTest.js
```diff
@@ -0,0 +1,72 @@
...
+    expect(providerId).toBe('github.com');
+    expect(accessToken).toBe('abc123');
+  });
+
+  test('other error', async() => {
+    const errorMessage = 'authentication problem!';
+    const otherError = new Error(errorMessage);
+
+    linkGithub.mockRejectedValue(otherError);
+    bugsnagClient.notify.mockResolvedValue(null);
```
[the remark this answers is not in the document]
jwang1919: Yes. This was a something I added earlier because I couldn't get the test to pass before. Thought I needed to add an argument but problem came from something else.

src/logic/__tests__/linkGithubIdentityTest.js
```diff
@@ -0,0 +1,72 @@
...
+import {getProfileForAuthenticatedUser} from '../../clients/github';
+import {bugsnagClient} from '../../util/bugsnag';
+
+jest.mock('../../clients/firebase.js');
+jest.mock('../../clients/github.js');
+jest.mock('../../util/bugsnag.js');
+
+describe('linkGithubIdentity', () => {
+  test('success', async() => {
+    const mockCredential = {providerId: 'github.com'};
```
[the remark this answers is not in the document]
jwang1919: Hmm... okay, that sounds okay with me. Never worked with Rosie before so will take a bit more time to complete this.

src/logic/__tests__/linkGithubIdentityTest.js
```diff
@@ -0,0 +1,72 @@
...
+
+    linkGithub.mockResolvedValue({
+      user: {},
+      credential: mockCredential,
+    });
+
+    const {type, payload: {credential: {
```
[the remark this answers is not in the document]
jwang1919: Good idea with adding it to the contributor guide. I kind of messed around with it until eslint stopped complaining.

src/logic/__tests__/linkGithubIdentityTest.js
```diff
@@ -0,0 +1,72 @@
...
+  test('other error', async() => {
+    const errorMessage = 'authentication problem!';
+    const otherError = new Error(errorMessage);
+
+    linkGithub.mockRejectedValue(otherError);
+    bugsnagClient.notify.mockResolvedValue(null);
+
+    const {type, error, payload: {
+      message,
+    }} = await linkGithubIdentity.process();
```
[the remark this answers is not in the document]
jwang1919: Done

src/logic/__tests__/linkGithubIdentityTest.js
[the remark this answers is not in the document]
outoftime: I actually ended up running into a situation where I needed factories too… stay tuned, currently wrangling eslint 😭

jwang1919:
@outoftime Can this be merged? As discussed in chat the Rosie implementation will be coming later rather than sooner.

outoftime:
@jwang1919 it's not ideal to have this held up, but I think it's important not to introduce patterns into the new year suite that we don't want to see there in the long run—in this case, ad-hoc definition of test data. Once #1670 is merged, I think it should be pretty quick for you to add the factories you need

outoftime:
@jwang1919 the CodeMirror branch is going to take a bit more work before it’s ready for primetime, so I cherry-picked over the bit that sets you up to write factories and pushed it to your branch here. I recommend checking out the [documentation for rosie](https://github.com/rosiejs/rosie#usage) but you can also take a look at the [factory I set up in the CodeMirror branch](https://github.com/popcodeorg/popcode/blob/14a7cc6e3485401b9afb6c243f54abd045566832/__factories__/packages/codemirror.js) and the [reasonably non-annoying way to import factories into jest tests](https://github.com/popcodeorg/popcode/blob/14a7cc6e3485401b9afb6c243f54abd045566832/src/components/__tests__/CodeMirrorEditor.test.jsx) employed therein.

jwang1919:
@outoftime okay. Used Rosie with `linkGithubIdentityTest.js`. My approach to getting Rosie to create an Error object was a bit tricky, so let me know if I'm heading in the right direction with this.
Also what would a realistic Firebase object look like so I can Rosie it?

__factories__/clients/firebase.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const credential = new Factory().attrs({
+  providerId: 'github.com',
+  accessToken: 'abc123',
+});
+
+export const githubProfile = new Factory().attrs({
+  login: 'popcoder',
+});
```
outoftime: The top-level principle for organizing factories should be where the data they’re generating would come from in the real application—in this case this profile object would come from the GitHub client, so this should live in `__factories__/clients/github.js`.

Also, as I mentioned before, we’ll want to generate factory data that is a reasonably complete representation of what the code we’re testing will see. I don’t necessarily think we need to capture _every_ property of large objects like a GitHub profile, but we should try to capture a subset that generously covers all the properties Popcode might care about (this is a bit subjective, but I trust your judgement). As for where to find that information, [API docs](https://developer.github.com/v3/users/#response-with-public-and-private-profile-information) are probably your best bet; throwing a `debugger` or a `console.log` in the application code and then capturing a real example is another fine approach.

__factories__/clients/firebase.js
```diff
@@ -0,0 +1,31 @@
...
+
+export const firebaseError = Factory.define(
+  'firebaseError',
+  FirebaseError,
```
outoftime: This seems like the right approach to me! :+1:

__factories__/clients/firebase.js
```diff
@@ -0,0 +1,31 @@
...
+).attrs({
+  name: 'some other error',
+});
+
+export const credentialInUseError = new Factory().extend(firebaseError).attrs({
+  name: 'auth/credential-already-in-use',
+  credential: credential.build(),
```
outoftime: I think when an attribute is a full object (rather than a scalar value) we want to build it at instantiation time; otherwise all `credentialInUseError`s will hold a reference to the _same_ credential object, meaning mutating it in one place would mutate it everywhere. Of course running into that sort of problem is not practically very likely, but still better to establish a practice of avoiding it altogether imo. So here I think you would just want to do:

```
attr('credential', () => credential.build())
```

src/logic/__tests__/linkGithubIdentityTest.js
```diff
@@ -1,5 +1,12 @@
-import linkGithubIdentity from '../linkGithubIdentity';
+import {
+  credential as createCredential,
+  credentialInUseError,
+  firebaseError,
+  githubProfile as createGithubProfile,
```
outoftime: From a naming standpoint I prefer `githubProfileFactory` etc. rather than `createGithubProfile`, since the things we are importing aren’t themselves functions. This is the convention I am using [in the CodeMirror pull](https://github.com/popcodeorg/popcode/pull/1670/files#diff-48da82123a3b32cc381d1f03147a30a6R9)

src/logic/__tests__/linkGithubIdentityTest.js
```diff
@@ -0,0 +1,72 @@
...
+jest.mock('../../clients/firebase.js');
+jest.mock('../../clients/github.js');
+jest.mock('../../util/bugsnag.js');
+
+describe('linkGithubIdentity', () => {
+  test('success', async() => {
+    const mockCredential = {providerId: 'github.com'};
+
+    linkGithub.mockResolvedValue({
+      user: {},
```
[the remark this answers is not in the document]
outoftime: To answer your question the [Firebase API docs](https://firebase.google.com/docs/reference/js/firebase.User.html#properties_1) are pretty good for figuring out the shape of a Firebase User; and again throwing a `debugger` or `console.log()` in the application code is a good way to see what the object actually looks like in a concrete instance.

src/logic/__tests__/linkGithubIdentityTest.js
```diff
@@ -59,13 +63,12 @@ describe('linkGithubIdentity', () => {
       error.credential.accessToken,
     );
     expect(type).toBe('ACCOUNT_MIGRATION_NEEDED');
-    expect(providerId).toBe('github.com');
-    expect(accessToken).toBe('abc123');
+    expect(providerId).toBe(error.credential.providerId);
+    expect(accessToken).toBe(error.credential.accessToken);
```
outoftime: :+1: one of the best things about using factories is that we stop duplicating value literals between setup and expectations!

__factories__/clients/firebase.js
outoftime: Looks like this is still outstanding?

jwang1919: Huh, I thought I changed this already. I'll check it out again.

jwang1919:
So the lint errors seem to only be showing up in Travis but not when I run `yarn test`.
Ex.
On this line, I added a comment to hide the jslint errors and warnings. `yarn test` passes for me, but when it Travis thinks it is no bueno.
https://github.com/jwang1919/popcode/blob/jest/src/logic/__tests__/linkGithubIdentityTest.js#L7

jwang1919:
@outoftime anyway yeah I need help, Major Payne is kicking my ass.

outoftime:
@jwang1919 got u

src/logic/__tests__/linkGithubIdentityTest.js
```diff
@@ -0,0 +1,88 @@
...
+    } = await linkGithubIdentity.process();
+
+    expect(linkGithub).toHaveBeenCalledWith();
+    expect(saveCredentialForCurrentUser).toHaveBeenCalledWith(mockCredential);
+    expect(type).toBe('IDENTITY_LINKED');
+    expect(providerId).toBe(mockCredential.providerId);
+    expect(user).toEqual(mockUser);
+  });
+
+  test('credential already in use in experimental mode', async() => {
```
outoftime: You can drop the “in experimental mode” from this, that’s an outdated description in the old test.

outoftime:
@jwang1919 looks like you might have forgotten to commit the new factory?

jwang1919:
Oops! My bad.

PR #1673, opened 2019-04-03
Introduce `CollapsedComponent` for minified component

alexpelan:
This is the some groundwork for allowing the preview to be minimized.
`CollapsedComponent` unifies the collapsed UI for the currently
collapsible components (the 3 editors and the console).

There shouldn't be user-facing changes here.

Starts to fix: #1340

### DIFF
--- a/src/components/CollapsedComponent.jsx
+++ b/src/components/CollapsedComponent.jsx
@@ -7,26 +7,25 @@ import classnames from 'classnames';
 
 export default function CollapsedComponent({
   component,
-  isRightJustified = true,
+  isRightJustified,
   onComponentUnhide,
   projectKey,
   text,
 }) {
   return (
     <div
-      className="editors__collapsed-editor"
+      className="collapsed-component"
       onClick={partial(
         onComponentUnhide,
         projectKey,
-        `editor.${component}`,
+        component,
       )}
     >
       <div
         className={classnames(
           'label',
-          'editors__label',
-          'editors__label_collapsed',
-          {editors__label_left: !isRightJustified},
+          'collapsed-component__label',
+          {'collapsed-component__label_left': !isRightJustified},
         )}
       >
         {text}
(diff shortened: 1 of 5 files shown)


src/components/CollapsedComponent.jsx
```diff
@@ -0,0 +1,46 @@
...
+export default function CollapsedComponent({
+  component,
+  isRightJustified = true,
+  onComponentUnhide,
+  projectKey,
+  text,
+}) {
+  return (
+    <div
+      className="editors__collapsed-editor"
```
outoftime: Seems like this would be the time to make this a top-level BEM block?

src/components/CollapsedComponent.jsx
```diff
@@ -0,0 +1,46 @@
...
+        projectKey,
+        `editor.${component}`,
+      )}
+    >
+      <div
+        className={classnames(
+          'label',
+          'editors__label',
+          'editors__label_collapsed',
+          {editors__label_left: !isRightJustified},
```
outoftime: Ibid. BEM

src/components/Console.jsx
```diff
@@ -33,6 +34,19 @@ export default function Console({
   onToggleVisible,
   requestedFocusedLine,
 }) {
+  if (!isOpen) {
+    return (
+      <CollapsedComponent
+        component="console"
+        isRightJustified={false}
+        key="console"
```
outoftime: Do we need this?

src/components/CollapsedComponent.jsx
```diff
@@ -0,0 +1,46 @@
...
+  projectKey,
+  text,
+}) {
+  return (
+    <div
+      className="editors__collapsed-editor"
+      onClick={partial(
+        onComponentUnhide,
+        projectKey,
+        `editor.${component}`,
```
outoftime: I think the `editor.` bit should be part of what the outer component is passing in as the `component` prop?

alexpelan:
Addressed all your comments

PR #1674, opened 2019-04-03
Run all node_modules through babel

wylieconlon:
This does not significantly increase the bundling time or bundling size, but should catch any cases where node_modules are using newer standards than our browsers support.


babel.config.js
```diff
@@ -29,10 +29,5 @@ module.exports = (api) => {
     ],
     plugins,
     compact: false,
-    overrides: [
-      {
-        include: './node_modules/parse5-sax-parser/lib/index.js',
-      },
-    ],
```
outoftime: Wait what was this even doing?

webpack.config.js
```diff
@@ -251,19 +251,8 @@ module.exports = (env = process.env.NODE_ENV || 'development') => {
...
-            matchModule('chalk'),
-            matchModule('lodash-es'),
-            matchModule('parse5'),
-            matchModule('parse5-sax-parser'),
-            matchModule('postcss-html'),
-            matchModule('recast'),
-            matchModule('redux'),
-            matchModule('stylelint'),
-          ],
+          test: /\.[jt]sx?$/u,
```
outoftime: Hmm, if we actually did encounter a TypeScript file (which is not something we would expect in a node module) we wouldn’t be able to compile it, would we? Since we don’t have the babel TS compiler in the build chain?

So it seems misleading to include `.ts` in the extensions this rule catches…

wylieconlon: You are correct, this implies support for typescript in a way that doesn't make sense.

babel.config.js
wylieconlon: I couldn't tell!

Issue #1301, opened 2017-12-19
Command S


pwjablonski:
I noticed and heard from several students returning to popcode after using a different editor that it is frustrating when they command S. Currently it gives the browser defualt option. Can we diasable this or maybe include a visual indicator to note that their work is saved.

outoftime:
@pwjablonski I’m down, although given the rarity of this complaint I don’t think this is terribly urgent / probably worth doing the “right thing”, which I would say is something like:

* If you do Command S when logged out, prompt the user to log in
* If you do Command S when logged in, show a message confirming that work is saved (closely tied to #1128)

PR #1675, opened 2019-04-04
Pressing Command/Ctrl S saves project

pwjablonski:
@outoftime 

This PR addresses #1301. I got another little burst of momentum. Pretty straight forward here following #1615 for capturing key command in ace. 

A couple of notes:
1) This only works when the editor is in focus
2) Should I be using logic now instead of sagas? If so there is quite a bit to be migrated over since saveCurrentProject is used throughout the file. This migration process might warrant a separate PR. I'm also entirely sure how logics should be structured. Is each logic housed in a separate file? 
3) I haven't written tests yet but waiting for feedback and answers to question 2. 

Looking forward to hearing your thoughts!


src/components/Editor.jsx
```diff
@@ -130,6 +130,13 @@ class Editor extends React.Component {
           this.props.onAutoFormat();
         },
       });
+      this._editor.commands.addCommand({
```
pwjablonski: An `addCommands` method would be nice here to combine with the above but doesn't look like that is an option.

src/sagas/projects.js
```diff
@@ -143,6 +148,15 @@ export function* saveCurrentProject() {
   }
 }
 
+export function* saveProjectSaga() {
+  const isLoggedIn = yield select(isUserAuthenticated);
+  if (isLoggedIn) {
+    yield* saveCurrentProject();
```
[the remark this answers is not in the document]
pwjablonski: Yes - the second was my line of thinking. I don't feel particularly strongly either way but thought if we are going to implement a save feature it should probably actually do the saving.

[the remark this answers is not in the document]
outoftime: Well I think it's debatable whether we are implementing a save feature vs. implementing the _appearance_ of a save feature (that has always existed), but anyway I think that line of reasoning is sensible enough.

pwjablonski:
Hey @outoftime think I'm ready for the next round of review. I have a couple of thoughts/questions

1. Where does it make the most sense to put the HotKeys component/s? Seems a single one in the Workspace component does the trick but also could see how adding decedent hotkeys could give more specificity.
2. I generalized the account-migration and login prompt header and button under the modal definition (I now realize I created a new definition so will go back and combine) Also I think it would make sense update this on the classroom creator modal in a follow-up.
3. Took my first shot a the logic. They are fairly straight forward but not positive the implementation is quite right (especially the delay). I also have some more questions about implementing the tests that I will follow up on.

src/util/keyMap.js
```diff
@@ -0,0 +1,3 @@
+export const keyMap = {
```
outoftime: Seems like this would make sense as the default export?

src/logic/saveProject.js
```diff
@@ -0,0 +1,18 @@
...
+import {openLoginPrompt} from '../actions/ui';
+import {isUserAuthenticated} from '../selectors';
+
+import {saveCurrentProject} from './saveCurrentProject';
+
+export default createLogic({
+  type: 'SAVE_PROJECT',
+  async process({getState}) {
+    const state = getState();
+    const isLoggedIn = await isUserAuthenticated(state);
```
outoftime: No need to `await` a selector

src/logic/saveProject.js
```diff
@@ -0,0 +1,18 @@
+import {createLogic} from 'redux-logic';
+
+import {openLoginPrompt} from '../actions/ui';
+import {isUserAuthenticated} from '../selectors';
+
+import {saveCurrentProject} from './saveCurrentProject';
+
+export default createLogic({
+  type: 'SAVE_PROJECT',
```
outoftime: So I’m hoping to move us away from hard-coding action types as string constants around the codebase—one thing `redux-actions` lets us do is:

```js
export default createLogic({
  type: saveProject.toString(),
```

With `saveProject` being imported from `actions`. What do you think of this?

src/previewSupport/handleKeyEvents.js
```diff
@@ -0,0 +1,18 @@
...
+
+import channel from './channel';
+
+const {SAVE} = keyMap;
+
+export default function handleKeyEvents() {
+  bind(SAVE, handleSave);
+}
+
+function handleSave() {
```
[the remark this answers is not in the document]
pwjablonski: By inline do you mean an anonymous function?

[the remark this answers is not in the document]
outoftime: In this case, yes (although inlining a function, in general, just refers to removing an explicit function definition and moving the code directly into the function had been called from)

.vscode/launch.json
```diff
@@ -0,0 +1,35 @@
+{
```
pwjablonski: Thought I ignored this...

src/css/application.css
```diff
@@ -841,12 +837,18 @@ body {
   text-align: center;
 }
 
-.account-migration__buttons {
+/** @define modal */
```
pwjablonski: I think we'll want  to standardize/ apply these to the assignment-creator at some point

src/logic/projectSuccessfullySaved.js
```diff
@@ -0,0 +1,18 @@
...
+  showSaveIndicator,
+  hideSaveIndicator,
+} from '../actions/ui';
+
+export default createLogic({
+  type: 'PROJECT_SUCCESSFULLY_SAVED',
+  debounce: 1000,
+  async process(data, dispatch, done) {
+    dispatch(showSaveIndicator());
+    setTimeout(() => {
```
pwjablonski: setTimeout this way worked better with the tests as opposed to waiting for the promise to resolve

outoftime: Hmm, so I don’t think this will give us incorrect behavior, since all `redux-logic` cares about is the `done()` callback in this case, but I think `async` functions should generally only resolve when they have actually done their work, which is not the case here. What was the problem with the tests?

src/components/Workspace.jsx
```diff
@@ -175,21 +179,32 @@ export default class Workspace extends React.Component {
     );
   }
 
+  _handleSave() {
+    this.props.onSave();
+    return false;
```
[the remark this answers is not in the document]
pwjablonski: Ah yes the false prevents the default browser save action

See https://www.npmjs.com/package/mousetrap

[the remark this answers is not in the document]
outoftime: OK—I don’t think this is the right method to have that in, then, since a `_handleSave()` method (based on its name) would not naturally know anything about `mousetrap`. I would probably just put it in the handlers themselves e.g.

```js
    const handlers = {
      SAVE: () => {
        this._handleSave();
        return false;
      },
    };
```

src/logic/projectSuccessfullySaved.js
pwjablonski: Looks like this was the issue - https://stackoverflow.com/questions/52177631/jest-timer-and-promise-dont-work-well-settimeout-and-async-function

outoftime:
@pwjablonski looks like you’ve got merge conflicts and build failures, want to clear those up and then I’ll give it a look over?

pwjablonski:
Hey @outoftime looks like my tests are passing locally but not here. The issue seems to be in `src/logic/__tests__/projectSuccessfullySavedTest.js` and idea why this would be?

outoftime:
> Hey @outoftime looks like my tests are passing locally but not here. The issue seems to be in `src/logic/__tests__/projectSuccessfullySavedTest.js` and idea why this would be?

Dealing with some disk space issues on my laptop rn so I can’t test the theory, but I think the issue might be that you need to affirmatively wait for the process method to complete before asserting on its behavior, so:

```js
test('should show project saved successfully', async() => {
  jest.useFakeTimers();
  const dispatch = jest.fn();
  const data = {};
  const logicDone = new Promise((resolve) => {
    projectSuccessfullySaved.process(data, dispatch, resolve);
  });
  expect(dispatch).toHaveBeenCalledWith(showSaveIndicator());
  expect(dispatch).not.toHaveBeenCalledWith(hideSaveIndicator());
  jest.advanceTimersByTime(1000);
  await logicDone;
  expect(dispatch).toHaveBeenCalledWith(hideSaveIndicator());
  expect(done).toHaveBeenCalledWith();
});
```

src/components/PreviewFrame.jsx
```diff
@@ -139,6 +139,10 @@ class PreviewFrame extends React.Component {
     this.props.onConsoleLog(printedValue, compiledProjectKey);
   }
 
+  _handleSave() {
+    this.props.onSave();
+  }
```
outoftime: No big deal either way but I don’t see much advantage to having this in its own method vs. just calling the prop callback directly from the handler.

src/components/Workspace.jsx
```diff
@@ -314,21 +318,32 @@ export default class Workspace extends React.Component {
...
-      <div className="layout">
-        <AssignmentCreator />
-        <TopBar />
-        <NotificationList />
-        <div className="layout__columns">
-          <Instructions />
-          {this._renderInstructionsBar()}
-          <div className="workspace layout__main">
-            {this._renderEnvironment()}
+      <HotKeys handlers={handlers} keyMap={keyMap}>
```
outoftime: I don’t have a super-strong opinion about this, but I think it would be a little nicer to extract a small component to handle this along the lines of `<KeyboardHandler>` so we can avoid having too many implementation details in `<Workspace>`. Best to think of `<Workspace>`’s job as stitching together the top-level components of the application and leave all the details out of it.

src/logic/__tests__/saveProjectTest.js
```diff
@@ -0,0 +1,44 @@
...
+const defaultState = reduce(undefined, {type: 'NONE'});
+
+jest.mock('../../clients/firebase.js');
+
+test('not pristine project should save', async() => {
+  const mockCredential = credentialFactory.build();
+  const mockUser = userFactory.build();
+  let state = reduce(defaultState, projectCreated('123456'));
+  state = reduce(state, updateProjectSource('123456', 'css', ''));
+  state = reduce(state, userAuthenticated(mockUser, [mockCredential]));
```
outoftime: No big deal, but we can do this a little more concisely by using an actual `reduce`:

```js
const state = [
  projectCreated('123456'),
  updateProjectSource('123456', 'css', ''),
  userAuthenticated(mockUser, [mockCredential]),
].reduce(reduce, defaultState);

File: src/logic/__tests__/saveProjectTest.js [hunk 14]
```diff
@@ -0,0 +1,44 @@
...
+});
+
+test('pristine project should not save', async() => {
+  const mockCredential = credentialFactory.build();
+  const mockUser = userFactory.build();
+  let state = reduce(defaultState, projectCreated('123456'));
+  state = reduce(state, userAuthenticated(mockUser, [mockCredential]));
+  const getState = jest.fn(() => state);
+  const result = await saveProject.process({getState});
+  expect(getState).toHaveBeenCalledWith();
```
outoftime: It doesn’t hurt anything really, but I generally don’t explicitly assert that side-effect-free mocks have been called—generally we can infer that they have been called based on the rest of the test passing (concretely: if the logic weren’t getting the information provided by `getState`, then `result` would not be the right thing).

src/logic/shared/saveCurrentProject.js
```diff
@@ -0,0 +1,17 @@
+import {
+  getCurrentProject,
+  getCurrentUserId,
+} from '../../selectors';
+import {isPristineProject} from '../../util/projectUtils';
+import {saveProject} from '../../clients/firebase';
+import {projectSuccessfullySaved} from '../../actions/projects';
+
+export async function saveCurrentProject(state) {
```
outoftime: Probably can make this a default export?

src/logic/shared/saveCurrentProject.js
```diff
@@ -0,0 +1,17 @@
...
+  const userId = await getCurrentUserId(state);
+  const currentProject = await getCurrentProject(state);
```
outoftime: No need for the `await`s here—I think we can assume selectors are synchronous

src/logic/projectSuccessfullySaved.js
outoftime: I will play around with this

src/logic/shared/saveCurrentProject.js
outoftime: And also maybe make an index module that reexports it?

src/logic/projectSuccessfullySaved.js
outoftime: Hey, looks like this will do it:

```diff
diff --git a/package.json b/package.json
index 9ca8782..e3a7863 100644
--- a/package.json
+++ b/package.json
@@ -197,6 +197,7 @@
     "chrono-node": "^1.3.11",
     "classnames": "^2.2.6",
     "css": "^2.2.4",
+    "delay": "^4.3.0",
     "enum": "^2.5.0",
     "enumify": "^1.0.4",
     "es6-set": "^0.1.4",
diff --git a/src/logic/__tests__/projectSuccessfullySavedTest.js b/src/logic/__tests__/projectSuccessfullySavedTest.js
index 02a9eaa..13dc346 100644
--- a/src/logic/__tests__/projectSuccessfullySavedTest.js
+++ b/src/logic/__tests__/projectSuccessfullySavedTest.js
@@ -8,12 +8,13 @@ import {
 test('should show project saved successfully', async() => {
   jest.useFakeTimers();
   const dispatch = jest.fn();
-  const done = jest.fn();
   const data = {};
-  projectSuccessfullySaved.process(data, dispatch, done);
+  const logicDone = new Promise(
+    resolve => projectSuccessfullySaved.process(data, dispatch, resolve),
+  );
   expect(dispatch).toHaveBeenCalledWith(showSaveIndicator());
   expect(dispatch).not.toHaveBeenCalledWith(hideSaveIndicator());
   jest.advanceTimersByTime(1000);
+  await logicDone;
   expect(dispatch).toHaveBeenCalledWith(hideSaveIndicator());
-  expect(done).toHaveBeenCalledWith();
 });
diff --git a/src/logic/projectSuccessfullySaved.js b/src/logic/projectSuccessfullySaved.js
index 46d2727..4e5e9ad 100644
--- a/src/logic/projectSuccessfullySaved.js
+++ b/src/logic/projectSuccessfullySaved.js
@@ -1,4 +1,5 @@
 import {createLogic} from 'redux-logic';
+import delay from 'delay';
 
 import {
   showSaveIndicator,
@@ -10,9 +11,8 @@ export default createLogic({
   debounce: 1000,
   async process(data, dispatch, done) {
     dispatch(showSaveIndicator());
-    setTimeout(() => {
-      dispatch(hideSaveIndicator());
-      done();
-    }, 1000);
+    await delay(1000);
+    dispatch(hideSaveIndicator());
+    done();
   },
 });
```

(The key is to await the `done()` callback getting called in the test)

src/components/PreviewFrame.jsx
pwjablonski: Fair - it seemed a bit more consistent to do it this way, though I can see how its more verbose

src/logic/shared/saveCurrentProject.js
pwjablonski: Had trouble with linting the index module so just changed to default

src/containers/KeyboardHandler.js
```diff
@@ -0,0 +1,22 @@
+import {connect} from 'react-redux';
+
+import KeyboardHandler from '../components/KeyboardHandler';
+
+import {saveProject} from '../actions';
+
+function mapStateToProps() {
```
pwjablonski: Should this be left here?

src/components/KeyboardHandler.jsx
```diff
@@ -0,0 +1,23 @@
...
+      onSave();
+      return false;
+    });
+
+    return () => {
+      mousetrap.unbind('mod+s');
+    };
+  }, [onSave]);
+
+  return <>{children}</>;
```
outoftime: Per IRL discussion, taking children here may give the impression that the keypresses are only captured if the children are focused, but that is not the case—better to make it a void component.

Issue #1680, opened 2019-04-12
Immediate visual feedback when I click to create an assignment


outoftime:
Right now when I click one of the buttons in the assignment creator field, nothing happens until the assignment creation process completes successfully, at which point the modal closes and a notification is displayed.

I should get visual feedback that something is happening as soon as I click, though—at a minimum, greying out the buttons and changing the text of the one I clicked to something like *Saving…*

Followup to #1511

outoftime:
Closing since we pulled the Classroom integration.

PR #1681, opened 2019-04-12
Saving assignment indicator

pwjablonski:
Hey @outoftime this addresses #1680. I'm not a huge fan of using the spinner again but it was too easy to implement. Also not sure if the `exportingAssignment` flag should live in the client reducer or the googleClassroom record. Let me know what you think.


src/components/AssignmentCreatorForm.jsx
```diff
@@ -69,43 +70,61 @@ export default function AssignmentCreatorForm({
...
+              className={classnames(
+                'assignment-creator__button',
+                'assignment-creator__button_disabled',
+              )}
+              type="button"
+            >
+              {t('assignment-creator.creating')}
+            </button>
+          </div> :
+          <div>
```
outoftime: You can use a [fragment](https://reactjs.org/docs/fragments.html) here! No need for the extra `<div>` just for JSX structural reasons.

PR #1686, opened 2019-05-02
1260: Normalize console errors

omardeleo:
The purpose of this PR is to normalize the shape of all errors generated in Popcode. Console errors now have the same shape as errors generated from the JS editor, and are also mapped to more reader-friendly error messages. This addresses #1260.

**Console error message (Before)**
Screen Shot 2019-05-30 at 3 55 18 PM

**Console error message (After)**
Screen Shot 2019-05-30 at 4 00 21 PM

**Console error shape (Before)**
Screen Shot 2019-05-30 at 3 59 21 PM

**Console error shape (After)**
Screen Shot 2019-05-30 at 4 01 48 PM


omardeleo:
Hey @outoftime for some weird reason `yarn test` passed all tests locally on my machine, even though I see the linting errors when I click the Travis CI link you sent. I'll go ahead and make corrections, but do you have any idea why that could be?

Screen Shot 2019-05-16 at 11 58 36 AM

omardeleo:
**Update:** When I try to correct the linting errors, I then fail `yarn test` :////

Screen Shot 2019-05-16 at 12 10 35 PM

outoftime:
@omardeleo it looks to me like the error in Travis is the **same** one you are seeing locally—namely Popcode’s style guide says you should not put spaces inside `{` and `}`. Have you pushed your latest changes up to GitHub?

Anyway, I would try this just to be sure:

```sh
$ yarn
$ yarn run fixlint
```

And then commit/push anything that changes.

src/reducers/__tests__/console.test.jsx
```diff
@@ -37,14 +38,19 @@ test('consoleValueProduced adds value to existing entry', () => {
...
   const key = '123';
-  const name = 'NameError';
-  const message = 'bogus is not defined';
+  const error = {
+    name: 'TypeError',
+    raw: 'You tried to call `i()` as a function, but `i` is not a function.',
+    reason: 'not-a-function',
+    text: 'You tried to call `i()` as a function, but `i` is not a function.',
+    type: 'error',
+  };
```
outoftime: OK ONE MORE little thing here for total perfection: this kind of sentinel value for a piece of structured data is the perfect use case for a factory. Check out the factories in [`__factories__/clients`](https://github.com/popcodeorg/popcode/tree/master/__factories__/clients), for examples, although in this case I think the right place for your factory would be `__factories/validations/errors.js` since this object shape is based on the shape of the errors the validation module returns.

omardeleo: Created `consoleErrorFactory` and added to `__factories/validations/errors.js`.

Issue #1443, opened 2018-05-04
Switch from tape to jest


outoftime:
Jest has become the pretty clear favorite for testing the kind of JavaScript application that Popcode is. I’ve never been amazingly happy with tape anyway, so let’s switch to the framework that will be most familiar to contributors.

Also, the new tests should avoid repeating the mistakes I made in structuring the old test suite. Specifically:

* The reducer tests use the `reducerTest` helper function, which takes a before-state, an action, and an expected after-state. This however does not match the way one naturally thinks about testing reducers, namely: start with this state; apply this action; then make _specific assertions_ about the resulting state (rather than just saying “this is what the resulting state should be”).
* Reusable test helpers, particularly for creating test data, are unintuitive and not well-structured. We should have factory modules for each category of object we’d like to create: e.g., one for objects that the Firebase SDK would give us; one for objects that the GitHub API would give us; etc. We should use something like [`rosie`](https://github.com/rosiejs/rosie) to make factories simple to build and easy to extend.
* Test code should not do direct manipulation on Redux state data structures. Redux state used in tests should always be generated by running a series of actions through the reducers. We should have a helper module to make this boilerplate-free.

## Reducers

- [ ] `clients`
- [ ] `compiledProjects`
- [x] `console`
- [ ] `currentProject`
- [ ] `errors`
- [ ] `googleClassroom`
- [ ] `index`
- [ ] `projects`
- [ ] `resizableFlex`
- [ ] `ui`
- [ ] `user`

## Sagas

Test should be ported as part of migration to `redux-logic`.

## Validations

Validation tests will likely involve no major changes _other_ than moving from Tape to Jest. So it might be worth exploring the use of a [codemod](https://github.com/skovhus/jest-codemods) to do this automatically.

- [ ] `html`
- [ ] `css`
- [ ] `javascript`

## Miscellaneous

Other Tape tests that can be found in `test/unit`:

- [ ] `higherOrderComponents/resizableFlex/calculateFlexGrowAfterDrag`
- [ ] `localization/localization`
- [ ] `Analyzer`
- [ ] `util/javascript`
- [ ] `util/queryParams`

PR #1687, opened 2019-05-09
Refactor console reducer to use handleActions, test with Jest

outoftime:
Introduce two patterns to the console reducers that we would like to eventually use for all reducers:

* Reducers use `handleActions()` rather than a hand-rolled switch statement; these use the action creators themselves as keys to avoid hard-coding constant names
* Reducer is tested with Jest; tests run a series of actions through the reducer and then make discrete assertions about final state


src/reducers/__tests__/console.test.jsx
```diff
@@ -0,0 +1,163 @@
...
+  consoleLogBatchProduced,
+  consoleValueProduced,
+  evaluateConsoleEntry,
+  nextConsoleHistory,
+  previousConsoleHistory,
+  consoleInputChanged,
+} from '../../actions';
+
+test('evaluateConsoleEntry adds entry to history', () => {
+  const expression = '1 + 1';
```
jwang1919: These consts could be a Rosie Factory object, where key is '123' and maybe you can override with that expression and value when appropriate

src/reducers/__tests__/console.test.jsx
```diff
@@ -0,0 +1,163 @@
...
+  const key = '123';
+  const {history} = applyActions(
+    evaluateConsoleEntry(expression, key),
+  );
+  expect(history.size).toBe(1);
+  expect(history.get(key)).toMatchObject({expression});
+});
+
+test('consoleValueProduced adds value to existing entry', () => {
+  const value = 2;
```
jwang1919: Rosie here too?

src/reducers/__tests__/console.test.jsx
```diff
@@ -0,0 +1,163 @@
...
+  const {history} = applyActions(
+    evaluateConsoleEntry('1 + 1', key),
+    consoleValueProduced(key, value),
+  );
+
+  expect(history.get(key)).toMatchObject({value});
+});
+
+test('consoleErrorProduced adds error to existing entry', () => {
+  const key = '123';
```
jwang1919: Rosie here too?

src/reducers/__tests__/console.test.jsx
```diff
@@ -0,0 +1,163 @@
...
+
+test('consoleInputChanged updates currentInputValue', () => {
+  const value = '3';
+  const {currentInputValue} = applyActions(consoleInputChanged(value));
+
+  expect(currentInputValue).toEqual(value);
+});
+
+test('consoleLogBatchProduced adds entries to history', () => {
+  const entries = [
```
jwang1919: Rosie here too?

src/reducers/console.js
```diff
@@ -17,85 +32,75 @@ function updateConsoleForHistoryIndex(state, index) {
...
-          payload.compiledProjectKey,
-        ),
-      );
-    case 'EVALUATE_CONSOLE_ENTRY':
-      return payload.trim(' ') === '' ? state : state.setIn(
-        ['history', meta.key],
-        new ConsoleEntry({expression: payload}),
-      ).
+  [evaluateConsoleEntry]: (state, {payload: expression, meta: {key}}) =>
+    expression.trim(' ') === '' ?
```
jwang1919: Is this a String `trim` method? If so it shouldn't require `' '` as an argument.

jwang1919: Is `expression` an object? I thought `trim` doesn't require arguments.

outoftime: gc

src/reducers/__tests__/console.test.jsx
outoftime: Hmm, so I generally think of the main goals of using a factory like Rosie to be to reduce duplication and to centralize knowledge about object structure in one place. Do you think either of those goals would be served by bringing a factory into the picture here? What kind of objects would the factory actually produce? Where in the application code would objects with the same shape appear?

I would characterize what we’re doing here as creating test sentinels—values that are basically meaningless except that we want to give them as input to the code under test, and then check to see if they are in the right place in the output. This is basically the same thing that factories do, but factories always produce objects, and specifically they product objects with a shape that you would find somewhere else in application code (e.g. we are using factories to produce objects with the same shape that the GitHub and Firebase SDKs produce, etc.)

So I guess my question is, what shape object are you expecting to see here? What would the factories actually produce?

src/reducers/__tests__/console.test.jsx
```diff
@@ -25,11 +28,12 @@ test('evaluateConsoleEntry adds entry to history', () => {
 });
 
 test('consoleValueProduced adds value to existing entry', () => {
-  const value = 2;
-  const key = '123';
+  const consoleValueProducedAction =
+    actionFactory[consoleValueProduced].build();
```
outoftime: Obviously this could be structured more normally as ES6 modules that just have a named export for each action factory, but I like this a little better because:

* It makes it very clear that the factory is _not_ an action creator function
* It ties the factory explicitly to the action creator itself

src/reducers/__tests__/console.test.jsx
outoftime: @jwang1919 so the most obvious answer I could think of for what a factory would generate here was the action itself—essentially wrapping the action creator but having sensible defaults/sequences for all the args that get passed in. Take a look at [this commit](https://github.com/popcodeorg/popcode/pull/1687/commits/f11a2380365a6bb0dd8f3858b27f463770fb1ca0) and see if you think it’s an improvement; if so, I will apply the pattern to all the action creators used in that test.

jwang1919: The intention was to remove the repetitive lines of code where `key`, `value`, `expression` and `compiledProject` were declared. The secondary intention was to avoid hard-coding expected values in the assertions by using the factory created objects instead.

I only thought Rosie was being used as a means to generate test objects. I do agree it makes more sense to have Rosie generate more sensible objects that can be seen in the application.

TLDR; I'm okay skipping this suggestion.

outoftime: Gotcha. What do you think of the [experiment with using factories to generate actions in tests](https://github.com/popcodeorg/popcode/pull/1687/commits/f11a2380365a6bb0dd8f3858b27f463770fb1ca0)? I am on the fence—I like removing the burden of generating sentinel values from the tests themselves, but I worry that it creates extra test boilerplate without a really compelling improvement to abstraction/encapsulation. Also, I think it’s probably better to have tests coupled to the interface of action creators (by calling them directly) than to the output of action creators (by extracting values from the action payloads to make assertions about).

jwang1919: Yeah I would skip it, I think it is extra syntactic sugar that would confuse new developers.

outoftime: OK!

outoftime:
Step toward #1690 
Step toward #1443

PR #1689, opened 2019-05-12
Have `Workspace` handle collapsed column logic

alexpelan:
If every component in a column is collapsed, instead render that component below the editors so that the remaining column gets the entirety of the width. 

This is part 2 of 3 for #1340. The third part will be making the preview window itself minimize-able.

Relevant discussion for this approach: https://github.com/popcodeorg/popcode/pull/1631#pullrequestreview-209893749


alexpelan:
Let me know how you want me to squash up the commit messages, I forget what the convention is.

alexpelan:
And here's a screenshot of all three editors collapsed, giving us a full-width preview. Should we think about moving the console text to the right in this case (or all cases)? 
Screen Shot 2019-05-12 at 11 53 14 AM

outoftime:
> Let me know how you want me to squash up the commit messages, I forget what the convention is.

We do squash-and-merge now so do whatever you want!!

outoftime:
> And here's a screenshot of all three editors collapsed, giving us a full-width preview. Should we think about moving the console text to the right in this case (or all cases)?
> Screen Shot 2019-05-12 at 11 53 14 AM

Good catch. We should definitely make the alignment of the labels consistent in this case, although I would lean toward left-aligning all of them. I’m fine with leaving this for a followup.

src/components/Workspace.jsx
```diff
@@ -133,9 +138,88 @@ export default class Workspace extends React.Component {
...
   }
 
+  _renderHiddenLanguages() {
+    const {
+      currentProject,
+      hiddenLanguages,
+      onComponentToggle,
+    } = this.props;
+    return (
+      <React.Fragment>
```
outoftime: I think we prefer the empty `<>` style for fragments (seems like there should be a linter rule for this?)

src/components/Workspace.jsx
```diff
@@ -133,9 +138,88 @@ export default class Workspace extends React.Component {
...
+  }
+
+  _renderHiddenLeftColumnComponents() {
+    const {
+      hiddenLanguages,
+    } = this.props;
+
+    if (
+      hiddenLanguages.length === 0 ||
+      hiddenLanguages.length === LANGUAGES.length
```
outoftime: I think it would make semantic sense to delegate this to `!_shouldRenderLeftColumn()`

src/components/Workspace.jsx
```diff
@@ -133,9 +138,88 @@ export default class Workspace extends React.Component {
...
+    return this._renderHiddenLanguages();
+  }
+
+  _renderHiddenRightColumnComponents() {
+    const {
+      currentProject,
+      onComponentToggle,
+    } = this.props;
+    const rightColumnComponents = ['console'];
+    return rightColumnComponents.
```
outoftime: So unless I’m misreading here, `rightColumnComponents` is always a one-element array containing just `'console'`, right? In which case it seems like the `filter`/`map` of it all is a bit of overkill?

src/components/Workspace.jsx
```diff
@@ -188,6 +298,7 @@ export default class Workspace extends React.Component {
             {this._renderEnvironment()}
           </div>
         </div>
+        {this._renderFullyMinimizedColumns()}
```
outoftime: A couple of thoughts on this. First, in the case where one column is totally collapsed, I think we would ideally put all the hidden components together in the DOM—so basically I would move this up to right below the call to `_renderHiddenRightColumnComponents()`.

Second, I would propose rearranging the logic a little bit:

* Remove the check for “all left column components hidden” from `_renderHiddenLeftColumnComponents()`, and inline `_renderHiddenLanguages()`. The method already won’t be called if all the components are hidden, because of the check on line 238.
* Get rid of the `_renderFullyMinimizedColumns()` method (I didn’t find “fully minimized” an intuitive description on the first read-through of the code) and just inline a check here (or wherever this line ends up), i.e. `{!this._shouldRenderLeftColumn() && this._renderHiddenLeftColumnComponents()}`

src/components/Workspace.jsx
```diff
@@ -133,9 +138,88 @@ export default class Workspace extends React.Component {
...
+
+  _renderFullyMinimizedColumns() {
+    if (!this._shouldRenderLeftColumn()) {
+      return this._renderHiddenLanguages();
+    }
+    return null;
+  }
+
+  _shouldRenderLeftColumn() {
+    return !(this.props.hiddenLanguages.length === LANGUAGES.length);
```
outoftime: More idiomatic to use `!==` (or in this case `<` would work) rather than negating the whole expression

src/components/Workspace.jsx
alexpelan: went with `_shouldRenderHiddenLeftColumnComponents` to match

src/components/Workspace.jsx
alexpelan: For now...the thought was the next revision is going to add the preview to the list of right column components and then it'll just be adding that to the array.

PR #1692, opened 2019-05-23
Run all node_modules through babel

wylieconlon:
This is a revert of a revert, previously reviewed at #1674 

This PR causes a runtime error that prevents Popcode from loading at all in Chrome 40 and below.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -29,10 +29,5 @@ module.exports = (api) => {
     ],
     plugins,
     compact: false,
-    overrides: [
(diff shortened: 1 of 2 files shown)


wylieconlon:
With the latest change to exclude `@babel/plugin-transform-typeof-symbol`, the build is compiled with compatibility for Chrome 38 and above.

wylieconlon:
Closing for https://github.com/popcodeorg/popcode/pull/1705

PR #1701, opened 2019-06-08
Complete migration of user.js redux-saga to redux-logic

jwang1919:
Converted logout and startAccountMigration sagas.


__factories__/clients/github.js
```diff
@@ -45,3 +45,17 @@ export const githubProfileFactory = new Factory().attrs({
     private_repos: 10000,
   },
 });
+
+export const githubRepositoryFactory = new Factory().attrs({
```
outoftime: This looks like it’s in the shape of our own stored Firebase data, not something we would get back from the GitHub client, right? If that’s correct I would put it in a separate top-level factory module, along the lines of `__factories__/data/firebase.js` (I don’t think it belongs in `clients/firebase.js` because the shape is determined by how we choose to store the data, not anything intrinsic about the Firebase client).

__factories__/clients/github.js
```diff
@@ -45,3 +45,17 @@ export const githubProfileFactory = new Factory().attrs({
...
+  instructions: '',
+  isArchived: false,
+  projectKey: '123456789',
+  sources: {
+    css: '',
+    html: '<\!DOCTYPE html>' +
+      '<html><head><title>Sample Project</title></head><body></body></html>',
+    javascript: '',
+  },
+  updatedAt: 1560012978057,
```
outoftime: Can we make this a dynamic attribute that uses the current time?

__factories__/clients/github.js
```diff
@@ -45,3 +45,17 @@ export const githubProfileFactory = new Factory().attrs({
     private_repos: 10000,
   },
 });
+
+export const githubRepositoryFactory = new Factory().attrs({
+  hiddenUIComponents: ['console'],
+  instructions: '',
+  isArchived: false,
+  projectKey: '123456789',
```
outoftime: Can we make this a dynamic attribute that generates a project key? In particular if a test needs to generate two projects from this factory, we would expect them to have different keys.

src/logic/__tests__/startAccountMigration.test.js
```diff
@@ -0,0 +1,107 @@
...
+} from '@factories/clients/firebase';
+
+import {
+  githubRepositoryFactory,
+} from '@factories/clients/github';
+
+
+jest.mock('../../actions/user.js');
+jest.mock('../../clients/firebase.js');
+jest.mock('../../selectors');
```
outoftime: Is this necessary?

src/logic/__tests__/startAccountMigration.test.js
```diff
@@ -0,0 +1,107 @@
...
+jest.mock('../../util/bugsnag.js');
+
+const getState = jest.fn();
+const dispatch = jest.fn();
+const done = jest.fn();
```
outoftime: It doesn’t _really_ matter, but I think it’s a little nicer to put stuff like this inside the `describe` block they are being used by, just to make clear what their scope is. Feel free to ignore this though, works fine either way.

src/logic/__tests__/startAccountMigration.test.js
```diff
@@ -0,0 +1,107 @@
...
+      githubRepositoryFactory.build(),
+      githubRepositoryFactory.build(),
+    ];
+    migrateAccount.mockResolvedValue({
+      user: mockUser,
+      migratedProjects: mockProjects,
+    });
+
+    const emptyDispatch = new Observable((subscriber) => {
+      subscriber.complete();
```
outoftime: I don’t think the `action$` observable would ever be expected to complete in the real application—that would essentially imply that there will be no more actions dispatched to Redux, which isn’t a situation that would ever obtain. So I think this test is actually simulating an impossible situation, specifically that `cancelPromise` in the logic resolves to a `false` value.

I believe the situation we want to be testing here (the user does not cancel the migration) is that the `continuePromise` resolves to `true`, which would mean having this observable be _truly_ empty (does not dispatch or complete), and then use a [jest timer mock](https://jestjs.io/docs/en/timer-mocks) to simulate the timeout completing without making the test actually sit there for five seconds.

src/logic/__tests__/startAccountMigration.test.js
```diff
@@ -0,0 +1,107 @@
...
+    const cancelDispatch = new Observable((subscriber) => {
+      subscriber.next({type: 'DISMISS_ACCOUNT_MIGRATION'});
+      subscriber.complete();
+    });
+    await startAccountMigration.process(
+      {action$: cancelDispatch, getState},
+      dispatch,
+      done,
+    );
+    expect(accountMigrationUndoPeriodExpired).not.toHaveBeenCalled();
```
outoftime: Corollary of the above, but this assertion is considerably more meaningful if we run out the timer with a timer mock.

src/logic/startAccountMigration.js
```diff
@@ -0,0 +1,54 @@
...
+export default createLogic({
+  type: 'START_ACCOUNT_MIGRATION',
+  async process({getState, action$}, dispatch, done) {
+    const continuePromise = new Promise((resolve) => {
+      setTimeout(resolve, 5000, false);
+    });
+
+    const cancelPromise = action$.pipe(first(
+      ({type}) => type === 'DISMISS_ACCOUNT_MIGRATION',
+      false,
```
outoftime: Closely linked to my [comment on the test above](https://github.com/popcodeorg/popcode/pull/1701/files#diff-37da696ec8de026c3ea562a8340aeb2bR51), I don’t think the optional default value makes sense here—`action$` is a continuous stream of actions that continues for the lifetime of the application. A default value for `first()` would only be needed if the observable might complete without ever emitting another value, which as discussed above is not something we would expect to happen in real-world code (and thus should not simulate in tests either).

src/logic/__tests__/startAccountMigration.test.js
```diff
@@ -0,0 +1,107 @@
...
+const getState = jest.fn();
+const dispatch = jest.fn();
+const done = jest.fn();
+
+describe('startAccountMigration', () => {
+  test('not dismissed during undo period, successful migration', async() => {
+    const mockCredential = credentialFactory.build();
+    getCurrentAccountMigration.mockResolvedValue({
+      firebaseCredential: mockCredential,
+    });
```
outoftime: Rather than mocking the behavior of the selector, let’s just give it a `state` object that contains the data in question. That way, if we decide to change how selectors are used in the future in the logic (in a way that still produces the correct outcome), the test won’t break.

The preferred way to create a state object is to just run a series of actions through the reducer and grab the output. See the [console reducer test](https://github.com/popcodeorg/popcode/blob/master/src/reducers/__tests__/console.test.jsx#L143-L149) for a good example. In this case though we will want to run the actions through the root reducer, rather than just the reducer for the console namespace.

src/logic/__tests__/startAccountMigration.test.js
jwang1919: Yes.

src/logic/__tests__/startAccountMigration.test.js
jwang1919: May need help on this one, currently seems that Promises and timer mocks don't play well.
Used solution here: https://stackoverflow.com/questions/51126786/jest-fake-timers-with-promises, but can't seem to get it to use await sync properly.

src/logic/__tests__/startAccountMigration.test.js
outoftime: Even in the context of [this](https://github.com/popcodeorg/popcode/pull/1701#discussion_r291980415)?

jwang1919: I wrote that comment before I read that part ;). Working on fixing that then I'll adapt.

src/logic/__tests__/startAccountMigration.test.js
jwang1919: Okay looked into this, but got a stuck when running the reducer.

The `action` parameter is coming in null for me:
https://github.com/popcodeorg/popcode/blob/master/src/reducers/user.js#L64

My reducer code looks something like this:
```
import reduce from 'lodash-es/reduce';
import reduceRoot from '../../reducers/index';
...
// inside test
const state= applyActions(
    startAccountMigrationAction,
    accountMigrationUndoPeriodExpired,
    accountMigrationNeeded({credential: mockCredential}),
);
...
function applyActions(...actions) {
  return reduce(
    actions,
    (state, action) => reduceRoot(state, action),
    undefined,
  );
}
```

outoftime: Hmm well one thing that jumps out is you’ve gotta call all the action creators, e.g. `startAccountMigrationAction()` not `startAccountMigrationAction`

src/logic/__tests__/startAccountMigration.test.js
outoftime: Can you push the non-working test so I can play around with it?

outoftime: One thing that I imagine might not be intuitive is that you will need to hold a reference to the promise; then advance the timers; then await the promise, something like:

```js
const migrationDone = startAccountMigration.process(
  {action$: emptyDispatch, getState},
  dispatch,
   done,
);
jest.runAllTimers();
await migrationDone;
```

Not sure if that’s helpful but figured I’d throw it out there.

jwang1919: Ah yeah that actually works out.

__factories__/data/firebase.js
```diff
@@ -0,0 +1,18 @@
+import {Factory} from 'rosie';
+
+export const firebaseRepositoryFactory = new Factory().attrs({
```
outoftime: I’d probably give this a name that makes it clear that it is _project_ data stored in Firebase (since there are several other data types/shapes we store in Firebase alongside it)

src/logic/startAccountMigration.js
```diff
@@ -0,0 +1,53 @@
...
+      continuePromise,
+      cancelPromise,
+    ]);
+
+    if (shouldCancel) {
+      return;
+    }
+
+    await dispatch(accountMigrationUndoPeriodExpired());
+    const {firebaseCredential} = await getCurrentAccountMigration(getState());
```
outoftime: No need to `await` here I don’t think?

src/logic/__tests__/startAccountMigration.test.js
outoftime: I don’t think we need this mock?

src/logic/__tests__/startAccountMigration.test.js
```diff
@@ -0,0 +1,137 @@
...
+
+import {
+  credentialFactory,
+  firebaseErrorFactory,
+  userFactory,
+} from '@factories/clients/firebase';
+import {githubProfileFactory} from '@factories/clients/github';
+import {firebaseProjectFactory} from '@factories/data/firebase';
+
+jest.mock('../../clients/firebase.js');
```
outoftime: Nit, but I don’t think you need to write the `.js` extension here :slightly_smiling_face:

src/logic/__tests__/startAccountMigration.test.js
```diff
@@ -0,0 +1,137 @@
...
+jest.mock('../../selectors');
+jest.mock('../../util/bugsnag.js');
```
outoftime: Ibid. `.js`

src/logic/__tests__/startAccountMigration.test.js
```diff
@@ -0,0 +1,137 @@
...
+  const done = jest.fn();
+
+  test('not dismissed during undo period, successful migration', async () => {
+    const mockUser = userFactory.build();
+    const mockCredential = credentialFactory.build();
+    const mockProfile = githubProfileFactory.build();
+    const mockProjects = [
+      firebaseProjectFactory.build(),
+      firebaseProjectFactory.build(),
+    ];
```
outoftime: Small thing, but you can use the [`buildList()`](https://github.com/rosiejs/rosie#buildlist) function to avoid having to manually create multiple instances

jwang1919:
Please hold off from merging, I have not been able to test a full migration with this code yet. Experiencing issues authenticating with the github account.

Issue #1574, opened 2018-09-24
Directly import SVG icons


outoftime:
Rather astonishingly, the `index.es.js` files in the FontAwesome packages just inline all the SVG definitions and then export them. This means that we’re bundling every single FontAwesome icon just to get the ten or so we actually use.

Fortunately, the icons also appear to be exposed as individual modules. So we should be able to simply go through and replace things like:

```js
import {faExchangeAlt} from '@fortawesome/free-solid-svg-icons';
```

With:

```js
import faExchangeAlt from '@fortawesome/free-solid-svg-icons/faExchangeAlt';
```

And get a big bundle size improvement.

---

Note that this is the kind of thing that [tree shaking](https://webpack.js.org/guides/tree-shaking/) is supposed to take care of, but it doesn’t look like the FontAwesome packages specify `sideEffects`, and I haven’t dug too deeply into what the affordances are for us to declare that at the project root level. I certainly wouldn’t _mind_ a tree-shaking based fix but I’m also fine with simply changing all the imports along the lines specified above.

outoftime:
Per discussion in #1710, we are getting the right outcome here already—the only SVGs that appear in the compiled bundle are the ones we’re actually using. I think I was misunderstanding how tree-shaking/dead code elimination works.

PR #1710, opened 2019-07-12
Directly import svg icons

razzius:
https://github.com/popcodeorg/popcode/issues/1574

Before this change
image

After this change
image


razzius:
Strangely the gzip size is higher after directly importing the icons...

outoftime:
> Strangely the gzip size is higher after directly importing the icons...

Weird, any theories as to why? Have you taken a look at the bundle profiles before/after?

razzius:
Maybe it has something to do with the gzip algorithm; by pulling in multiple modules separately, the similarity of each svg is not as easily identifiable as compressible. I can take a look at the profiles and see if anything obvious comes to light but it's not something I'm super familiar with.

outoftime:
Hey @razzius did you get a chance to look at the profiles?

razzius:
No, haven't taken another look, and I'm not sure what I could look at beyond the screenshots above. Any ideas?

outoftime:
@razzius OK I think I know what’s going on—the `free-solid-svg-icons` package has an `index.es.js` package that does ES exports, which is properly configured as the `module` entrypoint in its `package.json`; apparently that’s enough for Webpack to tree-shake out the icons we don’t import. I have confirmed this by grepping the compiled production JS bundle for the SVG paths of icons we do and don’t use—the ones we do use appear in the bundle, but the ones we don’t use don’t.

So, I think probably the current approach is preferable—I’m guessing it’s slightly smaller because there is a little extra boilerplate for each imported module, and `index.es.js` inlines all the icons. It’s also a bit more concise to just do named imports out of the index module. So I am going to close this—sorry for wasting your time with my incomplete understanding of the way the existing imports were working!!

PR #1711, opened 2019-07-12
Swap out moment for date-fns

razzius:
PROOF that the date functionality still works (time from now is displayed in projects list):

image

The other use case is formatting a date: 

```js
> new moment().format('MMM Do [at] h:mm A')
'Jul 11th at 7:43 PM'
> format(new Date(), 'MMM Do [at] h:mm A')
'Jul 11th at 7:43 PM'
```


razzius:
The bundle sizes are actually a regression, probably because `chrono-node` (a natural language date parser) depends on moment.

Stats from `env PROFILE_BUILD=true npm run preview`
http://localhost:3000/profile/webpack-visualizer.html:

Before:
Stat 11.09 MB
Parsed 4.91 MB
Gzipped 1.3 MB

After:
Stat 11.49 MB
Parsed 5.09 MB
Gzipped 1.33 MB

razzius:
The date parsing is occurring in `AssignmentCreatorForm.jsx`

I've never used the `AssignmentCreator` functionality. Just to make sure, can anybody comment on how it's part of the app, maybe with a helpful screenshot?

razzius:
If we removed chrono-node, the bundle is as expected significantly smaller.

Stat 10.81 MB
Parsed 4.74 MB
Gzipped 1.25 MB

webpack.config.js
```diff
@@ -276,11 +276,6 @@ module.exports = (env = process.env.NODE_ENV || 'development') => {
           ],
           use: ['null-loader'],
         },
-        {
-          test: /\.js$/u,
-          include: matchModule('moment/locale'),
-          use: ['null-loader'],
-        },
```
outoftime: I’m guessing this might be why the bundle size increased substantially? (Just adding `date-fns` should not move the needle much)

paulproteus:
I see that `chrono-node` now depends on `dayjs` as its one dependency. https://github.com/wanasit/chrono/blob/master/package.json#L49 Perhaps someone else coming by this PR will find an interest in checking the size impact of upgrading chrono as well as migrating popcode itself to dayjs.

This PR as-is could be a good approach, too!

PR #1732, opened 2019-08-01
New True Development Environment

outoftime:
After facing the fact that Docker adds a substantial performance cost in the local development environment, build a new One True Development Environment based on [`nodeenv`](https://github.com/ekalinin/nodeenv) (not to be confused with `nodenv`, an NVM analog/alternative). Nodeenv is essentially virtualenv for Node; it allows us to set up an isolated Node environment in a directory in the project root.

The only dependency of Nodeenv is Python 2.7+, which should already be installed on nearly any machine someone is using for development. Installation is quick and painless, and it’s straightforward to reset one’s environment and rebuild from scratch by just deleting the whole nodeenv directory.

While `nodeenv` (like virtualenv) has an affordance for “activating” the environment in the current shell session, thus putting all `nodeenv` executables on the PATH, we also provide a set of shims in the `tools` directory to run useful executables like `yarn`, `node`, `npx`, etc. in the nodeenv without having to activate it in the shell. All `tools/` scripts degrade gracefully if there is no nodeenv installed, so folks who prefer to maintain their own Node version can still use the VS Code configurations that invoke the `tools/` scripts if desired.

The setup script and all the `tools/` wrappers are written in Python, allowing us to maintain a single set of script for OS X/Linux/Windows, and introducing no extra dependencies beyond what nodeenv already needs.


jwang1919:
Looking good! Can confirm it will work on Windows computers running Python 3

PR #1737, opened 2019-08-15
moves project picker drop down functionality into project picker modal

### LINKED ISSUE #768: Add more project management functionality
missing:
- delete project
- change project title

would be nice:
- download entire project as zip

---

👍 for this project, btw.

pwjablonski:
Hey @outoftime this is the first draft for creating a project picker modal. 

A few  things this PR does
- Removes ProjectPicker dropdown menu and replaces it with ProjectPickerButton
- Clicking on ProjectPickerButton opens ProjectPickerModal
- ProjectPickerModal displays a list of ProjectPreviews by filter type (active or archived)
- Clicking on the ProjectPickerModal navigation tabs updates the filter type
- For active projects  (non-archived projects) 
  -Clicking on ProjectPreview changes the current project and closes the ProjectPickerModal
  -Clicking on the archive button changes that project's archived flag to true
- For archived projects 
  -Clicking on ProjectPreview changes the current project,  changes that project's archived flag to false, and closes the ProjectPickerModal

Hoping to get some high-level feedback on the approach before cleaning up some of the lower-level stuff. 

In particular, I'm curious about your thoughts on
- Layout
- User flow for archiving,  unarchiving, selecting projects
- Does filtering in this way make sense or would it be better to have dedicated components for each view.

Looking forward!

Closes #768


config/lintignore
```diff
@@ -5,3 +5,5 @@
 /nodeenv
 /test/data
 /templates
+/.cache
+/nodeenv
```
outoftime: Not needed—both of these appear earlier in this file.

outoftime:
BTW I haven’t actually looked at the code for the most part, since it sounds like you’re mostly looking for feedback on functionality/UI at this point. Let me know when you’re ready for me to look at code!

src/components/Modal.jsx
```diff
@@ -23,3 +25,12 @@ Modal.propTypes = {
 Modal.defaultProps = {
   isOpen: true,
 };
+
+const ModalContents = onClickOutside(
```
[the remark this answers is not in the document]
outoftime: This seems fine! A simpler approach might be to eschew `onClickOutside` altogether in this case, since there’s  only one element, the `div.modal`, that we’re listening for clicks on. So you could just add an `onClick` to the outermost div, and then a second one to `.modal__contents` to stop propagation.

src/components/ProjectPickerModal.jsx
```diff
@@ -0,0 +1,88 @@
...
+  projectsFilter,
+  onCloseProjectPickerModal,
+  onFilterProjects,
+}) {
+  if (!isOpen) {
+    return null;
+  }
+
+  let visibleProjects;
+  const activeProjects = filter(projects, ({isArchived}) => !isArchived);
```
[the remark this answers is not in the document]
outoftime: > Can/should this be extracted into selectors? ie. getArchivedProjects, getActiveProjects?

Yeah, definitely—right now we’re doing linear-time work here (filtering the projects takes time proportional to the number of projects) in the render method; using a reselect selector gets that down to constant time in the average case.

> Also is filtering the right way to be thinking about this functionality?

It’s entirely sensible, although in this case you’re doing more work than you need to be, doing two full passes over the collection:

* If the active tab is highlighted, you need the collection of active projects, but you only need to know if there is _any_ archived project. The latter can be determined more quickly in the average case with a `some` (which stops iteration as soon as it finds an archived project)
* If the archived tab is highlighted, all you need is the collection of archived projects; the `activeProjects` collection is entirely unused.

Fortunately there is a way to have & eat cake here using the [`partition()` function](https://lodash.com/docs/4.17.11#partition)—by partitioning you can create both `activeProjects` and `archivedProjects` using a single pass over the collection, thus avoiding wasted work even if you’re only using one of them lightly/not at all.

(N.B. this is all a bit angels-on-the-head-of-a-pin level of performance reasoning; with collections this small it’s very unlikely to make a difference one way or another. But I still think, as a rule, it’s worth doing any iteration in the cheapest available way that doesn’t substantially add to code complexity)

src/css/application.css
```diff
@@ -477,6 +486,50 @@ body {
...
+  padding: 0.75rem 2rem;
+  border: 1px solid transparent;
+  border-top-left-radius: 0.25rem;
+  border-top-right-radius: 0.25rem;
+  margin-bottom: -1px;
+  cursor: pointer;
+}
+
+.project-picker__nav-tab-active {
+  border-color: var(--color-chrome) var(--color-chrome) #fafafa;
```
outoftime: Is there no existing gray we can use in place of the ad-hoc `#fafafa`?

src/components/Modal.jsx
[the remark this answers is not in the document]
pwjablonski: Ah that's much simpler!

src/components/ProjectPickerModal.jsx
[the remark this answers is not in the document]
pwjablonski: Yeah! that sounds great!

src/css/application.css
pwjablonski: Yah I am using what was defined on the `.modal__contents` selector. Would it make sense to pull this out into a variable or change the `.modal__contents` background-color to something predefined.

src/reducers/__tests__/ui.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+  it('archived projects', () => {
+    const {projectsFilter} = applyActions(filterProjects('archived'));
+    expect(projectsFilter).toEqual('archived');
+  });
+});
+
+function applyActions(...actions) {
+  return reduce(actions, (state, action) => reducer(state, action), undefined);
+}
```
outoftime: Can we move this into a helpers module, e.g. `src/reducers/__tests__/helpers/index.js`? @jwang1919 is using an identical function in #1701

src/logic/__tests__/projectSuccessfullySavedTest 2.js
```diff
@@ -0,0 +1,17 @@
+import projectSuccessfullySaved from '../projectSuccessfullySaved';
```
outoftime: Seems like we got duplicates of this and some other files? Note the ` 2` in the filename…

src/reducers/__tests__/helpers/index.js
```diff
@@ -0,0 +1,7 @@
+import reduce from 'lodash-es/reduce';
+
+import reducer from '../../ui';
```
outoftime: Since this is intended as a helper for all reducer tests, I don’t think we want to specifically import and use the `ui` one :slightly_smiling_face:

src/reducers/ui.js
```diff
@@ -223,6 +222,14 @@ export default function ui(stateIn, action) {
         false,
       );
 
+    case 'OPEN_PROJECT_PICKER_MODAL':
+      return state.setIn(['isProjectPickerModalOpen'], true);
+
+    case 'CLOSE_PROJECT_PICKER_MODAL':
+      return state
+        .setIn(['isProjectPickerModalOpen'], false)
```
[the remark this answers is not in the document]
outoftime: Looks like you missed this one

src/selectors/getPartitionedProjects.js
```diff
@@ -0,0 +1,9 @@
+import {createSelector} from 'reselect';
+
+import partition from 'lodash-es/partition';
+
+import getAllProjects from './getAllProjects';
+
+export default createSelector([getAllProjects], projects =>
+  partition(projects, {isArchived: false}),
```
outoftime: EXTREMELY NITPICKY ALERT i think you can just pass the string `'isArchived'` here rather than `{isArchived: false}`

src/selectors/isLoginPromptOpen 2.js
```diff
@@ -0,0 +1,3 @@
+export default function isLoginPromptOpen(state) {
```
outoftime: Same here with the duplicate file

src/components/TopBar/ProjectPickerButton.jsx
```diff
@@ -1,19 +1,30 @@
...
+  isUserAuthenticated,
+  projectKeys,
+  shouldShowSavedIndicator,
+  onClick,
+}) {
+  if (
+    !(
+      isUserAuthenticated &&
+      !isEmpty(projectKeys) &&
+      Boolean(currentProjectKey)
```
outoftime: Similar to previous comment, I think it’s more precise to write `!isNil(currentProjectKey)`

outoftime: Also I would probably distribute the negation so the whole thing reads:

```js
if (
  !isUserAuthenticated ||
  isEmpty(projectKeys) ||
  isNil(currentProjectKey)
) {
```

src/reducers/__tests__/ui.test.js
```diff
@@ -423,6 +411,34 @@ test('finish dragging column divider', () => {
...
+
+test('closeProjectPickerModal closes project picker modal', () => {
+  applyActions(openProjectPickerModal());
+  expect(applyActions(closeProjectPickerModal())).toMatchObject({
+    isProjectPickerModalOpen: false,
+  });
+});
+
+describe('filterProjects filters', () => {
+  it('active projects', () => {
```
outoftime: We typically use `test` instead of `it` (especially if the test description isn’t forming a sentence after the “it”)

src/reducers/__tests__/ui.test.js
```diff
@@ -423,6 +411,34 @@ test('finish dragging column divider', () => {
...
+    applyActions(filterProjects('archived'));
```
outoftime: This line isn’t doing anything

src/reducers/__tests__/ui.test.js
```diff
@@ -423,6 +411,34 @@ test('finish dragging column divider', () => {
...
 });
 
+test('openProjectPickerModal opens project picker modal', () => {
+  expect(applyActions(openProjectPickerModal())).toMatchObject({
+    isProjectPickerModalOpen: true,
+  });
+});
+
+test('closeProjectPickerModal closes project picker modal', () => {
+  applyActions(openProjectPickerModal());
```
outoftime: This line isn’t doing anything

PR #1748, opened 2019-08-29
Convert validation tests to jest

wylieconlon:
This adds a few shims for testing in jest, which does not have a full browser DOM. Specifically, it does not have the ability to implement `createRange`, which we use to do reporting of the HTML line number of HTML validations. Instead, we are shimming it by counting the number of newlines in text nodes in the test environment, which gives the same results in this limited scope.


src/validations/__tests__/javascript.test.js
```diff
@@ -0,0 +1,84 @@
...
+};
+
+const analyzerWithExternalScript = {
+  enabledLibraries: [],
+  containsExternalScript: true,
+};
+
+describe('javascript validation', () => {
+  test('invalid LHS error followed by comment', () => {
+    validationTest(
```
outoftime: Looks like the tests here are still calling the async `validationTest` method without awaiting or returning their result?

src/validations/__tests__/runRules.test.js
```diff
@@ -0,0 +1,177 @@
...
+            *done() {},
+          },
+        ],
+        '<img>',
+      );
+    }).not.toThrow();
+  });
+
+  test('openTag skips self-closing non-void tag', async () => {
+    expect(async () => {
```
outoftime: Similar situation here, I don’t think `expect(…).not.toThrow()` would work out of the box when passed an async function—it looks like you can use [`.rejects`](https://jestjs.io/docs/en/expect#rejects), and return/await the result of the `expect`, to do what you want here.

src/validations/__tests__/validationHelper.js
```diff
@@ -0,0 +1,20 @@
...
+) {
+  try {
+    const errors = await validate(input);
+    expect(
+      map(orderBy(errors, ['reason', 'row']), error =>
+        pick(error, ['reason', 'row', 'payload']),
+      ),
+    ).toEqual(orderBy(expectedErrors, ['reason', 'row']));
+  } catch (e) {
+    throw e;
```
outoftime: Do we need this `try`/`catch`?

src/validations/__tests__/javascript.test.js
wylieconlon: Fixed

src/validations/__tests__/runRules.test.js
wylieconlon: Had to read the original runRules code for this. As long as the `await runRules` finishes successfully, the test is valid because it's testing that a tag was skipped.

src/validations/__tests__/validationHelper.js
wylieconlon: No

PR #1834, opened 2019-10-02
Clarify "try" suggestion language

thundernixon:
Teaching in a code nation classroom today, a beginner student say the `Try: href="myvalue"` suggestion, and went to paste that exact string into his `img` element. This was an understandable attempt to fix the issue, and I had to correct him that it was showing the format to follow, rather than something specific to add.

I think we might make it more clear if we change from this:

> **attribute-value:** Every attribute needs a value. Try: `myattribute="myvalue"`

...to this:

> **attribute-value:** Every attribute needs a value. Try coding it like this: `myattribute="myvalue"`

I'm not sure that my exact phrasing is perfect ... but I think a small change here might help students in the future (or it's possible that it _is_ the best version right now). I'm happy to discuss/adjust/close my PR. :)


locales/en/translation.json
```diff
@@ -227,7 +227,7 @@
       "use-new-object": "If you are making a new object, be sure to use it or store it in a variable",
       "unnecessary-semicolon": "You don't need a semicolon `;` on this line",
       "missing-semicolon": "You need a semicolon `;` at the end of this line",
-      "missing-parentheses": "There should be parentheses here.\nTry: `new {{-object}}()`",
+      "missing-parentheses": "There should be parentheses here.\nTry coding it like this: `new {{-object}}()`",
```
outoftime: I think in this case, we have enough information from the validation error that we can give a suggestion that the students can use as-is?

locales/en/translation.json
```diff
@@ -237,7 +237,7 @@
         "different": "You should use `!==` instead of `!=`",
         "custom-case": "There should be a `{{-goodOperator}}` where you wrote `{{-badOperator}}`"
       },
-      "declare-variable": "You haven't declared `{{-variable}}`.\nTry: `var {{-variable}};`",
+      "declare-variable": "You haven't declared `{{-variable}}`.\nTry coding it like this: `var {{-variable}};`",
```
outoftime: Same here—we know the variable name, so in this case the student can take the suggestion literally

locales/en/translation.json
```diff
@@ -195,9 +195,9 @@
     "css": {
       "missing-opening-curly": "You should have a `{` at the end of this line.",
       "missing-closing-curly": "You have a starting `{` but no ending } to go with it.",
-      "property-missing-colon": "Put a colon (`:`) between the property and the value.\nTry: `color: red;`",
-      "selector-missing": "Start every block of CSS with a selector, such as an element name or class name.\nTry: `p {\n  color: red;\n}`",
-      "block-expected": "Start a block using `{` after your selector.\nTry: `{{-error}} {`",
```
outoftime: Same here (just the last one)—I think the suggestion can be taken literally

thundernixon:
Great comments, @outoftime! I admit, I didn't fully understand several of these error messages adjusting the `Try` in them. I'm definitely happy to leave out the extra words where suggestions can be used literally.

As for `coding` vs `writing`, I think either can work. I considered that, and also `Try formatting it like:`. I slightly do prefer either `coding` or `formatting`, because we are telling the student that they must type something very specifically. In my head, `writing` is more associated with natural language, like what I'm writing right now. That's just my thought, though, and it's hard for me to know what a student might find most clear.

thundernixon:
Updated! 

Maybe for the coding/writing word choice, we could get one more person to weigh in on what they think is most natural?

fiddlerpianist:
I think `writing` is more natural than `coding`

thundernixon:
Cool, thanks for weighing in! Updated to "writing."

I think maybe this is good to merge, then? Please let me know if you see anything I missed.

PR #1856, opened 2019-10-12
Allow tel: and mailto: patterns in href attributes

### LINKED ISSUE #1827: Allow tel: and mailto: patterns in href attributes
Currently popcode gives an error if your url doesn't start with http:// or https:// . I'd also like to allow for tel: and mailto: so we can give the students some more interesting bonus challenges to try with links.

joshling1919:
Closes #1827 

1. Edited the `href-style` error in the `errorMap` to make exceptions for `mailto:` and `tel:`
2. Updated the `href-style` error message in `translation.json` to more accurately reflect the requirements of `htmllint`.


src/validations/html/htmllint.js
```diff
@@ -46,7 +46,17 @@ const errorMap = {
 
   E008: () => ({reason: 'doctype'}),
 
-  E009: () => ({reason: 'href-style'}),
+  E009: (error, source) => {
+    const lines = source.split('\n');
+    const hrefValExpr = /href="([^"]*)/u;
```
outoftime: I don’t think this expression will match in all cases where we might see this error—for instance, the following all trigger an `E009` along with some other error (except in the first example):

```html
<a href='tel:1231231234'>hello</a>
<a href=tel:1231231234>hello</a>
<a HREF="tel:1231231234">hello</a>
<a href = "tel:1231231234">hello</a>
```

It should be fairly straightforward to generalize the expression to match those cases, but I would suggest also, in the spirit of regular expressions being poor substitutes for parsers, account for the possibility that the regexp won’t match (in which case I would argue it’s fine to have a false positive for a `tel:` link).

Alternatively, you could try feeding the relevant part of the source into an HTML fragment parser, though that’s probably overkill.

src/validations/html/htmllint.js
```diff
@@ -46,7 +46,17 @@ const errorMap = {
...
+    const hrefVal = hrefValExpr.exec(
+      lines[error.line - 1].slice(error.column),
+    )[1];
+    if (hrefVal.indexOf('mailto:') > -1 || hrefVal.indexOf('tel:') > -1) {
```
outoftime: Nit, but [`startsWith()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) is a more expressive way to do these checks

joshling1919:
@outoftime Thanks for the detailed feedback! I know you mentioned using an HTML parser might be a bit overkill, but it ended up being one of the cleaner ways to handle all the different cases. Let me know if you'd prefer the regex route.

outoftime:
> @outoftime Thanks for the detailed feedback! I know you mentioned using an HTML parser might be a bit overkill, but it ended up being one of the cleaner ways to handle all the different cases. Let me know if you'd prefer the regex route.

This is good!! “Overkill” was probably the wrong word—more that the regex solution (with safety for non-match) was probably practically good enough, but I definitely think what you’ve done here is the better approach.

PR #1930, opened 2019-10-30
fix: handle duplicate js variable declaration

erikwlarsen:
This fix will show a helpful error message for multiple declarations using ES6 syntax (let/const). Doesn't do anything for multiple var declarations with the same name.

This is my first PR to Popcode, should I add tests? If so could you tell me where I should be looking for tests in a similar vein? Thanks!

### DIFF
--- a/src/validations/__tests__/javascript.test.js
+++ b/src/validations/__tests__/javascript.test.js
@@ -72,6 +72,18 @@ describe('javascript validation', () => {
       partialRight(javascript, analyzer),
     ));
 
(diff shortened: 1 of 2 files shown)


outoftime:
Hey @erikwlarsen, any chance you’d have a a little time to add some test cases to this PR? Let me know either way, I’ll take it over if not 🙂

erikwlarsen:
Hey, totally forgot about this, my apologies! Should be able to throw some
tests in there tomorrow.

Issue #1934, opened 2019-11-01
Migrate from redux-saga to redux-logic


outoftime:
`redux-saga` is a very powerful tool but it also introduces a completely different approach to async programming than the one we use in all other JavaScript code, and requires anyone who uses it ramps up on a whole new vocabulary of effects. For new contributors who come in without experience with `redux-saga`, it bends the learning curve dauntingly.

[`redux-logic`](https://github.com/jeffbski/redux-logic), while undeniably less popular than `redux-saga`, accomplishes the same basic goals using a much simpler approach. Logics are written using regular `async`/`await` code; if we need to deal with more complex scenarios, we can fall back to the underlying RxJS, which has a similar learning curve to `redux-saga` but is a much more general-purpose tool. This, anyway, should be rare.

It should be noted that `redux-saga` _can_ do things that `redux-logic` can’t, in particular around cancellation of in-flight asynchronous actions. However dealing with cancellation correctly is quite tricky, and anyway we haven’t come across any situations where we need to.

### Code structure

While our current saga code is organized into modules of related sagas, which are then consolidated into a single default export, logics should be one per module.

Note also that our sagas are often named after the action that triggers them; however, this presumes a 1:1 mapping between actions and sagas (or actions and logics) that does not actually obtain. So, let’s name logics after what they _do_, not what action(s) they consume.

### Testing

As part of this transition, we should write the tests for the new logics in Jest, rather than Karma where the Saga tests live. Thus this should help us move toward #1443 as well.

# Sagas to migrate

### `sagas/user.js`

- [x] `linkGithubIdentity`
- [x] `logout`
- [x] `startAccountMigration`
- [x] `unlinkGithubCredentials`

### `sagas/errors.js`

- [ ] `validateCurrentProject`
- [ ] `updateProjectSource`
- [ ] `toggleLibrary` (note that this probably does not need to be its own logic—it should just be subsumed into `validateCurrentProject`)

### `sagas/ui.js`

- [ ] `userDoneTyping`
- [ ] `popOutProject`
- [ ] `exportProject`
- [ ] `projectSuccessfullySaved`

### `sagas/clients.js`

- [ ] `createSnapshot`
- [ ] `exportProject`
- [ ] `applicationLoaded`

### `sagas/compiledProjects.js`

- [ ] `validatedSource`

### `sagas/projects.js`

- [ ] `applicationLoaded`
- [ ] `createProject`
- [ ] `changeCurrentProject`
- [ ] `projectExported`
[...]

PR #1954, opened 2019-11-16
Redux logic errors

joshling1919:
Will need to add tests, but had some quick questions that I was hoping to get some feedback on.

1. For testing redux-logic, I was thinking about using https://github.com/jeffbski/redux-logic-test. Specifically, I would dispatch actions and then verify that the logic ran properly and resulted in the correct state. Not sure if this approach would assert on too many pieces outside of redux-logic. For example, it would depend on reducers being configured in a specific way. Would be interested in your thoughts!

2. I'm having a little bit of trouble wrapping my head around the way the original errors saga was set up, to where all of these would be running in parallel, and if there was already validation happening for a language, it would then cancel the validation call for that language. In https://github.com/popcodeorg/popcode/issues/1934, you mentioned that redux-logic can't cancel in-flight asynchronous actions. Is this one of of the cases you were talking about, or was that something else?

3. Any thoughts on the organization of the redux-logic `errors`? I saw that @jwang1919 had split up to have one logic per file, but in terms of the `errors` saga, both `validateCurrentProject` and `updateProjectSource` called on the same `validateSource` function that will now not be its own `logic`, which is why I decided to organize it all into one file.

Thanks!


src/logic/errors.js
```diff
@@ -0,0 +1,77 @@
...
+          dispatch,
+        ),
+      );
+    }
+    await Promise.all(validatePromises);
+    done();
+  },
+});
+
+const updateProjectSource = createLogic({
```
outoftime: I would actually probably call this `validateProjectOnChange` or something like that—generally I’d like to follow the principle of naming logics after what they actually do, rather than the action(s) they consume. (N.B. the sagas tend to follow the former naming convention, which is my fault!)

src/logic/errors.js
```diff
@@ -0,0 +1,77 @@
...
+
+const validateSource = async (
+  {language, source, projectAttributes},
+  dispatch,
+) => {
+  const validations = await importValidations();
+  const validate = validations[language];
+  const validationErrors = await validate(source, projectAttributes);
+  await dispatch(validatedSource(language, validationErrors));
+};
```
outoftime: Rather than putting multiple logics in one module, let’s move this into its own module, probably inside a `src/logic/helpers` package that we can use in the general case when we need to share code between different logics. Then we can put each of the logics defined here in its own module.

src/logic/errors.js
```diff
@@ -0,0 +1,77 @@
+import {createLogic} from 'redux-logic';
+import Analyzer from '../analyzers';
+import {getCurrentProject} from '../selectors';
+import {validatedSource} from '../actions/errors';
+import retryingFailedImports from '../util/retryingFailedImports';
+
+export async function importValidations() {
```
outoftime: Does this need to be exported?

joshling1919:
@outoftime so sorry, this took longer than expected but I've implemented most of your feedback! The only thing I still wasn't sure about was the cancellation/ignoring the validation of a source if there is already one in flight. I tried to experiment with and research on the topic a little more but I think I'm still a little confused on what the current problem is vs what the end goal should be when it comes to these cancellations. Is it that these validations are firing way more than necessary? What would be the best way to test that we've achieved the correct cancellation logic? Thanks for your help!

src/logic/helpers/validateSource.js
```diff
@@ -0,0 +1,18 @@
...
+function importValidations() {
+  return retryingFailedImports(() =>
+    import(
+      /* webpackChunkName: "mainAsync" */
+      '../../validations'
+    ),
+  );
+}
+
+export default async ({language, source, projectAttributes}, dispatch) => {
```
outoftime: _Extreme_ nit / not a blocker, but all else being equal I’d make this a regular named function rather than an arrow function (since it’s at the top level of the module / not being inlined in a larger expression).

src/logic/validateCurrentProject.js
```diff
@@ -0,0 +1,32 @@
...
+    const validatePromises = [];
+    for (const language of Reflect.ownKeys(currentProject.sources)) {
+      const source = currentProject.sources[language];
+      validatePromises.push(
+        validateSource(
+          {language, source, projectAttributes: analyzer},
+          dispatch,
+        ),
+      );
+    }
```
outoftime: Nit: a `map` would be a more idiomatic way to construct this array; I would probably use lodash’s map directly over `currentProject.sources` since it will iterate over object entries.

src/logic/helpers/validateSource.js
```diff
@@ -0,0 +1,18 @@
...
+  const validations = await importValidations();
+  const validate = validations[language];
+  const validationErrors = await validate(source, projectAttributes);
+  await dispatch(validatedSource(language, validationErrors));
```
outoftime: How come we’re awaiting the `dispatch` here? `dispatch` should be a void function; I wouldn’t expect it to return a promise…

src/logic/__tests__/validateCurrentProject.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+import Analyzer from '../../analyzers';
+import validateSource from '../helpers/validateSource';
+import {getCurrentProject} from '../../selectors';
+
+const mockHtmlSource = '<html></html>';
+const mockCssSource = 'div {}';
+
+jest.mock('../../analyzers');
+jest.mock('../helpers/validateSource');
```
outoftime: So this I think may be going a little further than we want with mocking—in particular the fact that `validateCurrentProject` delegates some of its logic to the `validateSource` helper is essentially an implementation detail of `validateCurrentProject`; if in the future we decided to pull all of that logic directly into `validateCurrentProject`, I don’t think we would consider that breaking from the standpoint of the unit tests.

So my proposal would be to just test the behavior of the `validateCurrentProject` logic here, including any behavior that happens to be implemented in `validateSource`; and remove the unit test for the `validateSource` helper.

src/logic/__tests__/validateCurrentProject.test.js
```diff
@@ -0,0 +1,56 @@
...
+jest.mock('../../selectors', () => ({
+  getCurrentProject: jest.fn(() => ({
+    sources: {
+      html: mockHtmlSource,
+      css: mockCssSource,
+    },
+  })),
+}));
```
outoftime: To be honest this also intuitively is probably not something I would have mocked if I were writing this test, although I think the argument is _much_ less clear-cut than with the helper module commented on above. I could truly go either way on this, but wanted to flag it as something to think about, particularly in terms of “which decision is going to make this test more robust?”, with “robust” meaning “sensitive to breaking changes in the code under test, and insensitive to non-breaking changes in the code under test”.

src/logic/__tests__/validateCurrentProject.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+const dummyState = {};
+
+test('should validate current project', async () => {
+  const dispatch = jest.fn();
+  const done = jest.fn();
+  await validateCurrentProject.process(
+    {
+      getState: () => dummyState,
```
outoftime: I think `eslint-plugin-lodash` (introduced after you created this branch) is going to tell you to use `constant` here.

src/logic/__tests__/validateCurrentProject.test.js
```diff
@@ -0,0 +1,56 @@
...
+    },
+    dispatch,
+    done,
+  );
+  expect(getCurrentProject).toHaveBeenCalledWith(dummyState);
+  const analyzerPayload = getCurrentProject(dummyState);
+  expect(Analyzer).toHaveBeenCalledWith(analyzerPayload);
+  const analyzer = new Analyzer(analyzerPayload);
+  expect(validateSource.mock.calls).toEqual([
```
outoftime: I would probably use [`expect.arrayContaining`](https://jestjs.io/docs/en/expect#expectarraycontainingarray) here since we’re not sensitive to the order of the calls.

src/logic/__tests__/validateCurrentProject.test.js
```diff
@@ -0,0 +1,56 @@
...
+    },
+  })),
+}));
+
+const dummyState = {};
+
+test('should validate current project', async () => {
+  const dispatch = jest.fn();
+  const done = jest.fn();
+  await validateCurrentProject.process(
```
outoftime: So, this is a bit of a pain in the ass, but `await`ing the call to `process` is actually testing implementation details of that method; from the standpoint of `redux-logic`, `process` does not have to be an async (or Promise-returning) function; it just needs to call `done()` when it’s done. I think the best way to test the relevant interface is something like:

```js
await new Promise((resolve) => {
  validateCurrentProject.process(
    {getState: constant(dummyState)},
    dispatch,
    resolve,
  )
});
```

FWIW I was looking through other logic tests for examples and am realizing I totally failed to catch a bunch of other cases of what you did here, so I’ll need to go back and fix those—sorry!!

outoftime: Hey so! I dug into fixing this issue for the existing tests, and ended up just writing a test helper that actually runs logics through the `redux-logic` middleware, so the tests can focus on what the logics _do_ rather than trying to mimic the implementation of `redux-logic`. Check out #2026 for all the sordid details!

joshling1919:
hi! sorry just got back from winter break vacation. Thanks so much for the detailed and helpful feedback. Will take another pass and implement your feedback by end of week!

joshling1919:
@outoftime tried my best to implement your feedback. Please let me know if there was anything that I may have misunderstood!

Also, I was thinking more the case where there might be a race condition between `validateCurrentProject` and `validateProjectOnChange`. What are your thoughts on just merging the two into something that looks like:

```js
export default createLogic({
  type: [
    'CHANGE_CURRENT_PROJECT',
    'GIST_IMPORTED',
    'SNAPSHOT_IMPORTED',
    'PROJECT_RESTORED_FROM_LAST_SESSION',
    'TOGGLE_LIBRARY',
    'UPDATE_PROJECT_SOURCE',
  ],
  latest: true,
  async process(
    {
      getState,
      action: {
        payload: {language, newValue},
      },
    },
    dispatch,
    done,
  ) {
    const state = getState();
    const currentProject = getCurrentProject(state);
    const analyzer = new Analyzer(currentProject);

    if (newValue) {
      await validateSource(
        {language, source: newValue, projectAttributes: analyzer},
        dispatch,
        getState,
    } else {
      const validatePromises = map(
        Reflect.ownKeys(currentProject.sources),
        language =>
          validateSource(
            {
              language,
              source: currentProject.sources[language],
              projectAttributes: analyzer,
            },
            dispatch,
            getState,
          ),
      );

      await Promise.all(validatePromises);
    }

    done();
  },
});
```

Not sure how you feel about throwing the conditional on there to prevent unnecessary validations, but I think this might address the race condition.

outoftime:
Right on! Will take a look as soon as I can. Looks like the build failed so that might be something to check out in the meantime 🙂

joshling1919:
@outoftime implemented your latest feedback!

I'm not sure if I implemented your vision of a multiplexor `process` function correctly though. When I initially tried to make it into two functions that the process function would delegate to, I found that there were too many arguments being passed around between functions and overall it felt like there was too much going on.

Let me know if the way I went with is okay. I'm not sure if I ended up making it less verbose at the expense of readability.

src/logic/__tests__/validateProject.test.js
```diff
@@ -0,0 +1,80 @@
...
+test('should validate current project', async () => {
+  const testLogic = makeTestLogic(validateProject);
+
+  [
+    changeCurrentProjectAction,
+    gistImportedAction,
+    snapshotImportedAction,
+    projectRestoredFromLastSessionAction,
+    toggleLibraryAction,
+  ].forEach(async action => {
```
outoftime: Having an un-awaited `async` function as the iterator is going to prevent the assertions from being executed before the test completes—I would recommend iterating over the actions _outside_ the `test` declaration, e.g.

```js
for (const action of [
  changeCurrentProjectAction,
  gistImportedAction,
  // ...
]) {
  test(`validates current project on ${action}`, async () => {
    // …
  }
}
```

src/logic/validateProject.js
```diff
@@ -0,0 +1,73 @@
...
+    done,
+  ) {
+    const state = getState();
+    const currentProject = getCurrentProject(state);
+    const analyzer = new Analyzer(currentProject);
+
+    const sources =
+      type === 'UPDATE_PROJECT_SOURCE'
+        ? {[language]: newValue}
+        : currentProject.sources;
```
outoftime: So what I was thinking in terms of a multiplexer would be something like:

```js
if (type === 'UPDATE_PROJECT_SOURCE') {
  await validateSource({language, source: newValue, projectAttributes: analyzer}, dispatch);
} else {
  await validateSources(currentProject.sources, analyzer, dispatch)
}
```

Not a big deal either way, but the above approach saves us an unnecessary hop through `validateSources`.

(Also, getting a bit nitpicky, I would probably try to be a bit more consistent about positional arguments vs. args object in the two function headers)

src/logic/__tests__/validateCurrentProject.test.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+  const testLogic = makeTestLogic(validateCurrentProject);
+
+  [
+    changeCurrentProjectAction,
+    gistImportedAction,
+    snapshotImportedAction,
+    projectRestoredFromLastSessionAction,
+    toggleLibraryAction,
+  ].forEach(async action => {
```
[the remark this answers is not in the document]
joshling1919: sorry, totally missed this comment

src/logic/__tests__/validateProject.test.js
joshling1919: makes sense! will fix!

src/logic/validateProject.js
```diff
@@ -0,0 +1,79 @@
...
+
+async function validateSource({language, source, projectAttributes}, dispatch) {
+  const validations = await importValidations();
+  const validate = validations[language];
+  const validationErrors = await validate(source, projectAttributes);
+  dispatch(validatedSource(language, validationErrors));
+}
+
+async function validateSources({sources, projectAttributes}, dispatch) {
+  const validatePromises = map(Reflect.ownKeys(sources), language =>
```
outoftime: Definitely not going to block on this, but a minor improvement here would be to map over `sources` itself, which would give you `(source, language)` as the arguments to the iterator.

PR #1977, opened 2019-12-02
Prefer optional chaining/nullish coalescing

outoftime:
Replaces most instances of lodash `get()` with [optional chaining](https://tc39.es/proposal-optional-chaining/) and [nullish coalescing](https://tc39.es/proposal-nullish-coalescing/).


outoftime:
Taking a look at what Babel outputs for a typical optional chain, I don’t think this is a good idea—there’s potential to substantially increase bundle size. Better to just keep using `_.get` and friends.

Issue #1685, opened 2019-04-25
Less-than sign is escaped in error message


frewsxcv:
https://popcode.org/?snapshot=cbe41493-ab6d-47ea-84e6-f5024248638a

<img width="512" alt="Screen Shot 2019-04-24 at 9 57 52 PM" src="https://user-images.githubusercontent.com/416575/56704687-1bfe4e00-66dc-11e9-8758-e62fb41967ba.png">

TSamee:
Found the immediate cause. The `text` attribute on errors is assigned in src/validations/Validator.js, lines 49-57:

```
  _convertErrorToAnnotation(rawError) {
    //...
    const error = this.mapError(rawError)
    //...
    return assign({}, location, error, {
      text: remark()
        .use(stripMarkdown)
        .processSync(message)
        .toString(),
      raw: message,
      type: 'error',
    });
  }
```
During the call to `processSync(message)`, remark parses the error message and compiles the resulting syntax tree into a string. In the compilation phase, `remark-stringify` escapes punctuation. The function escape() is defined in remark-stringify/lib/escape.js, which exports a factory function that returns escape().

escape() iterates over the value of a node and its siblings character by character. It contains a hardcoded check for the less-than character on line 94, and will replace them with '&lt;' according to a mapping defined in the module. So '<' will be converted to '&lt;' for any error whose message is processed through remark (which I think is all of them).

There's the how. Still working on the why, as I don't know why the check for '<' specifically is there, or whether the error message syntax tree is routing correcting through the compiler.

outoftime:
@TSamee good investigation! I think what you’re seeing here may be correct behavior on the part of Remark, and an incorrect assumption on Popcode’s part. In particular, the goal of the code in Validator is to convert the Markdown messages into plain text. However, taking a swing through the docs, it looks like what the code is actually doing is converting it to HTML, but without any formatting (the latter being what `stripMarkdown` is doing). Since it’s still HTML, the `<` _should_ still be escaped.

What we want, though, is plain text. Seems like we have a couple of options:

* Find something within the `remark` ecosystem that outputs actual plain text. I did see one plugin but it hasn’t been touched in four years and does not seem to have any adoption
* Just do a second step that converts Remark’s HTML output to plain text. No opinion on the right way to do this but I am guessing there are good options

PR #2105, opened 2020-01-22
Validator decodes HTML entities in error text

TSamee:
Attempt at a fix for #1685, adds `he` as a dependency.

Found that remark as called by Validator will return HTML instead of plain text. Right now this only produces error messages with all '<' characters encoded as '\&lt;', but it may affect future error messages with colons, ampersands, pipes and tildes as well.

I couldn't find a suitable addition to remark that would output plain text or decode the output of the last compiler step, so I opted to decode the result of the call to remark in Validator (line 52). 

Using the decode function from `he` seemed like the cleanest way to do this, but I know that adding a dependency might be poor form, so I'm open to defining a decode() function in Validator instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -217,6 +217,7 @@
     "first-input-delay": "0.1.3",
     "github-api": "3.3.0",
     "hast-util-sanitize": "2.0.1",
+    "he": "^1.2.0",
     "highlight.js": "9.17.1",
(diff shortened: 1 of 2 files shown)


TSamee:
Seems I didn't install the package correctly. Taking this down for now.

PR #2106, opened 2020-01-23
Validator decodes HTML entities in error text

### LINKED ISSUE #1685: Less-than sign is escaped in error message

TSamee:
Closes #1685 
Adds `he` as a dependency.

Found that remark as called by Validator will return HTML instead of plain text. Right now this only produces error messages with all '<' characters encoded as '\&lt;', but it may affect future error messages with colons, ampersands, pipes and tildes as well.

I couldn't find a suitable addition to remark that would output plain text or decode the output of the last compiler step, so I opted to decode the result of the call to remark in Validator (line 52). 

Using the decode function from `he` seemed like the cleanest way to do this, but I know that adding a dependency might be poor form, so I'm open to different approaches.


outoftime:
> adding a dependency might be poor form

I’m not terribly worried about this, btw; a quick glance at `package.json` should confirm that we are not dependency minimalists around here. I _do_ worry about bundle size but mostly for the critical path chunks (the code that’s loaded synchronously and is needed to bootstrap the application). The validations package is already loaded asynchronously, so adding a dependency to it won’t have any impact on the critical load path.

TSamee:
Just saw this, done. I must have applied my global Babel installation, or something like that.

Issue #804, opened 2017-04-29
User Feedback


outoftime:
Submitter: Z
Email: zwilkinson@westinghousehs.org
I can’t use the browser’s Camera API from inside the preview window, because the preview is a `data:` URI that is not considered secure.

outoftime:
Fixed by #2107

PR #2107, opened 2020-01-24
Secure preview frame

pwjablonski:
Hey @outoftime!

Here's my first pass at #804 

- [x] creates static `preview.html` file
- [x] loads `previewSupport` into `preview.html` via webpack
- [x] Replaces `srcDoc` with `src` reference to `preview.html`
- [x] Sends message with `source` from `previewFrame` component to  the iframe
- [x] Recieves source from the message and write content to iframe document
- [ ] ~fixes iframe settings to allow for camera/audio access~

Appreciate your feedback especially around the chunking in webpack


src/html/preview.html
```diff
@@ -0,0 +1,5 @@
+<!DOCTYPE html>
+<html>
+  <head> </head>
```
pwjablonski: Delete extra space

src/html/preview.html
```diff
@@ -0,0 +1,5 @@
+<!DOCTYPE html>
```
pwjablonski: Does it make sense to include this as the default html given it will be written over immediately?

webpack.config.js
```diff
@@ -361,4 +304,56 @@ module.exports = (env = process.env.NODE_ENV || 'development') => {
     },
     devtool,
   };
+
+  const mainConfig = {...baseConfig};
```
pwjablonski: This seems a bit hacky - open to suggestions on how to better make two configs from the same base.

src/components/PreviewFrame.jsx
```diff
@@ -39,7 +35,7 @@ class PreviewFrame extends React.Component {
           name={`preview-frame-${nextId++}`}
           ref={this._attachToFrame}
           sandbox={sandboxOptions}
```
pwjablonski: I could be wrong but it seems that having the sandbox setting on at all prevents us from accessing the camera/audio regardless of if we load the HTML from the same origin.

webpack.config.js
outoftime: This is totally fine/quite idiomatic! The only thing I would change is just constructing each variant of the config in a single literal, e.g.:

```js
const mainConfig = {
  ...baseConfig,
  name: 'main',
  entry: isTest
    ? undefined
    : {
        inline: 'first-input-delay',
        main: './application.js',
      },
  //...
};
```

webpack.config.js
```diff
@@ -361,4 +304,56 @@ module.exports = (env = process.env.NODE_ENV || 'development') => {
...
+    previewConfig.plugins.push(
+      new HtmlWebpackPlugin({
+        filename: 'preview.html',
+        template: './html/preview.html',
+        chunksSortMode: 'dependency',
+      }),
+    );
+  }
+
+  return [previewConfig, mainConfig];
```
outoftime: Out of curiosity, can you tell if the main and preview bundles are still sharing certain chunks?

src/previewSupport/handleSourceUpdates.js
```diff
@@ -0,0 +1,8 @@
+import channel from './channel';
+
+export default function handleSourceUpdates() {
+  channel.bind('updateSrc', (_trans, source) => {
+    document.open();
+    document.write(source);
```
outoftime: I would throw a [`document.close()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/close) in here as well.

src/html/preview.html
outoftime: Yes—the page that gets loaded into the preview frame should still be valid HTML (and the doctype is particularly important). You _can_ skip the `<head>` and `<body>` tags, though.

src/components/PreviewFrame.jsx
outoftime: OK! Per Slack conversation let’s merge this anyway and then keep hammering on the camera API.

src/components/PreviewFrame.jsx
```diff
@@ -93,6 +89,13 @@ class PreviewFrame extends React.Component {
     });
   }
 
+  _updateSource(source) {
```
outoftime: Not a huge deal but I think it’s a bit strange to define this as an instance method—would make more sense to me to just have the logic inlined in the `onReady` below.

webpack.config.js
pwjablonski: I'm not quite sure but It seems that they are not sharing any chunks
preview.html has...
   Entrypoint preview = vendors~preview.js preview.js
index.html has...
    Entrypoint inline = inline.js
    Entrypoint main = vendors~main.js main.js
    a bunch of other chunks

src/components/PreviewFrame.jsx
```diff
@@ -190,8 +186,12 @@ class PreviewFrame extends React.Component {
     this._channel = Channel.build({
       window: frame.contentWindow,
       origin: '*',
```
outoftime: It occurs to me, I think we can set this to `window.location.origin` since the frame now has a real origin?

webpack.config.js
outoftime: Oh well, would be ideal but not a big deal either way. Definitely feels like exporting separate webpack configs is the overall best option considering the limited affordances in `HtmlWebpackPlugin`.

src/components/PreviewFrame.jsx
outoftime: Hmm never mind I just poked around with this and it does not work straightforwardly…

PR #2122, opened 2020-01-31
Revert "Redux logic errors"

### LINKED ISSUE #2121: Cannot read property 'toString' of undefined
## Error in Popcode

**TypeError** in **/**
Cannot read property 'toString' of undefined

[View on Bugsnag](https://app.bugsnag.com/popcode/popcode/errors/5e3425f52cddbb001afdcfd8?event_id=5e3432a80055e473fa0f0000&i=gh&m=ci)

## Stacktrace

    validations/Validator.js:26 - t.value
    validations/javascript/esprima.js:102 - t.value
    validations/Validator.js:33 - t.value
    validations/Validator.js:21 - t.<anonymous>

[View full stacktrace](https://app.bugsnag.com/popcode/popcode/errors/5e3425f52cddbb001afdcfd8?event_id=5e3432a80055e473fa0f0000&i=gh&m=ci)

*Created by Matthew Brown via Bugsnag*

outoftime:
Reverts popcodeorg/popcode#1954

Fixes #2121


outoftime:
FWIW I donʼt think #1954 introduced the root cause of the regression—I suspect that an existing problem in the validation code path had more severe user-facing consequences because of differences in error handling between redux-saga and redux-logic. So, the followup here will be to both fix the specific uncaught error (calling `toString()` on a possibly-nullish value) as well as adding more graceful uncaught error handling to our redux-logic deployment.

PR #2148, opened 2020-02-10
Upgrade all transitive dependencies to latest

outoftime:
Upgrades all transitive dependencies to latest by deleting and regenerating `yarn.lock`. This is safe for direct dependencies because all versions are pinned in `package.json`.

Also fixes duplicate entry for `highlight.js` in `package.json` while we’re in there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -226,7 +226,6 @@
     "github-api": "3.3.0",
     "hast-util-sanitize": "2.0.2",
     "he": "1.2.0",
-    "highlight.js": "9.17.1",
     "highlight.js": "9.18.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -411,5 +410,8 @@
     "webpack-dev-middleware": "3.7.2",
     "webpack-visualizer-plugin": "0.1.11",
     "yarn-deduplicate": "1.1.1"
+  },
+  "resolutions": {
+    "inline-style-prefixer/**/browserslist": "<=4.8.3"
```
outoftime: See https://github.com/robinweser/inline-style-prefixer/issues/191

PR #2150, opened 2020-02-11
Redux logic errors

joshling1919:
Reverts #2122 

@outoftime It looks like `undefined of toString` in Validator didn't end up being the reason that the code preview failed. That should have been resolved by switching the order of the `saga` and `logic` middleware. Do you know of any scenarios where saga might be listening for actions that from `redux-logic`?

I also tried to recreate the specific `undefined of toString` error, and unfortunately was not able to. According to the stacktrace it came from the `esprisma` validations. I've wrapped the validations inside of a try catch block, and configured it so that it would send bugsnag what the current sources are at the time of an error. Hopefully this will help us hone in on the exact issue. I verified that errors that originate from a validation won't break the app, and validations for other languages will still display. For example, if the `esprisma` validation throws an error, the HTML and CSS errors will still show.


joshling1919:
@outoftime thanks for clarifying the root cause of the issue!

PR #2187, opened 2020-03-13
Fix missing divider when html collapsed

### LINKED ISSUE #1676: Divider missing when html collapsed
The draggable divider disappears when the both the CSS and JS editors are open and the HTML editor is collapsed.

Screen Shot 2019-04-04 at 1 09 46 PM

ianplunkett:
This PR fixes #1676.  The change adds the optional index argument (as idx) of the forEach function that loops through visibleLanguages and uses that variable in the conditional to check whether to add a `<DraggableCore>` component.

Screen Shot 2020-03-12 at 9 38 30 PM


src/components/EditorsColumn.jsx
```diff
@@ -39,7 +39,7 @@ export default function EditorsColumn({
 
   const editors = [];
 
-  visibleLanguages.forEach(({language, index}) => {
+  visibleLanguages.forEach(({language, index}, idx) => {
```
outoftime: Can we rename the `index` and `idx` variables to make it clearer what each of them is indexing?

ianplunkett: @outoftime thanks for the quick turnaround. Before I put a new change up, what do you think about renaming like this?

`idx => visibleLanguagesIndex`

`index => languageIndex`

outoftime: Yep sounds good to me!

ianplunkett: @outoftime I've update the index names.  I also ran prettier, which should account for all the diffs in the file besides the renaming.

PR #2236, opened 2020-04-16
Project picker tile sizes

### LINKED ISSUE #2102: Untitled Project in Project Picker
If a project does not have a title (is an empty string) the Project Preview item is thinner. Consider either a default display name if no title exists or an additional div in the Project Preview's label with a min-height.

Screen Shot 2020-01-20 at 10 36 49 PM

TSamee:
Closes #2102

Attempted both of the fixes suggested in the issue, but in the end I just set project-preview height relative to the viewport. Rationale was:

- Adding a separate div to act as a spacer may muddy separation of style and content
- My approaches for cleanly setting default names and timestamps broke function purity
- Absence of a project name is a small (if messy) tell that something basic needs fixing

### DIFF
--- a/src/css/application.css
+++ b/src/css/application.css
@@ -453,6 +453,7 @@ body {
   padding: 1em;
   border-bottom: 1px solid var(--color-light-gray);
   cursor: pointer;
+  height: 4.55vh;
 }
 
 .project-preview:hover {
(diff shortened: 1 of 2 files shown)


TSamee:
Thanks! Will play with both of those and revise.

outoftime:
Maybe an even cuter solution: Add an `::after` pseudoselector that adds some arbitrary character (like a `. `) with visibility hidden

PR #2275, opened 2020-05-11
Move `mustache.js` from bower dependency to npm dependency

alexpelan:
Same method as other dependency moves like #1704

Also updates mustache to latest version.


alexpelan:
Can't believe I never merged this - total pandemic brain

Issue #1888, opened 2019-10-20
Experiment with switch from Webpack to Rollup


outoftime:
Rollup apparently has much lower per-module overhead and better tree shaking, so quite possibly could substantially reduce load and parse time over Webpack. However we do rely heavily on various affordances available in Webpack and its ecosystem, so not clear whether a migration would be possible without a ton of work. Goal here is just to establish whether there are any huge hurdles to using Rollup, and if not, whether a quick-and-dirty Rollup build would give us substantial performance improvements.

outoftime:
Environment variables: https://www.npmjs.com/package/rollup-plugin-env

wilsonj806:
I can check this out! Is there any other objectives asides from meeting the above objectives and how do you want the results reported?

outoftime:
Nice @wilsonj806 ! I started a spike on this a while ago, just rebased the branch and pushed it [here](https://github.com/popcodeorg/popcode/compare/master...outoftime:rollup), so that might be useful. Where I left it was an issue where certain UMD modules, after being processed by babel, weren’t recognized by the `commonjs` plugin; but if I put the `commonjs` plugin first, it barfed on JSX files because it only parses standard JS. I _think_ the move is probably just to build two Babel passes into the plugin chain—one that _only_ compiles JSX to regular JS, done before `commonjs`; and one that applies the rest of the transforms after `commonjs`. But I’m open to other approaches if you think there’s a better one!

In terms of objectives, basically I just want to establish whether a working and functionally equivalent Rollup bundle is meaningfully smaller than what Webpack produces now. So in terms of results reported, basically:

* A branch that I can check out and see that the app works
* Comparison of combined bundle sizes when compiled in production mode

That all make senser?

wilsonj806:
Yep makes sense!

wilsonj806:
Quick update but I've got it up and running locally and have messed with it a bit. Ran into the below issues, but not sure if it's the same stuff you ran into:
- Weird transpilation error/ issue where importing default exports like: `import LRU from 'my-module` didn't work
  - sort of hacked the solution to get past it
- Error: UMD and IIFE output formats are not supported for code-splitting builds
  - this seems like a general issue with Rollup

Will need to play with Rollup some more to see.

outoftime:
Thanks for the update @wilsonj806! Feel free to open a draft PR with your progress so far, Iʼd love to check it out.

[...]

PR #2289, opened 2020-07-03
Experiment with Rollup for building assets

wilsonj806:
# General Notes
Some general notes while I was messing with Rollup:
- As an initial report, Rollup JS bundle size is about the same as Webpack(just getting file sizes from `du -sh`
- Rollup's build takes 2 minutes to build over 5 minutes with Webpack on my Macbook
  - just have the bare minimum plugins for the Rollup config so that could be explain part of why it's so fast as of 7/3/2020
- Requires you to specify local CommonJS files in the Rollup config file

Added some [process notes](https://github.com/wilsonj806/popcode/blob/TEST/rollup-experiments/branch-notes/rollup-things.md) as well.

# General Questions
- I have no clue what `locales/index.js` is for even though it's imported by `src/init/initl8n.js`

Relates to #1888


outoftime:
Thanks for the great writeup @wilsonj806! 

> As an initial report, Rollup JS bundle size is about the same as Webpack

Do you think this is dispositive? Any reason to think we could tweak the Rollup configuration to get more size efficiency out of tree shaking? If we think this result will hold, then I would say we are pretty much done here—a worthwhile experiment to be sure, but if there is no improvement to bundle size, Iʼm inclined to stick with Webpack given its richer ecosystem, fact that we have a working Webpack implementation already, etc. Thoughts?

wilsonj806:
@outoftime No problem! And yeah, at a glance it doesn't seem worthwhile to continue to mess with Rollup when there's a working Webpack config already.

outoftime:
Ten four, thanks for the investigation @wilsonj806 . Even though we’re not going to move forward it’s quite useful to have ruled out a switch to Rollup as an avenue for for bundle size optimization. I’m going to go ahead and close #1888 and you’re welcome to do the same here. 

Let me know if you’re interested in picking up another Popcode project—happy to help find one that fits your interests 🙂

wilsonj806:
@outoftime Yeah that'd be cool! I'm mostly interested in getting better at React/ Redux right now, so if you have anything related to that, I'd be down for it.

outoftime:
Heyo! Definitely lots of projects that will have you deep in React/Redux territory. A couple questions so I can calibrate my suggestions:

- How much React experience do you have now? How much experience with Redux?
- How big a project are you looking to take on? The most impactful projects will be on the bigger side, generally involving multiple pull requests, a planning phase before coding starts, etc. But we have more bite-sized work to be done as well if thatʼs your preference.

Let me know!

wilsonj806:
Re your questions:
- React experience-wise it's about a year or so, and more specifically, I feel comfortable enough with the React workflow that I can pickup things on the go
- Redux-wise I'm not nearly as comfortable, I know the basic patterns and have used Redux Thunk but haven't used Sagas as you guys are using. I'm open to learning it though
- Project size, I think I want to tackle something a bit more bit-sized to start with!

outoftime:
Nice @wilsonj806! Want to join us on https://slack.popcode.org to continue the discussion? I have one or two ideas that I think could be a good fit.

PR #2379, opened 2020-09-02
Enable sublime keymap from CodeMirror

### LINKED ISSUE #2276: User Feedback
Submitter: Peter Jablonski
Love the Code Mirror update!

It looks like the multiline comment feature is not enabled (cmd + /)

thayerve:
@paulproteus and I found that the commenting function called by Cmd-/ is one of several keyboard shortcuts included with the 'sublime' keymap of CodeMirror, so this commit just enables that whole shebang documented here: https://codemirror.net/demo/sublime.html 

Also imports the Comment addon from CodeMirror because it's called in the keymap to toggle comment.

Caveat: This may not include Windows-friendly commands, as the cmd button is specified in the keymap. 

Close #2276


outoftime:
Superseded by #2386

PR #2382, opened 2020-10-21
Adds button to hamburger menu to format code

adrianfalleiro:
This PR adds a button to the hamburger menu to format code.

### DIFF
--- a/src/components/__tests__/HamburgerMenu.test.jsx
+++ b/src/components/__tests__/HamburgerMenu.test.jsx
@@ -1,6 +1,4 @@
-import i18next from 'i18next';
 import {fromJS} from 'immutable';
-import constant from 'lodash-es/constant';
 import React from 'react';
 import {Provider} from 'react-redux';
 import {act, create} from 'react-test-renderer';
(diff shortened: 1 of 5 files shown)


arrowtype:
Wow, I’m glad I asked the community for someone who knows React & Redux. This was more involved than I realized! I can’t comment on the exact code quality here (hopefully @outoftime can take a look soon), but thanks so much for hopping in and making it happen.

src/components/__tests__/HamburgerMenu.test.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+describe('hamburger menu', () => {
+  let menu;
+
+  beforeEach(() => {
+    menu = renderComponent();
+  });
+
+  it('calls the passed `onAutoFormat()` method when format code button is pressed', () => {
+    const firstMenuItem = menu.root.findAllByType(MenuItem)[0];
```
outoftime, line 63: Would it be fairly straightforward to scan for a menu item with the text label we expect? I could see this turning into a false failure if a future change introduced a menu item above format code that was always on, for instance.

outoftime (approved):
Left one comment on the tests, but not a blocker—this looks great, thanks! Reviewing from my phone, will merge once I can get in front of a computer to do a quick manual validation.

src/components/__tests__/HamburgerMenu.test.jsx
adrianfalleiro: @outoftime This is a good callout. The issue is that the translation function `i18next.t()` is mocked and returns a fixed value. 

What I can do is amend the mock to return back the translation key and use that to select the menu item.

Ideally would be good to use `data-testid` attrs in combination with `@testing-library/react` or `enzyme` but that's for another PR.

PR #2386, opened 2020-12-10
Adds support for Ctrl-/ block commenting

adrianfalleiro:
Adds support to the editor to be able to block comment code.

Kept the original commit from @thayerve so they get credit.


outoftime (approved):
Nice!!

PR #2430, opened 2021-01-24
Merge branch 'enlarge-by-default' of https://github.com/ifeLawal/popcode into enlarge-by-default

ifeLawal:


ifeLawal:
This is a fix to issue popcodeorg#1554 Zoom enlarged by default

I used a CSS media query to increase the font-size for the editor text

I assumed the iFrame text should not be increased since students have access to that via code, but just in case it would also be beneficial I tackled making the iFrame font size bigger as well by embedding a style tag in the Preview Frame source

Would love feedback and thoughts on the approach / if there should be any changes

